PR #3, opened 2022-02-03
bugfix: correct pwm duty cycle

talltechdude:
PWM output was not getting past 10% in `Channel_OnChanged` (confirmed on Oscilloscope). Copied similar math operation from `PIN_SetPinRoleForPinIndex` function


btsimonh:
Q: is HA intended to send 0-999 or 0-99?

openshwprojects:
HA sends whatever we want as defined in configuration yaml.

I think that @talltechdude is correct, I will check again and approve. Thanks!

PR #8, opened 2022-02-07
feat: redirect root page / to /index page

talltechdude:
Redirects the root `/` url to `/index` instead of showing `Not found`.


openshwprojects:
You're sending a redirect, but wouldn't it be easier to just check for both OR / or /index in the if statement? Maybe your way is more elegant, I'll approve this change.

PR #10, opened 2022-02-07
Btsimonh5 - webserver updates

btsimonh:
The main thing this does is base everything on a http_request_t structure.
It also introduces a 'poststr(p_request, str)' for adding data to the http response, which automatically checks the available size, and really sends if it has to.

This should make it easier to create (for example) a rest style API, essential for an external web interface.
It should also avoid breaking the server by expanding the number of pin entries, etc.

Please check it over carefully.  I have built & tested in windows, but not on a real device (as I have no serial connection temporarily!).

### DIFF
--- a/src/httpserver/http_tcp_server.c
+++ b/src/httpserver/http_tcp_server.c
@@ -39,7 +39,6 @@ static void tcp_client_thread( beken_thread_arg_t arg )
 {
   OSStatus err = kNoErr;
   int fd = (int) arg;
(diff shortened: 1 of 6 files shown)


btsimonh:
logs indicate mqtt disconnecting all the time, reason 256 - could be my mqtt server, it was restarted since last time I checked logs.
Everything else (including OTA) seems to be working.

btsimonh:
MQTT issues with two devices resolved?

PR #11, opened 2022-02-08
Btsimonh7

btsimonh:
a few bugfixes (windows first request).
Adds a JSON parser and rest_interface.c as a *test* for JSON usage.
Adds /lograw endpoint
Enables replying in binary.... (won't help unless we sent content-length).


btsimonh:
also added MSCV 2010 project....

PR #12, opened 2022-02-09
MQTT small mods

btsimonh:
Enables MQTT host to be a hostname
moves MQTT features to a new folder/file. (no material changes)
rename tuya_main.c to user_main.c


btsimonh:
I've moved the primary webapp source to the OpenBekenIOT organisation.

PR #16, opened 2022-02-11
fix: MQTT client never gets created

talltechdude:
After the shuffling of `mqtt_client_new()` in fac54bd376695981d9447f37f5e1ad834b517a89 it seemed to never get called even in the timer loop since `mqtt_client` was always == 0 and never initialsed the first time. Feel free to cancel if @btsimonh has already fixed.

### DIFF
--- a/src/user_main.c
+++ b/src/user_main.c
@@ -156,7 +156,11 @@ static void app_led_timer_handler(void *data)
 			example_do_connect(mqtt_client);
 			loopsWithDisconnected = 0;
 		}
-	}
+	} else if (mqtt_client == 0) {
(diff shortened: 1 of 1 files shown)


btsimonh:
nope, no mods around that.
I'm not using MQTT regularly (yet), so had not noticed if it was not working :(.

talltechdude:
No probs. Been wrestling with my CCT downlights getting them into Home Assistant via MQTT so found this issue pretty quickly.

openshwprojects (commented):
This is indeed a bug but I fixed it already on my side

PR #17, opened 2022-02-11
fix: send pwm values (0...100) back via MQTT

talltechdude:
MQTT publish was only sending 0 or 1 even for PWM pins. Refactored `Channel_OnChanged` to still send 0/1 for relay & led pins but 0...100 for PWM enabled pins.


openshwprojects:
It seems correct. Thanks.

PR #19, opened 2022-02-12
Littlefs - more

btsimonh:
Looking pretty good now....

lfs file endpoints - 
POST /api/lfs/folder/filename to write a file.  
GET /api/lfs/folder/filename to read/serve a file.

add reboot POST /api/reboot (3 second delay)
add OTA POST /api/ota (not used this yet, but below work well)
add fs backup GET /api/fsblock (reads as 512k file)
add fs flash POST /api/fsblock (writes the data your provide - should be a 512k little FS image)

Note: LittleFS blocks are at OTA + 0x1000 (so we don't write the header).
It's overwritten by OTA, hence the ability to backup and restore it.....

I still have random crashes/disconnects.  serial logs have some clues, so will try to 'catch it at it' and look into the logs.
Most likely we still have some thread conflict issues.

### DIFF
--- a/src/httpserver/http_tcp_server.c
+++ b/src/httpserver/http_tcp_server.c
@@ -57,7 +57,9 @@ static void tcp_client_thread( beken_thread_arg_t arg )
 
   request.fd = fd;
   request.received = buf;
-  request.receivedLen = recv( fd, request.received, 1024, 0 );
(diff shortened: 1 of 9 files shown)


btsimonh:
added a few bits, and merged your changes

PR #25, opened 2022-02-16
move config from SDK to app - the SDK mods can now be removed

btsimonh:
minor changes just to finish freeing ourselves from the net_params in the SDK.


btsimonh:
also adds a link to webapp, and adds obktype in webapp html, in case we need it later.

openshwprojects (approved):

PR #27, opened 2022-02-17
add GET /api/flash/<hexstart>-<hexend> to read any part of flash

btsimonh:
e.g. http://192.168.1.176/api/flash/1e0000-1000
reads RF setup table
add POST /api/flash/<hex addr> to WRITE anywhere after 0x132000


btsimonh:
in a browser, will download a binary file :).

openshwprojects (approved):

PR #36, opened 2022-02-23
MQTT updates

btsimonh:
Adds ability to add callbacks to receive MQTT messages.
NOTE: I have not had time to test this this morning, but I thought you would like to see it..... (and maybe test?).

I think we should change the subscribed topic to be more TAS - 
e.g. cmnd/<basename>/<object>
Then we can subscribe to all cmnd/<base> without hearing stat/<base>/<object>

If/when we have scripting, this will be useful.  e.g. we can add the functionality to LUA to get relevant MQTT delivered direct to it, processed, and cause actions.  The specific topics can be requested directly from LUA....

br,
s

### DIFF
--- a/src/httpserver/new_http.c
+++ b/src/httpserver/new_http.c
@@ -67,6 +67,7 @@ Connection: keep-alive
 #elif defined(PLATFORM_BK7231T)
 #define USER_SW_VER "BK7231T_Test"
 #else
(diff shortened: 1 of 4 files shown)


btsimonh:
Ready now......

openshwprojects (approved):

PR #44, opened 2022-02-27
Testbuild

btsimonh:
builds withing submodules?

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -79,9 +79,9 @@ jobs:
           rm -rf OpenBK7231T || true
           rm -rf OpenBK7231N || true
           rm -rf OpenXR809 || true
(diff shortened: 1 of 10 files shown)


talltechdude:
@btsimonh, were these submodules causing you build issues locally? They should only be checked out if running the builds from this makefile, e.g. `make OpenBK7231T` etc and shouldn't be in the way otherwise (unless I missed something!)

This also allows dev to happen on just one copy of the `_App` repo instead of juggling all 3 checkouts inside different SDK repos previously.

btsimonh:
@talltechdude - yep, the second time I merged in sourcetress, it checked out all three.. :(.

btsimonh:
yep, it's a sourcetree option.... now disabled.  we'll mothball this PR...

PR #45, opened 2022-02-27
feat: update sdks from Makefile

talltechdude:
I've added a Makefile command to automatically update the SDK submodules, effectively just a git pull on each and commit the changes.
```
make update-submodules
```

Also fixed the fail-fast logic as it was cancelling other builds if any platform failed.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -58,6 +58,7 @@ jobs:
     needs: refs
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         platform: [OpenBK7231T, OpenBK7231N, OpenXR809]
     steps:
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,13 @@ full: clean all
 submodules:
 	git submodule update --init --recursive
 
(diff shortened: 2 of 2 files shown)


openshwprojects:
Seems good, I will check this out.

I have a question - do having a submodules in "sdk" directory forces us to have them also on disk when checking up the App repository? 
I have reports from @btsimonh that is the case and  he'd prefer the repositories to be checked out on github-only side, see:
https://github.com/openshwprojects/OpenBK7231T_App/pull/44

talltechdude:
The submodules should only be checked out if running the builds using the new Makefile when `git submodule init` is run so they shouldn't download if you're not using it.

The submodules feature should also be useful for normal development too as you can run make builds directly from the App repo instead of having to manually check out all the SDKs and keep 3 separate app repos updated (to each their own there though)

openshwprojects:
@talltechdude  , would it be possible to run this entirely on Github, and not on my local machine?
Maybe automatically?

I just updated the XR809 SDK (added file to makefile) and now I need to run "make update-submodules", would it be possible for Github build script to do it entirely online, everytime automatic build is run?

PR #46, opened 2022-02-27
Newlines all changed to unix style

ExploWare:


ExploWare:
at least to one style. Not sure windows or unix in this commit

talltechdude:
@ExploWare, this is something you can configure on your own Git client to handle automatically (e.g. https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings). Git can automatically handle the different line endings for you and you don't need to keep switching between them with different commits/prs:

From the Windows config linked above for Global settings:
```
$ git config --global core.autocrlf true
# Configure Git to ensure line endings in files you checkout are correct for Windows.
# For compatibility, line endings are converted to Unix style when you commit files.
```

It also walks through normalising your repo again to get the line endings back to expected - https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#refreshing-a-repository-after-changing-line-endings

ExploWare:
@talltechdude thanks
I have found that solution already and applied it on my local system.

This PR is to get all current files in shape, as it is possible to edit in Github through the browser which would also correct line-endings over the whole file, when only one character was edited, and so @openshwprojects did receive a lot of unneeded modified lines.

openshwprojects:
@talltechdude I agree, I think all line endings should be fixed to be unix. Actually only a very small portion of endings might be windows standard, because I started http server as a windows app to ease the development process.

PR #47, opened 2022-02-28
Comments, 'Uptime' lowercase 't'

ExploWare:
trying to improve readability step by step.
Added some comments on functions
Please see this as part one of many. Also as a hint for others to improve documentation in comments during coding, so others understand what is happening where, and are more able to improve/add functionality.

### DIFF
--- a/src/httpserver/new_http.c
+++ b/src/httpserver/new_http.c
@@ -1,24 +1,24 @@
-
-
 #include "../new_common.h"
 #include "ctype.h" 
-#if WINDOWS
+#if WINDOWS // debug mode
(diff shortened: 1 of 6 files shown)


openshwprojects:
According to our new automated build system, this change fails to compile on XR809 platform. Also, maybe you should wait with formatting for a more stable stage of development.

ExploWare:
I noticed the fail. Surprisingly on a broadly used function.
Formatting is less important, but commenting is not. 
I do not understand the whole yet, and comments would help a lot.
 So that's where I'm going for snd vlean lingo is helpful with that also

PR #59, opened 2022-03-06
Cmnds

btsimonh:
commands: all strings now const (no more modification).
split CMD_Exec into two - for easier use in MQTT.
move some commands into cmnds/taslike.c
Fix bug in CMD_Find.

commands added:
POWER n (channel 0)
POWERm n (channel m)
COLOR #RRGGBB[cw][ww] - PWM control assuming colour.

(these were already in MQTT, but now MQTT just calls CMD_Execute instead!!!)

### DIFF
--- a/src/cmnds/fortest.c
+++ b/src/cmnds/fortest.c
@@ -1,0 +1,68 @@
+
+
+#include "../logging/logging.h"
+#include "../new_pins.h"
+#include "../obk_config.h"
+#include <ctype.h>
+#include "../new_cmd.h"
+#ifdef BK_LITTLEFS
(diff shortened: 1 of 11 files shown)


btsimonh:
added another file 'cmnds/fortest.c'
Adds a command ADDCMD
This can be used to store an alias for any other command(s). 

e.g. 
addcmd fred loglevel 5

then
fred

-> loglevel changed to 5.

or addcmd red color #FF00000000
now command 'red' is installed which makes the colour red....

btsimonh:
note 
addcmd fred fred would not be good? :).

PR #70, opened 2022-04-12
Publish UG artifacts in releases for stock OTA

khalednassar:
Changes the GH workflow to publish UG files (Tuya OTA format) as well. This is useful for users who want a quick n' easy download for a particular OpenBK7231 image to remotely OTA with the stock firmware.


btsimonh:
thankyou :).  I'll let @openshwprojects check it - I guess we'll want this in the N version modified as well...

khalednassar:
Oh, I also added the change for N version since it was in the outputs. Got me under the impression that this is now the "common" repo for T/N and the XR809. In any case, I don't see it on the repositories page, is it private?

btsimonh:
no, I think you are right.  There may be some SDK differences to content with though...

openshwprojects:
Thank you. I just tought it would be needed, and here it is!

PR #71, opened 2022-04-12
Added callback for input pin

bigMemer:


bigMemer:
Hey my friend, reading the pin worked great... except that since it isn't hooked in to the callback it doesn't actually broadcast the change via MQTT

PR #75, opened 2022-04-23
add support for moes WIFI Smart Dimmer Module 105 (TuyaMCU dimmer, WB2S based)

heikobihr:
This PR adds suppport for moes WIFI Smart Dimmer Module 105 (WB2S).

* add a new channel type "dimmer", using same slider control as PWM
* maps TuyaMCU's dimming range to a normalized interval 0..100 (for web UI and MQTT)
* propagate new channel values to TuyaMCU (via TuyaMCU_OnChannelChanged)
* store TuyaMCU data point type for each mapped channel

moes WIFI Smart Dimmer Module 105 is controlled by two data points:
1: DP_TYPE_BOOL: on/off
2: DP_TYPE_VALUE: dimming level (0..1000)

autoexec.bat (working for me):
```
startDriver TuyaMCU
setChannelType 1 toggle
setChannelType 2 dimmer
tuyaMcu_setDimmerRange 0 1000
linkTuyaMCUOutputToChannel 1 1 1
linkTuyaMCUOutputToChannel 2 2 2
```

### DIFF
--- a/src/driver/drv_tuyaMCU.c
+++ b/src/driver/drv_tuyaMCU.c
@@ -573,7 +573,7 @@ void TuyaMCU_ApplyMapping(int fnID, int value) {
 		addLogAdv(LOG_DEBUG, LOG_FEATURE_TUYAMCU,"TuyaMCU_ApplyMapping: mapped value %d (TuyaMCU range) to %d (OpenBK7321T_App range)\n", value, mappedValue);
 	}
 
-	CHANNEL_Set(mapping->channel,mappedValue,0);
+	CHANNEL_SetEx(mapping->channel,mappedValue,false,false);
 }
 
 void TuyaMCU_OnChannelChanged(int channel, int iVal) {
--- a/src/driver/drv_tuyaMCU.c
+++ b/src/driver/drv_tuyaMCU.c
@@ -103,6 +103,8 @@ typedef struct tuyaMCUMapping_s {
 	int channel;
 	// data point type (one of the DP_TYPE_xxx defines)
 	int dpType;
+	// store last channel value to avoid sending it again
+	int prevValue;
 	// TODO
 	//int mode;
 	// list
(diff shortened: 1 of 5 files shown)


openshwprojects:
That's great, I had this device in queue so I will be able to test as well.

My only question is - is this stable? Did you check this running for at least a day?

heikobihr:
> That's great, I had this device in queue so I will be able to test as well.
Yes, I saw your post regarding this device on elektroda forums :-)

> My only question is - is this stable? Did you check this running for at least a day?
Yes, it looks stable :-)
The device is now running for 16 hours and still works (controlling via MQTT, WebUI and local push button).

openshwprojects:
@heikobihr  I am asking because I heard some reports from @btsimonh saying that I (and Tasmota as well) should follow TuyaMCU device state and at least wait for the packet reply before sending next one.

He stated that the same problem is both in Tasmota and my code and that it breaks his dimmer (on Tasmota, not on Openbeken)

I am going to do something to fix that that, maybe add a mechanism to queue TuyaMCU packets in case that device tries to send two TuyaMCU packets at once... and only send one, wait for reply, then send second

Do you know anything about this issue?

anyway, I don't think it collides with your changes, so of course I will accept this pull request. Great work, this is exactly what I have planned to do.

heikobihr:
> I am going to do something to fix that that, maybe add a mechanism to queue TuyaMCU packets in case that device tries to send two TuyaMCU packets at once... and only send one, wait for reply, then send second
> 
> Do you know anything about this issue?

No, I did not run into this issue yet. This is my first TuyaMCU based device, so I am still learning ;-)

> anyway, I don't think it collides with your changes, so of course I will accept this pull request. Great work, this is exactly what I have planned to do.

Thanks :-)
But in fact, I have to thank you for your great work on OpenBeken :-)

openshwprojects:
@heikobihr I am merging the pull request now and I will test it soon with my dimmer as well, but please also keep testing it yourself and report here (or on Elektroda) if you run into any issues.

openshwprojects:
I can confirm that it works, at least I have encoutered no problems after first 5 minutes of testing

PR #113, opened 2022-07-10
WiFi LED blink duration issue when changing PIN_TMR_DURATION

anogma:
If PIN_TMR_DURATION is different than 5ms, the WiFi LED blinks at incorrect intervals.

### DIFF
--- a/src/new_pins.c
+++ b/src/new_pins.c
@@ -18,10 +18,13 @@
 
 
 //According to your need to modify the constants.
-#define BTN_TICKS_INTERVAL    5	//ms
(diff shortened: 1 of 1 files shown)


openshwprojects:
Thanks, of course it's a better to make it flexible

anogma:
Sorry, it was not intended. I wanted to create two separate PRs. Is it possible to accept just the correct commit? If not, I can make another PR.

openshwprojects:
It would have to be done from command line, as I see no interface for that on git.

Can you submit that as separate PR?

PR #126, opened 2022-07-22
feat: add OpenW800 Platform

talltechdude:
Adding the new OpenW800 platform to the auto-build Github actions and releases.

As before, there's some SDK changes required too in https://github.com/openshwprojects/OpenW800/pull/1 to get this to work and once merged I can update the submodule back to `openshwprojects` repo.

One thing to note though is that each required source files need to be added to the OpenW800's `app/Makefile` similar to the `.project` file under the GUI IDE. Alternatively, we'd could add `#ifdef PLATFORM_W800` to all the `.c` files that aren't required for the W800 and recursively build all `.c` files if the list to exclude is much shorter.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -64,7 +64,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        platform: [OpenBK7231T, OpenBK7231N, OpenXR809, OpenBL602]
+        platform: [OpenBK7231T, OpenBK7231N, OpenXR809, OpenBL602, OpenW800]
     steps:
       - name: Source checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


openshwprojects:
I think adding the source files to OpenW800 repo makelist is a better and cleaner solution.

Well done, tell me when it's ready to be merged.

openshwprojects:
Btw: XR809 build fails from time to time because it expects some kind of "partition table file cfg" to be kept after failed build and it needs to use it again to build it correctly, would it be possible to fix it somehow?

The "dirty" solution would be to commit new cfg file every time, but maybe there is a way to somehow run the XR809 build twice in the Github, so the first run generates correct .cal file and then second build uses it..

openshwprojects:
Also please publish w800_ota.img for OTA on W800

talltechdude:
> Btw: XR809 build fails from time to time because it expects some kind of "partition table file cfg" to be kept after failed build and it needs to use it again to build it correctly, would it be possible to fix it somehow?
> 
> The "dirty" solution would be to commit new cfg file every time, but maybe there is a way to somehow run the XR809 build twice in the Github, so the first run generates correct .cal file and then second build uses it..

Ok, I've added a for loop to retry the build multiple times to try to get around the .cal file issue. The build is still failing based on a number of undefined references now.

Otherwise this PR should be ready to go, added the `_ota.img` file too.

openshwprojects:
I am aware about the unreferenced issues, I will fix them later. Thank you!

PR #140, opened 2022-07-31
feat: Quick Config template for CasaLife CCT Downlight

ilengyel:
Quick Config template for 'CasaLife CCT Downlight SMART-AL2017-TGTS'


openshwprojects:
Thank you!

Issue #138, opened 2022-07-27
Generate Home Assistant cfg using deprecated configuration format


dstreefkerk:
Hi there!

I used the "Generate Home Assistant cfg" feature to create the config for a 3-gang switch in Home Assistant, and later noticed in the Home Assistant logs that there was warning about a deprecated switch configuration.

To quote [this page](https://www.home-assistant.io/integrations/switch.mqtt/#new_format): 

> "The configuration format of manual configured MQTT items has changed. The old format that places configurations under the switch platform key should no longer be used and is deprecated."

Here's the config generated for my device by OpenBK:
```
switch:
  - platform: mqtt
    name: "familyroom-lightswitch 1"
    state_topic: "familyroom-lightswitch/1/get"
    command_topic: "familyroom-lightswitch/1/set"
    qos: 1
    payload_on: 0
    payload_off: 1
    retain: true
    availability_topic: "familyroom-lightswitch/connected"
  - platform: mqtt
    name: "familyroom-lightswitch 2"
    state_topic: "familyroom-lightswitch/2/get"
    command_topic: "familyroom-lightswitch/2/set"
    qos: 1
    payload_on: 0
    payload_off: 1
    retain: true
    availability_topic: "familyroom-lightswitch/connected"
  - platform: mqtt
    name: "familyroom-lightswitch 3"
    state_topic: "familyroom-lightswitch/3/get"
    command_topic: "familyroom-lightswitch/3/set"
    qos: 1
    payload_on: 0
    payload_off: 1
    retain: true
    availability_topic: "familyroom-lightswitch/connected"
```

And here's my revised config, following the new practice for MQTT switches:
```
mqtt:
  switch:
    - unique_id: familyroom_light_switch_1
      name: "Family Room Light Switch 1"
      state_topic: "familyroom-lightswitch/1/get"
      command_topic: "familyroom-lightswitch/1/set"
      availability:
        - topic: "familyroom-lightswitch/connected"
      payload_on: 1
      payload_off: 0
      qos: 1
      retain: true
    - unique_id: familyroom_light_switch_2
      name: "Family Room Light Switch 2"
      state_topic: "familyroom-lightswitch/2/get"
      command_topic: "familyroom-lightswitch/2/set"
      availability:
        - topic: "familyroom-lightswitch/connected"
      payload_on: 1
      payload_off: 0
      qos: 1
      retain: true
    - unique_id: familyroom_light_switch_3
[...]

PR #141, opened 2022-08-02
Emit HomeAssistant config in new style

iprak:
This addresses #138 

* The HA config is generated in new style as per https://www.home-assistant.io/integrations/switch.mqtt/#new_format. 
* The light config also has some matching adjustment for `availability`.

Both switch and light have `unique_id` with the same value as the `name`.  `unique_id` allows the entity to be customized and one  would see something like this instead: image


openshwprojects:
Wow, great, thanks. I knew it had to be done but just had a busy time this week.

PR #148, opened 2022-08-07
Quick config for Enbrighten WFD4103

iprak:
The device wiki is temporary/under review: https://www.elektroda.com/rtvforum/viewtopic.php?p=20133554#20133554


iprak:
The wiki page seems to have been approved.

openshwprojects:
Thank you for both profile and the teardown!

PR #157, opened 2022-08-15
Generate better uniqued id, fixed payload

iprak:
* Updated `unique_id` based on https://developers.home-assistant.io/docs/entity_registry_index/
* Also updated on/off payload  #138

### DIFF
--- a/src/httpserver/http_fns.c
+++ b/src/httpserver/http_fns.c
@@ -1246,8 +1246,7 @@ char *build_hass_unique_id(char *type, int index){
     //unique_id and I would think that longDeviceName should contain that.
     
     //e.g. longDeviceName_relay_1
-    int length = strlen(g_cfg.longDeviceName) + strlen(type)+ 1 + 3;
-    char *unique_id = (char*)malloc (length * sizeof (char));
+    char unique_id[128];    //64 for longDeviceName, 10 for type,3 for index .. 128 would be sufficient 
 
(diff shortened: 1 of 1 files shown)


pretoriano80:
@iprak i think that would be a better practice to split this in two PRs,one PR for each issue/bug fix.
It's easier to keep track of the changes and also easier if a revert it's required .

iprak:
> @iprak i think that would be a better practice to split this in two PRs,one PR for each issue/bug fix. It's easier to keep track of the changes and also easier if a revert it's required .

That is a good idea. I will try to keep that in mind for future merges. :-)

openshwprojects:
The idea is good but I need you to fix the memory leak, because you are doing "malloc" and not freeing the memory.

The best solution would to be put something like char buffer[128];  into the function generating config and just operate on that

openshwprojects:
One more fix - having:

char *getSmth() {
  char smth[127];
  strcpy(smth,"qqq");
  return smth;
}

void tst() {
  char *a = getSmth();
}

is not a correct practise, because C does not guarante that the buffer will still exist after returning from the function. This is like a russian roulette crash. The contents of smth are undetermined after returning from function.

You should rather do:

void getSmth(char *outBuffer, int bufferLen) {
  strcpy(smth,"qqq"); // better use safe strcpy
}

void tst() {
 char bufferToUse[256];
  getSmth(bufferToUse,sizeof(bufferToUse));
}

PR #160, opened 2022-08-16
Ntp config

iprak:
This request adds supports for custom NTP server in place of the default `217.147.223.78`. This allows NTP functionality to work even in firewalled devices.

* A new setting 20 char long has been created for this purpose.
* The new setting can be set using the command `ntp_setServer ipAddress`.
* Added another command `ntp_info` to view current NTP driver settings.

### DIFF
--- a/src/driver/drv_ntp.c
+++ b/src/driver/drv_ntp.c
@@ -68,6 +68,8 @@ int NTP_SetTimeZoneOfs(const void *context, const char *cmd, const char *args, i
 	addLogAdv(LOG_INFO, LOG_FEATURE_NTP,"NTP offset set, wait for next ntp packet to apply changes\n");
 	return 1;
 }
+
+//Set custom NTP server
 int NTP_SetServer(const void *context, const char *cmd, const char *args, int cmdFlags) {
 	Tokenizer_TokenizeString(args);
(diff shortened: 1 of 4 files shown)


openshwprojects:
Seems like a very good change. I will review it in detail later, but have you made sure not to break the earlier configs?

iprak:
> Seems like a very good change. I will review it in detail later, but have you made sure not to break the earlier configs?

I created the new setting out of an `unusedSector`. Since the byte positions are unchanged so I think existing settings will not change. I did deploy the firmware in a device and will test it on another device.

Based on field name and past history, it seemed that those unusedSectors were there to avoid resetting previous settings ?

openshwprojects:
This is correct. I think it should work OK then.

iprak:
As I understand `inet_addr` only accepts IPv4 address and not hostname which is why I limited it to 20. We could extend it to 32 if we plan to do dns translation some day.

iprak:
I increased the field size to 32, 32 seems like a good size and is sufficient to accommodate the common ntp pool servers. I can separately look at DNS resolution in this regards.

openshwprojects:
I reviewed it briefly and it seems like a good change. Good job! Pull request merged.

PR #161, opened 2022-08-17
Improve flags config html

iprak:
This request 
* Aligns the layout of checkbox (before the label and not underneath it)
* Makes the checkbox more accessible (label can be clicked to change the state)

Before:
image

New layout:
image

### DIFF
--- a/src/httpserver/http_fns.c
+++ b/src/httpserver/http_fns.c
@@ -1696,24 +1696,20 @@ int http_fn_cfg_generic(http_request_t *request) {
 
 	hprintf128(request,"<h5>Flags (Current value=%i)<h5>",CFG_GetFlags());
 	poststr(request,"<form action=\"/cfg_generic\">");
+
 	for(i = 0; i < OBK_TOTAL_FLAGS; i++) {
-		int bHas;
-		const char *flagName;
+		const char *flagName = g_obk_flagNames[i];
+/*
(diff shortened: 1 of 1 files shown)


openshwprojects:
Thanks, by the way, there is one change we need... we need to move the link for the javascript panel which is now "hidden" under options (a small URL under buttons) to the main page, to the button. Create a button with that link on the main page under "Options" so it's clearily visible. Would you like to do that?

iprak:
> Thanks, by the way, there is one change we need... we need to move the link for the javascript panel which is now "hidden" under options (a small URL under buttons) to the main page, to the button. Create a button with that link on the main page under "Options" so it's clearily visible. Would you like to do that?

I can do that but I am not following what you are referring to.

pretoriano80:
@iprak i think he mean to have a button to open the WebApp just like the other options (eg. the Quick Config button ).
At the moment there's just a small link/hyperlink at the bottom of the WebUI ,which is easy to miss .
And probably it's best to have it on the WebUI main page .

iprak:
I will let @openshwprojects confirm if that is what he requested. I agree that having `[Launch Web Application` on the main page would be more useful.

openshwprojects:
I think you are correct. My suggestion was to remove this:
image
and add here (with a better Title, better color, etc):
image

openshwprojects (approved):

PR #174, opened 2022-08-23
Handling client database settings

iprak:
This request allows completely defining device setting via the new webApp (currently in https://github.com/OpenBekenIOT/webapp/pull/4). Without this, the webApp will not show the new UI and still remain backward compatible.
* It includes handling of command and flag values.

### DIFF
--- a/src/httpserver/rest_interface.c
+++ b/src/httpserver/rest_interface.c
@@ -119,19 +119,6 @@ const char * apppage4 = "startup.js\"></script>"
 "</body>"
 "</html>";
 
-const KeyIntegerTuple_t OBKFlagValues[OBK_TOTAL_FLAGS] = {
-	{"OBK_FLAG_MQTT_BROADCASTLEDPARAMSTOGETHER", OBK_FLAG_MQTT_BROADCASTLEDPARAMSTOGETHER},
-	{"OBK_FLAG_MQTT_BROADCASTLEDFINALCOLOR", OBK_FLAG_MQTT_BROADCASTLEDFINALCOLOR},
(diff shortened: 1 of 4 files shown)


iprak:
@openshwprojects I am unable to figure out what the BL602 build is failing.

The error indicates that the function `tryGetTokenString` is not defined. I defined that function in jsmn_h.h/jsmn.c next to `jsoneq` and I don't see any conditional build definition, so very confused.

openshwprojects:
I will look into that.

Quick question: how hard would it be to get the rid off string names "OBK_FLAG_MQTT_BROADCASTLEDPARAMSTOGETHER" from the flash memory of device and let the Javascript App just communicate by using integer values of flags? 

Just to save the memory, reduce memory usage

iprak:
That would be very easy. I think only a couple definitions used flags.

openshwprojects:
The integer update is great, now this is very memory efficient. Now what about BL602, hmm

iprak:
I did not see I`jsmn.c` in the raw log, maybe it is not in the files-to-build list. 

But then how could compiles be succeeding? Could the compile be using some older cached `jsmn.a` ?

iprak:
The build failure seems to be due to how `jsmn` library is being consumed. jsmn.h actually contains the source code which is very confusing. 

Moreover in BL602 sdk, jsmn library comes from `OpenBL602/components/3rdparty/aws-iot/aws-iot-device-sdk-embedded-C/src/aws_iot_json_utils.c` and the local jsmn.c is never used. The embedded library 

I moved the function tryGetTokenString into rest_interface.c and reverted changes to jsmn files.

PR #177, opened 2022-09-01
Status information auto-updates, tweaked HTML

softplus:
First attempt to tweak some of the HTML in the UI. I'd like to update things to be a bit easier to use, potentially more consistent with Tasmota. Starting small as I figure things out in the code. 

* Made the links to repo, forum thread, donation open with target=_blank (in a new tab)
* Tweaked the UI to move all the status-information (state of devices, watchdog) together (this moves the shared UI items down a bit)
* Made it possible to just request the status information via http://URL/index?state= 
* Made the status-information section auto-updating with JavaScript every 3 seconds

First commit :)

### DIFF
--- a/src/httpserver/http_fns.c
+++ b/src/httpserver/http_fns.c
@@ -105,6 +105,7 @@ int http_fn_testmsg(http_request_t *request) {
     return 0;
 
 }
+
 int http_fn_index(http_request_t *request) {
     int j, i;
 	char tmpA[128];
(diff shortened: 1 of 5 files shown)


openshwprojects:
Seems like a great update. Thanks! I will merge it, but also test it a bit.

openshwprojects:
Well, after I short testing I've found a problem.
Have you tested the refreshing with Dimmers?
image
The current version does not respond to change of slider. This needs to be fixed before your changes goes public.

openshwprojects:
Can you fix those issues and open a new pull request with all your changes from the start?

softplus:
Interesting - I see that now. I'll try the various types and double-check that they work. thanks!

openshwprojects:
No problem, I'm eager to see the results. Remember that even if you don't have a device that is using PWM, you can still set PWM role to one of the PWM pins just to test stuff and see if it works.

PR #183, opened 2022-09-04
HomeAssistant MQTT Discovery

iprak:
This merge implements HA MQTT Discovery (https://github.com/openshwprojects/OpenBK7231T_App/issues/79).

### End User Changes
The `Home Assistant Cfg` page has some new button to send discovery MQTT and configure MQTT. An override discovery topic  can be defined as per https://www.home-assistant.io/docs/mqtt/discovery/.
image

After HomeAssistant has received the discovery message, it should automatically list the device.
image

Along with the discovery payload
image

### Code Changes
* The JSON for discovery is generated using cJSON. The HASS specific functionality is in a new file named `hass.c`.
* Clicking the button queues up this operation. `new_mqtt.c` has been updated to support queuing messages. Up to 3 queued messages are sent out at once. Testing (see later) confirmed that it worked well at least in BK7231T.
* I also added a new MANUFACTURER constant.

The BK7231T SDK has `cJSON.h` in `sdk\include` folder . The file defines redirects to `ty_cJSON_*` functions which are not implemented. This file is referenced by some other files in the same folder. The other platforms (BL602, XR809) have older cJSON implementations in the SDK. I opted to use the latest cJSON codebase and have it be inside `OpenBK7231T_App` so that the outcome is consistent.
* The cJSON library copied from https://github.com/DaveGamble/cJSON (b45f48e600671feade0b6bd65d1c69de7899f2be)  is present in the folder `cJSON`.
*  BK7231T - `sdk\include\cJSON.h` has been commented out to prevent build errors. https://github.com/openshwprojects/OpenBK7231T/pull/7
* BL602 - I modified `bouffalo.mk` as part of this merge to include build of the new cJSON folder.
* W800 - this SDK had an older cJSON. The Makefile has been adjusted to include new .cs files and skip building SDK cJSON component https://github.com/openshwprojects/OpenW800/pull/2.
* XR809 - same changes as W800 https://github.com/openshwprojects/OpenXR809/pull/3

### Testing
I tested creating a fake device with 3 relays which resulted in 3 MQTT messages being sent without any issues.

```
[...]

PR #188, opened 2022-09-05
Fixed Html content issue on the MQTT setup form

### LINKED ISSUE #185: MQTT setup / Clicking Client or Password label focused Port
This is due to `for="port"` being used for those 2 label. I will clean that up.

iprak:
Closes #185

* Refactored the label/field generation to a new functions `add_label_field`, `add_label_numeric_field`
* Changed port be be number field


openshwprojects:
Nice, seems like a good change.
Regarding the comment in code:
_"//These individual strings should be less than 256 .. yes hprintf128 uses 256 char buffer"_
it was my quick change that I recently added to avoid code crashing when.... when a "short command line" were longer than 128 characters.
It's alraedy fixed properly, but the increased buffer is still left in code. Just in case.
In general, it's better to avoid large sprintfs/va_sprintfs etc.

There is also one think that could be done to improve - the ecurrent hprintf128 will crash randomly if you create a string longer than 128 (or not, currently 256) bytes. There should be some kind of lenght within range check added to this function.

But for now, in general, just make sure to use multiple hprintf128 calls in order  to never exceed the buffer.

Btw, is the pull request ready? I seem to have it marked as "draft"

iprak:
It is ready. I plan to adjust the existing code to use the new function but I will do that as a separate PR.

iprak:
Please review https://github.com/openshwprojects/OpenBK7231T_App/pull/183 first.

I went ahead and adjusted the code to use the new function at many places.

Also fixed these labels which were linking to incorrect fields:
* Take action after this number of seconds with no reply
* Interval between pings

3 fields were using `size` attribute in an attempt to control field width but the styling was always forcing 100% width, so I took out the `size` attribute . These were in http_fn_cmd_tool, http_fn_startup_command and http_fn_uart_tool.

openshwprojects:
One of the parts that needs to be improved may be MQTT_QueuePublish using  newItem->topic = _strdup(topic);

I'd rather do something like that:
1.  make fields struct fixed size
typedef struct MqttPublishItem
{
  char topic[64];
  char channel[64];
  char value[64];
  int flags;
  struct MqttPublishItem *next;
} MqttPublishItem_t;
2. allocate MqttPublishItem_t only if there are no "empty" publishitems (use malloc)
3. after using MqttPublishItem, just set its topic[0] = 0 to mark as empty
4. I'd also maybe limit a total number of possible publishitems in queue
This is because for long running devices it's more important to avoid memory fragmentation than memory usage

Why call   MQTT_PublishWholeDeviceState();  from  MQTT_QueuePublish?

Was it tested on both T and N? BK7231N?

iprak:
I have pushed some fixes in https://github.com/openshwprojects/OpenBK7231T_App/pull/183.

* I did change that line to use _strdup in https://github.com/openshwprojects/OpenBK7231T_App/pull/188/commits/33f60314765949bdce5971c9ca437824dc0a29e5
* The MQTT_PublishWholeDeviceState call just forces MQTT publish the very next second. Yes that was unnecessary and have created a separate flag just for queued items.
* I have limited queue length to 5.
* This was primarily tested on a T device.

I considered fixed sizes at one point but I was concerned that I might end up reserving memory. In my testing I am seeing JSON close to 300. PWM devices adds around 100 more char content so the value can be 480, so total would be 608 (64+64+480) per queue entry. If we persisted and reused the queue then would this reserved space be a concern? In my 7231T device, the log tab shows consistent freeheap of around 103496, so maybe the max reserved space (608*5 = 3040) would not be a concern. Thoughts?

iprak:
I limited the queue size to 5 but am wondering how many switches/PWM entities have you ever seen on a device? If that number is more than 5 then discovery will always be incomplete.

iprak:
I did some testing with persistent queue storage. Here you can see 103496 drop to 102872 once 2 messages were sent and then stayed there.  In contrast, the freeheap size bounced back if I used my current dynamic approach.

```
Info:MAIN:Time 30, free 103496, MQTT 1, bWifi 1, secondsWithNoPing 0, socks 2/38
Info:GEN:dhcp=0 ip=192.168.1.211 gate=192.168.1.2 mask=255.255.255.0 mac=c8:47:8c:00:00:00 
Debug:HTTP:TCP will process packet of len 344

Info:MQTT:MQTT_QueuePublish queued message for dd/switch/enbrighten1_8C000000_relay_1/config 1 items queued
Info:MQTT:MQTT_QueuePublish queued message for dd/switch/enbrighten1_8C000000_relay_2/config 2 items queued
Debug:HTTP:TCP will process packet of len 325
Debug:HTTP:TCP will process packet of len 325

Info:MQTT:Publishing to dd/switch/enbrighten1_8C000000_relay_1/config retain=1
Info:MAIN:Time 31, free 102320, MQTT 1, bWifi 1, secondsWithNoPing 0, socks 2/38

Debug:HTTP:TCP will process packet of len 325

Info:MAIN:Time 32, free 98184, MQTT 1, bWifi 1, secondsWithNoPing 0, socks 3/38
Info:MAIN:Time 33, free 102872, MQTT 1, bWifi 1, secondsWithNoPing 0, socks 2/38
Info:MAIN:Time 34, free 102872, MQTT 1, bWifi 1, secondsWithNoPing 0, socks 2/38
Info:MAIN:Time 35, free 102872, MQTT 1, bWifi 1, secondsWithNoPing 0, socks 2/38
```

openshwprojects:
@iprak  can you tell me if you test on N, or on T, or on both? It would be important to also test on BK7231N

@iprak  regarding the "don't use malloc"/"don't use strdup"/"don't use free":
this is because some platforms (w800 maybe?) has primitive heap allocation algorithm:
https://www.freertos.org/a00111.html
image
image
the heap2 is default on w800 (I think) and heap4 is not ported there. I think so, it was so last time I checked.

iprak:
Ok. 
* I will switch to a pre-allocated structure. 
* I am going to use **7** as the max queue size. I noted many devices with 5 PWMs or 5 relays and 7 would give some buffer,

I regularly test on T with more liberal settings since I can UART re-flash it easily.
I only test N after I have made substantial changes. I will test it on N after I have changed to preallocated queue.

iprak:
@openshwprojects That information about heap internals was very useful. 

I have now implemented a fixed content queue, where items get reused as you suggested. The max queue length is 7. I also included checks in case topic/channel/value exceed the space available.

I tested more extensively on a T device with 3 relays. I did test the build on N device as well but with only 1 relay.

openshwprojects:
One more little hint.
I'd say that's acceptable to start with empty queue, and then alloc new item once when needed, then reuse it, and when you need a second item and the same time, you can also do alloc.
You don't really have to alloc all 7 items at once. 
But it's just a general hint, it's not necessary in this case

iprak:
> One more little hint. I'd say that's acceptable to start with empty queue, and then alloc new item once when needed, then reuse it, and when you need a second item and the same time, you can also do alloc. You don't really have to alloc all 7 items at once. But it's just a general hint, it's not necessary in this case

That is exactly how the implementation, sorry my commit message "Fixed queue size" is misleading.

iprak:
@openshwprojects  This request is also ready to go. The real changes are dated Sep 11, 2022.

openshwprojects:
@iprak  merged, let's see how it performs

Issue #184, opened 2022-09-04
MQTT Client ID setting is not used, it only uses the short name


softplus:
I was a bit confused why the names config setting mentions MQTT, and checked the code. It looks like it uses `shortDeviceName` as client-ID for MQTT instead of `mqtt_brokerName` (which should probably be renamed to `mqtt_client_id` too). 

I'll dig a bit more and submit a PR for a one-time transfer of the old setting to the new one, and renaming the variables / functions.

iprak:
So are you proposing just getting rid of `shortDeviceName` with perhaps some fall back generated id?

softplus:
No, I don't know where all it's currently used. My guess is it could be changed into "hostname", but I haven't looked into where all it's used. 

I've dug into the MQTT ClientID part a bit now, and that looks like an actual issue, not just a semantic one (also the naming within the code is a bit organic, I've cleaned that up a bit). Since the MQTT part is separate from the hostname / wifi-settings part, I suggest doing a shortname cleanup separtely.

iprak:
I was just about to submit a PR for fixing the MQTT setup page for https://github.com/openshwprojects/OpenBK7231T_App/issues/185. It would be a small adjustment to use whatever new field you come up with.

softplus:
I think I have it, it's fun figuring these things out :)

openshwprojects:
I will review this change and most likely accept it https://github.com/openshwprojects/OpenBK7231T_App/pull/189 but can I ask, @iprak , will you adjust your changes to work with that?

iprak:
Yes I will adjust my changes.

softplus:
With #189 merged, I think this is done. Thanks!

PR #189, opened 2022-09-05
Fix MQTT Client-Id usage (don't use shortDeviceName)

softplus:
Addresses https://github.com/openshwprojects/OpenBK7231T_App/issues/184 

It turns out the previous MQTT Client setting was never used, and instead the short device name was used. This PR fixes that. Existing MQTT integrations are not affected since the previous values are passed on. The old MQTT Client setting is overwritten, but it was only used in the UI (not for functionality). 

# Details

* Renames MQTT `BrokerName` to `ClientId` (Config variable name, function names)
* Moved config value of `shortDeviceName` to `mqtt_clientId` as a one-time migration (this overrides any previously set Client-ID, but since the MQTT Client-ID was never used, that doesn't change functionality)
* Incremented config version number to 3 to track value change
* Updated all MQTT usage to use the MQTT Client Id (through the function), away from `shortDeviceName`
* Set default MQTT Client-ID to also use the MAC-address based identifier
* Renamed variables in functions to be consistent about MQTT Client ID & short device name (instead of `baseName`, `topicName`, etc). Makes code more readable.
* Updated warning message regarding changing of shortDeviceName in the menu

# Testing

Anecdotally, it works for me, both from the MQTT server as well as to the MQTT server. I'm not really sure how to test all of the MQTT functionality without a ton of manual work...


openshwprojects:
There was indeed just a placeholder "test" value placed in config and the short name was used instead. I think it won't break anything. Can you fix the conflicts and then I will merge that?

Regarding the testing.... I don't think that you need to test all possible MQTT features after that change. The MQTT publish stuff is on different layer of abstraction, so if MQTT works after that change, then all should work the same.

There is another thing that needs to be done, it's a support of proper Tasmota-style MQTT topic syntax creation like:
image
or at least some kind of equivalent to that

PR #194, opened 2022-09-11
Fix bk7231 PWM and cw mode not dimming channel 3 and add cfg option for 600hz pwm

DSchndr:


openshwprojects:
it seems like a very nice change, I will merge it, but was it checked on both N or T or do you need help with testing?
Also why:
	_uint32_t period = (26000000 / pwmfrequency); //TODO: Move to global variable and set in init func so it does not have to be recalculated every time...
	uint32_t duty = (value / 100.0 * period); //No need to use upscaled variable_
Why 26000000 ?
26000000  / 600 = 43333.3333 and var type is integer so you will slightly lose precision but obviously at this frequency no one will notice

DSchndr:
Hehe :)
Was as confused as you xD

I looked at the tuya code & code from others...
Look at how the pwm module gets initialized here:
https://github.com/kuba2k2/libretuya/blob/master/arduino/beken-72xx/cores/arduino/wiring_analog.c

Notice the:
PWM_CLK_26M -> 26000000Hz the PWM module runs on

Seems like the PWM Module is like a fixed timer from the 26m clock and you have to calculate the period / duty from the fixed clock (up to where it should count and how long it should stay on if I understood correctly)

Much wierd since we have no PDF telling what gets showed in the registers and how it works :P
I don't think we have to be ultra precise since we upscale the duty cycle anyways and tuya devs did it the same way :P

I checked on both N and T (LSC Up-Down light and that CW bulb), worked on both without issues but would be nice if more users could test it besides me and [pretoriano80](https://github.com/pretoriano80) :)

DSchndr:
Also, remembered rn

Tuya code has disabled pwm and set pins low when the lamp was set to 0% / off to save energy (according to the comments)
Might be something worthwhile implementing at some time?

openshwprojects:
Hm ok, I will merge this in and also check on my devices.

_Tuya code has disabled pwm and set pins low when the lamp was set to 0% / off to save energy (according to the comments)
Might be something worthwhile implementing at some time?_
You can do that change as well if you want, but it might require some changes in the "pins" source code, not in HAL. Still, it shouldn't be too hard.
Would you be able to do measurements to see if it really saves current?

PR #213, opened 2022-09-25
Remove old, breaking debug code in ConnectToWifi

softplus:
HAL_ConnectToWiFi() in src/hal/bk7231/hal_wifi_bk7231.c was added with some debug code enabled, which forced connection to a specific Wifi AP MAC address, rather than using the SSID information from the configs. I removed this block - now I can get it to connect normally. 

(I'm not quite sure how execution ends up - it was initially not a problem. I noticed it when syncing to the newer BK7231N SDK.)

### DIFF
--- a/src/hal/bk7231/hal_wifi_bk7231.c
+++ b/src/hal/bk7231/hal_wifi_bk7231.c
@@ -236,23 +236,6 @@ void HAL_ConnectToWiFi(const char *oob_ssid,const char *connect_key)
 {
 	g_bOpenAccessPointMode = 0;
 
-#if 1
-	network_InitTypeDef_adv_st	wNetConfigAdv;
-
-	os_memset( &wNetConfigAdv, 0x0, sizeof(network_InitTypeDef_adv_st) );
-
-	os_strcpy((char*)wNetConfigAdv.ap_info.ssid, oob_ssid);
-	hwaddr_aton("48:ee:0c:48:93:12", (unsigned char *)wNetConfigAdv.ap_info.bssid);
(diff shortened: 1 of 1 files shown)


openshwprojects:
This is interesting, I on other hand remember that somehow we had issues with the second version of code (the disabled one).
Can anyone check out this change first and tell me if it works for them?
@iprak  @btsimonh  ? 
but maybe my memory is wrong, this seems like a correct fix. I will check this out later today.

softplus:
The main problem I see is that it sets a fixed BSSID, a fixed security level, and a fixed wifi channel -- this will break on all other wifi access points.

openshwprojects:
I know it sets a fixed BSSID. I suspect it was enabled because during very early days of OpenBeken we didn't have a WiFi connect delay and starting WiFi connection to given Access Point just after reboot was breaking things (crash?) and accidentally that code with fixed SSID was fixing stuff because first connect attempt at fixed SSID was failing thus WiFi start had a larger delay...

I'm not sure, it was a long, long time ago when even @btsimonh was an active contributor.

As said above, I will merge this pull request tomorrow. Nice spotting!

Btw, would you also have time to make the console commands return value consistent?

iprak:
Thanks. I will put in a separate request for the console command.

iprak:
The code in question seems to be coming from the demo samples for N so maybe it used and then forgotten.

openshwprojects:
As I said - I think the code was enabled because it was causing first connection attempt to fail, thus it was preventing from "initialized wifi too early" crash. All by accident.

Nice catch, let's see how well it performs in release.

PR #214, opened 2022-09-25
Update BK7231T build info

Pieter12345:
Elaborate on necessary steps to make the build process easier for new users and contributors. The side note about whitespaces not being allowed on the Cygwin installation folder is because a python script somewhere in the build is not properly constructed, parsing paths with whitespaces as two separate arguments.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,44 +1,43 @@
-
-
 # Building for BK7231T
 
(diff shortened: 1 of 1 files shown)


openshwprojects:
Thanks, this indeed can be helpful for beginners

PR #216, opened 2022-09-26
automatically turn off the bulb when brightness is below 1, turn on w…

Refuhr:
…hen above 1
I really liked the feature in Tasmota and thought it would be practical here as well. What do you think of this implementation and do you even want this feature?


openshwprojects:
Idea is good, but it seems you are doing it incorrectly.
You are just setting g_led_enableAll to 0, but you do not broadcast it's change over MQTT.
Is this intended?

Refuhr:
It wasn't intended, missed that one. I think I got it working now by running the "LED_SendEnableAllState()" function.

Refuhr:
Now it runs the LED_SendEnableAllState() function every time the brightness changes. I could also implement a check that compares a last brightness value with the current one and if it sees a difference it would send a MQTT update. This would stop an unnecessary MQTT update when it's still on.

openshwprojects:
Hmm it seems better. Do you think this behaviour should be always on, or will we add a flag for that? With CFG_HasFlag or whatever it was called....

Refuhr:
I now added a flag "[LED] always turn on bulb when color changes" (OBK_FLAG_LED_ON_CHANGE_TURN_ON) that when checked turns the lamp on when: color/temperature/brightness is changed. One feature I put on at default is turning the lamp off when brightness is 0 and once the brightness is above 0 turning the lamp on once. So it wouldn't turn on when changing the brightness from 20 to 30 for example (Can also be changed, but I thought it made sense. + HA only sets the brightness to 1 at minimum, so it wouldn't trigger turning it off.).

Refuhr:
I might have destroyed your indentation, working on it.

openshwprojects:
Current version doesn't seem to build?

Refuhr:
Don't know what happened there, I reset to an older commit.

openshwprojects:
There are still some things that are needed to be fixed.
https://github.com/openshwprojects/OpenBK7231T_App/pull/216/commits/8789093124512df8cdf38f833e6ce2cb9dfe74bc
In this commit, you are using "char * last_led_color = "#ffffff";" and trying to compare it with given args string "c != last_led_color &&", but it will never work correctly because in C language you are just comparing the pointer values that way (addresses in the memory). Futhermore, you are not writing to "last_led_color " so I think that "last_led_color " can be removed.

Refuhr:
Removed last_led_color and the check. Tested and seems to be working.

Refuhr:
I now updated my branch to the newest and implemented it again. To conclude: I added a flag ([LED] always turn on bulb when color changes), which always turns on the bulb when there is any state change (color/temperature/brightness). When the flag is turned off, the light turns only off when the brightness is below 1 and only turns on once when the last brightness is 0 and the current is above 0.
I tested again with my lamp and everything seems to be working.

openshwprojects:
I'll recheck this tomorrow and most likely merge it then. Thanks

openshwprojects:
I am not sure if the "basecolor" command can turn on the led. Can Home Assistant send a "basecolor" command to bulbs on reboot, even if they are turned off?

Refuhr:
For me, when the lights are turned off, I am only able to either change the brightness or use the on/off switch. The other controls are not even showed.

beren12:
I believe tasmota has this on as a default since it makes logical sense but you can disable it to do things like change the color of a light without turning it on so it will be set when you do turn it on, like if you have lights simulate sunrise or go cooler temp during the day then warm at night.

PR #217, opened 2022-09-26
Quick Config template for '[BK7231T] LSC Smart Connect Bulb 806lm'

pkoretic:


openshwprojects:
Thanks, but remember, now we are mostly using the online templates database:

pkoretic:
Ah, that's great. I was wondering why there was no such system :)
Sorry for that, let me push new pr.

PR #223, opened 2022-09-28
BP1658(CJ) support

Refuhr:
I recently bought a [Hama GU10 RGBCW 5,5W Lamp](https://www.hama.com/00176598/hama-wlan-led-lampe-gu10-5-5w-rgbw-dimmbar-refl-fuer-sprach-app-steuerung#article-details) , which incorporates the tuya [WBLC5](https://developer.tuya.com/en/docs/iot/wblc5-module-datasheet?id=K9duilns1f3gi) module and a BP1658(CJ) I2C LED driver IC (sadly no datasheet found) at P24 (DAT) and P26 (CLK). 
The delay from the BP5758 was a bit to slow for this chip, so I used 10 nops as a delay. Maybe it could be improved there.

### DIFF
--- a/src/cmnds/cmd_newLEDDriver.c
+++ b/src/cmnds/cmd_newLEDDriver.c
@@ -75,9 +75,9 @@ float led_temperature_current = 0;
 
 int isCWMode() {
 	int pwmCount;
-	
+
 	pwmCount = PIN_CountPinsWithRoleOrRole(IOR_PWM, IOR_PWM_n);
-	
+
 	if(pwmCount == 2)
 		return 1;
 	return 0;
@@ -85,7 +85,7 @@ int isCWMode() {
 
 int shouldSendRGB() {
 	int pwmCount;
-	
+
 	// forced RGBCW means 'send rgb'
 	// This flag should be set for SM2315 and BP5758
 	// This flag also could be used for dummy Device Groups driver-module
@@ -93,7 +93,7 @@ int shouldSendRGB() {
 		return 1;
 
 	pwmCount = PIN_CountPinsWithRoleOrRole(IOR_PWM, IOR_PWM_n);
-	
+
 	// single colors and CW don't send rgb
 	if(pwmCount <= 2)
 		return 0;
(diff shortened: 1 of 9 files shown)


openshwprojects:
Thanks, seems like a very good job.
I had a brief glance at code, because I wondered if it's the same chip as BP5758 and if could be **put in the same file**, but it seems that some of I2C procedures were modified by you, so most likely it's ok to have it separate.

I will review it more carefully tomorrow or so and then merge it.
PS: can you submit a short review/teardown of this bulb to our forums and to templates database?
https://www.elektroda.com/rtvforum/forum507.html

Refuhr:
Template is done, working on the review/teardown :).

openshwprojects:
Thanks. I will already merge this pull request, I think it's well done. I am unable to test it in practice myself because I don't have that specific LED driver chip.

esnet146:
My led lamp with bp1658 driver works. rgb colors are mixed up (corrected in home assistant's yaml) and reverse cold - warm.  SetOption37=1 for tasmota does not work.

Refuhr:
Mapping works via the BP1658CJ_Map command. The color mapping for my bulb is [Green][Red][Blue][Warm][Cold] instead of RGBCW. If for example Cold and Warm are mixed do: BP1658CJ_Map 0 1 2 3 4. It should fix your problem. But if after that there are still colors mixed up, just change the order for example if Green and Red are mixed up use: BP1658CJ_Map 1 0 2 3 4. Hope that helps.

esnet146:
Thanks. Startup commands BP1658CJ_Map 2 1 0 4 3

Refuhr:
You don't even have to put it as a startup command, you can just open the webapp and execute it there once. It saves your mapping in config, so it is still configured after a reboot.

Refuhr:
@openshwprojects I just noticed that you can't execute the BP1658CJ_Map command on the {ip-address}/cmd_tool site (it only says Command not found). Did I screw up there?

esnet146:
Yes, such a problem in command. and it's not in the list of commands. I have already found a command window in a web application. But if you fix it, it will be very good.

PR #230, opened 2022-10-01
Added quick config for Zemismart Light Switch (Neutral Optional) KS-8…

zorruno:
…11-3

ZemismartLight Switch (Neutral Optional) KS-811-3
https://www.aliexpress.com/item/4000979779573.html Has a BK7231N direct on board, but easy access to UART. There is a single and 2 switch version of this, which should be easy to template.
Device Info, Pics and Flashing Notes: https://zorruno.com/2022/zemismart-ks-811-with-openbk7231n-openbeken/

### DIFF
--- a/src/new_builtin_devices.c
+++ b/src/new_builtin_devices.c
@@ -833,3 +833,37 @@ void Setup_Device_Aubess_Mini_Smart_Switch_16A() {
 
 	CFG_Save_SetupTimer();
 }
+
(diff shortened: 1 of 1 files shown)


openshwprojects:
Thanks, but new way is to add device here:
also post a teardown article on Elektroda, if possible.

zorruno:
thanks - yes struggled to find documentation as to how to add a device template, and this was the first place in the code I saw it.  Thanks for your work on this.

PR #231, opened 2022-10-01
Defined USER_SW_VER

iprak:
@openshwprojects 
I pushed in a some fixes into to get Linux build working (https://github.com/iprak/OpenW600).  This request jut has a small fix to assign test USER_SW_VER. I also verified that the img can now be uploaded via direct ota. 

I am not sure what is the process to change build actions to build the firmware officially. But you might want to test the build on your device first.


openshwprojects:
Ah, that reminds me, the same needs to be done for BL602. Merged.

PR #233, opened 2022-10-02
W600 fixes

iprak:
Pulled it updated submodule


iprak:
I forgot that changed to W600 sdk won't get pulled in until the submodule link is updated.

PR #240, opened 2022-10-02
Add some buffer overflow protection for logging (& some cleanup) 

softplus:
Just some cleaning up in the logging function

* use strncpy() instead of strcpy() 
* use vsnprintf() instead of vsprintf() 
* move static char tmp[1024]; into the individual functions for cleaner code
* filter out double /r/n in logging (strip from original if exists, add once back)

### DIFF
--- a/src/logging/logging.c
+++ b/src/logging/logging.c
@@ -8,7 +8,6 @@
 // Commands register, execution API and cmd tokenizer
 #include "../cmnds/cmd_public.h"
 
-static char tmp[1024];
 int loglevel = 4; // default to info
 unsigned int logfeatures = (
     (1 << 0) |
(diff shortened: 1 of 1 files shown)


openshwprojects:
After a brief review, I think it's good and safe to merge. Still, please double check everything after updating devices. Thanks.

openshwprojects:
@softplus this pull request broke OTA because of moving 1024 buffer to log function causing the stack memory to run out, but it's already fixed. Please check all systems twice while doing a pull request.

softplus:
@openshwprojects  Woops - sorry about that! Do you have some steps that you use for testing? Does it show the used memory during the compilation process or somewhere? 

(I had this set up on my BK7231N devices and it seemed to work for me, but I definitely didn't try everything..)

iprak:
@softplus Did this change how logging happens? I now see lines merged together making it difficult to read.

Also I don't get what does `MQTT 1(1)` represent.

image

softplus:
@iprak yes, it removed any double /r/n from the log entries. I noticed there's a bug in the web-app that causes them to run on like this only when it's streaming (not when it pulls from the buffer). I'll submit a pr for that.

Issue #221, opened 2022-09-28
Some Daybetter 800lm 9W RGB+2700-6500K bulbs don't turn off fully


dwildstr:
I have OpenBeken flashed onto Daybetter RGBCT lights; these have a BK7231N chipset and use a BP5758D decoder on the light PCB, and a module setup with BP5758D_CLK on pin 7 and BP5758D_DAT on pin 8 works mostly well. It also needed the RGB channels remapped, thought --- blue is channel 0 and red channel 2. Sometimes, however, the lights exhibit a "leaky" behavior when turned off, and slowly brighten back up from dark up to maybe 10% brightness. This behavior persists for long enough that I'm reasonably confident it's not residual charge (and it stops promptly if I cut power to the bulbs). This happens regardless of the bulbs' color (or white temperature) and the color/temperature used before powering off is the one that appears after. It also happens regardless of initial brightness; indeed, if I turn the bulbs off from a very low brightness, the post-turn-off "creep" can end up brighter than it was before being turned off. With a few power cycles the behavior often seems to go away, although it returns when the bulbs are re-flashed with the most recent firmware.

It seems possible this might be at least in part a hardware issue, but there might be a way to control it with the firmware?

DSchndr:
Same issue as #175
Bug is probably related to the code only enabling sleep mode but not setting outputs to off, it makes sense when one refers to the datasheet
![image](https://user-images.githubusercontent.com/19175445/193464004-d5e8072b-d7a8-4950-8f06-118c01796616.png)

@dwildstr Would you mind flashing a test firmware over ota?

dwildstr:
>

dwildstr:
Oof, misclick. Sorry about that. I would be happy to flash a test firmware!

DSchndr:
[BP5758D_Fix.zip](https://github.com/openshwprojects/OpenBK7231T_App/files/9693684/BP5758D_Fix.zip)

Go into webapp, then ota, upload the rbl and test; report on how it behaves :)

dwildstr:
Looks like lights running this firmware turn fully off immediately. I can stress-test it a bit, try it on some of the lights which have been particularly stubborn about never turning off correctly, but early indications are good.

dwildstr:
After a bit more experimentation:

[...]

Issue #175, opened 2022-08-25
BP5758D + BK7231N Bugs


jmacato:
So i have a generic Tuya RGBCW lamp with the chips abovementioned, the hacking/flashing part worked brilliantly but i encountered three problems:

1) Setting the channel mapping on the initial configuration website doesnt seem to work (The values gets set to 40 0 40 0 etc, i suspect that there could be some url encoding issues?)

2) Setting colors with BP5758D_RGBCW is staggerred in the channels (blue turns on then red with around 200ms delay per each channel, etc.. instead of just instantly changing.

3) Setting colors to black/toggling off the lamp sometimes makes the lamp go slightly dim then goes brighter gradually after a minute, plateauing into a very unpleasant brightness.

openshwprojects:
I will try to look into it, but latest release (as of yesterday) runs fine on my BK7231N:
![image](https://user-images.githubusercontent.com/85486843/186837741-c36afb8e-b06e-48bd-bf6f-da0346725033.png)

Do you have a UART flash circuit still connected?

jmacato:
@openshwprojects hmm i was not able to do a direct uart flashing of the devices, only OTA via tuya-cloudcutter.. Should i try to update to the newest firmware?

jmacato:
<img width="443" alt="Screen Shot 2022-08-26 at 3 49 41 PM" src="https://user-images.githubusercontent.com/16554748/186851209-7e8b5569-3a13-426f-8d0b-8b4a74d6f9ec.png">

This is the current

jmacato:
tested on latest firmware, still the same 3 issues

openshwprojects:
can you show any more screenshots of the issues?

jmacato:
Issue 1 : 
![Screenshot from 2022-08-26 22-18-17](https://user-images.githubusercontent.com/16554748/186924829-314c62d8-1576-4681-9905-60fe52ccc742.png)

Issue 2:
https://user-images.githubusercontent.com/16554748/186924930-9651c4f2-ecbf-4eb3-a020-67a78bb3c91d.MOV

Issue 3:
https://user-images.githubusercontent.com/16554748/186925187-e48a69bc-c9b8-48ca-8f1a-9e8f59e84c75.MOV

jmacato:
@openshwprojects do tell if you need more info to help on resolving these bugs

openshwprojects:
"out of range" stuff fixed in:
[...]

PR #242, opened 2022-10-03
[drv_bp5758d] Fix sleep mode (or "dim up" issue)

DSchndr:
Reason for the bug and test can be seen in #221


jmacato:
partially addresses https://github.com/openshwprojects/OpenBK7231T_App/issues/175 (Issue no. 3)

openshwprojects:
Thanks. I am unable to test it now, but it seems to be a correct fix.
@jmacato does issue no 2 still persists?

jmacato:
@openshwprojects sadly yes it still persists but at least this major issue is fixed by @DSchndr 🙏

PR #244, opened 2022-10-04
added linear mapping function to map 0-255 to 0-1023

Refuhr:
I noticed that the 8 to 10 bit conversion is just a simple multiplication by 4. 
```c
cur_col_10[i] = rgbcw[g_channelOrder[i]] * 4;
```
I now added a simple linear mapping function to improve the 8 to 10 bit mapping:
```c
// Arduino mapping function
cur_col_10[i] = (rgbcw[g_channelOrder[i]] - 0x00) * (0x3FF - 0x00) / (0xFF - 0x00) + 0x00;
```

### DIFF
--- a/src/driver/drv_bp1658cj.c
+++ b/src/driver/drv_bp1658cj.c
@@ -83,10 +83,9 @@ void BP1658CJ_Write(byte *rgbcw) {
 	for(int i = 0; i < 5; i++){
 		// convert 0-255 to 0-1023
 		//cur_col_10[i] = rgbcw[g_channelOrder[i]] * 4;
-    // Arduino mapping function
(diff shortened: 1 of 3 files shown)


openshwprojects:
May I know what it exactly changes?

Refuhr:
Essentially it just changes the factor from 4 to (1023/255) and makes it so that you get the full output range. I could put it in header or just write (1023/255).

openshwprojects:
maybe better make it project-wide and put in new_common?

Refuhr:
I put it in new_common.h and tested it.

openshwprojects:
Thanks!

PR #245, opened 2022-10-05
Hass discovery/Json memory fixes

iprak:
Various fixes which were crippling W600 device.
* Fixed cJSON_Delete to nuke child nodes after they have been deleted.  cJSON_Delete recurses to clean up children but was not setting them to NULL. The caller (hass.c) was doing cleanup on child nodes in a NULL check and that caused crashes .. but only in W600.
* Defined constants for various configuration fields and converted discovery structure to use fixed size array based on those constants. This got rid of some malloc/free operations.
* Added logging for HASS.
* Discovery payload was always 1/0 and not the passed values which might be problem later.

This seems to resolve HASS discovery crashes on W600. The MQTT buffers in LWIP definitely helped as well.

### DIFF
--- a/src/cJSON/cJSON.c
+++ b/src/cJSON/cJSON.c
@@ -259,6 +259,7 @@ CJSON_PUBLIC(void) cJSON_Delete(cJSON *item)
         if (!(item->type & cJSON_IsReference) && (item->child != NULL))
         {
             cJSON_Delete(item->child);
+            item->child = NULL;
         }
         if (!(item->type & cJSON_IsReference) && (item->valuestring != NULL))
         {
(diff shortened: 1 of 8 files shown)


openshwprojects:
I wonder if the JSON code has to be so, so mallocy/freecy, so many allocations... but since it's called only once when doing discovery I'll let it be for now. Good job!

iprak:
> I wonder if the JSON code has to be so, so mallocy/freecy, so many allocations... but since it's called only once when doing discovery I'll let it be for now. Good job!

Yup. The `cJSON_PrintPreallocated` API worked out well for JSON string generation but there isn't such API for building the structure. I think the outcome is quite positive. If we ever need to reclaim that memory, we could consider doing an automatic restart.

PR #247, opened 2022-10-05
Hass discovery support for  RGB

iprak:
This is not done yet but I wanted to get feedback.

The discovery additions for RGB are based on `http_fn_ha_cfg`.

Changes
* Relay and PWM names now include `relay` or `light` to differentiate between them e.g. `w600CD2185A3 relay 1`. 
* RGB name does not need to include `index`. There should be only one RGB entity since that will only happen if all 5 PWMs were used or external driver was being used, right? 
* The hass config was appending 64 due to the relay loop before it. This has been fixed as well.

Issues
* Currently, PWM and relay have the same state_topic/command_topic (`w600Test/1/set`) which won't work. Instead, should it be  `w600Test/relay_1/set`? To avoid breaking existing settings, `w600Test/1/set` can still drive the relay.
* RGB: I think the topics are incorrect, shouldn't they start with `clientId` and not with `cmnd` ?
* HomeAssistant doesn't like rgb_command_template `hprintf128(request,"       rgb_command_template: \"{{ '#%02x%02x%02x0000' | format(red, green, blue)}}\"\n");` and throws

```
2022-10-05 06:05:16.093 ERROR (MainThread) [homeassistant.util.logging] Exception in rgb_received when handling msg on 'cmnd/w600Test/led_basecolor_rgb': '#ff80ff0000'
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/mqtt/debug_info.py", line 47, in wrapper
    msg_callback(msg)
  File "/usr/src/homeassistant/homeassistant/components/mqtt/light/schema_basic.py", line 490, in rgb_received
    rgb = _rgbx_received(
  File "/usr/src/homeassistant/homeassistant/components/mqtt/light/schema_basic.py", line 477, in _rgbx_received
    color = tuple(int(val) for val in payload.split(","))
  File "/usr/src/homeassistant/homeassistant/components/mqtt/light/schema_basic.py", line 477, in <genexpr>
    color = tuple(int(val) for val in payload.split(","))
ValueError: invalid literal for int() with base 10: '#ff80ff0000'
```

### DIFF
--- a/src/cmnds/cmd_newLEDDriver.c
+++ b/src/cmnds/cmd_newLEDDriver.c
@@ -259,6 +259,23 @@ int LED_GetMode() {
 	return g_lightMode;
 }
 
+const char *GetLightModeStr(int mode) {
+	if(mode == Light_All)
+		return "all";
+	if(mode == Light_Temperature)
+		return "cw";
+	if(mode == Light_RGB)
+		return "rgb";
+	return "er";
+}
+void SET_LightMode(int newMode) {
+	if(g_lightMode != newMode) {
+        ADDLOG_INFO(LOG_FEATURE_CMD, "Changing LightMode from %s to %s",
+			GetLightModeStr(g_lightMode),
+			GetLightModeStr(newMode));
+		g_lightMode = newMode;
+	}
+}
 OBK_Publish_Result LED_SendCurrentLightMode() {
 
 	if(g_lightMode == Light_Temperature) {
@@ -291,7 +308,8 @@ void LED_SetTemperature(int tmpInteger, bool bApply) {
 	baseColors[4] = (255.0f) * f;
 
 	if(bApply) {
-		g_lightMode = Light_Temperature;
+		// set g_lightMode
+		SET_LightMode(Light_Temperature);
 		sendTemperatureChange();
 		apply_smart_light();
 	}
@@ -393,7 +411,7 @@ static int dimmer(const void *context, const char *cmd, const char *args, int cm
 	//return 0;
 }
 void LED_SetFinalRGB(byte r, byte g, byte b) {
-	g_lightMode = Light_RGB;
+	SET_LightMode(Light_RGB);
 
 	baseColors[0] = r;
 	baseColors[1] = g;
@@ -425,9 +443,9 @@ int LED_SetBaseColor(const void *context, const char *cmd, const char *args, int
 				c++;
 
 			if(bAll) {
-				g_lightMode = Light_All;
+				SET_LightMode(Light_All);
 			} else {
-				g_lightMode = Light_RGB;
+				SET_LightMode(Light_RGB);
 			}
 
 			g_numBaseColors = 0;
@@ -581,7 +599,7 @@ void NewLED_RestoreSavedStateIfNeeded() {
 		byte rgb[3];
 		byte mod;
 		HAL_FlashVars_ReadLED(&mod, &brig, &tmp, rgb);
-		g_lightMode = mod;
+		SET_LightMode(mod);
 		g_brightness = brig * g_cfg_brightnessMult;
 		LED_SetTemperature(tmp,0);
 		baseColors[0] = rgb[0];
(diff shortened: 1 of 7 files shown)


openshwprojects:
From a quick glance at code seems very good, but now to address your questions...

_Currently, PWM and relay have the same state_topic/command_topic (w600Test/1/set) which won't work. Instead, should it be w600Test/relay_1/set? To avoid breaking existing settings, w600Test/1/set can still drive the relay._
I do not understand why is it a problem. Futhermore, for LED lights, we don't use the w600Test/1/set at all, we use things like led_enableAll, led_basecolor, etc. I would assume that Discovery just have to mimic the Yaml generator.

_RGB name does not need to include index. There should be only one RGB entity since that will only happen if all 5 PWMs were used or external driver was being used, right?_

so far I am working under assumption that each device has only single RGB or RGBCW driver. So no index for RGB.

_HomeAssistant doesn't like rgb_command_template hprintf128(request," rgb_command_template: \"{{ '#%02x%02x%02x0000' | format(red, green, blue)}}\"\n"); and throws_

Again, I do not understand why is this a problem - we already have a working YAML generation for RGBCW lights, here:
https://github.com/openshwprojects/OpenBK7231T_App/commit/2284d0ec97c428eb5422b627211c680f81ff8392
And this works for me.
Is there some difference between Discovery and the Yaml generator?
Or are you saying that the generated yaml is also invalid?

In general, I'd say, just refer to this change:
https://github.com/openshwprojects/OpenBK7231T_App/commit/2284d0ec97c428eb5422b627211c680f81ff8392
Unless, is there something I don't know, and maybe the linked commit above is also partially incorrect?

iprak:
For plain PWM, currently `http_fn_ha_cfg` passes the same `state_topic/command_topic` as relay when pwmCount > 0 (this is the very last block) and maybe it doesn't since we do have  `brightness_command_topic`.

The Hass error actually happens when I pick a color in the UI, no errors happen in actual discovery phase.

I will look more into depth later.

openshwprojects:
"For plain PWM, currently http_fn_ha_cfg passes the same state_topic/command_topic as relay when pwmCount > 0 (this is the very last block) and maybe it doesn't since we do have brightness_command_topic."
image

This is a hacky approach. This is because it's an old code with direct PWM access. It has no support for "enable/disable" or for "brightness mult". This allows HA to write directly to Beken PWMs and bypass everything else. It should not be used anymore, as we have a LED driver frontend with "basecolor_rgb" etc etc stuff.

iprak:
I think I figured out why the changes not being received at Hass end just after writing my previous comment. Based on the serial log

* color_temp_state_topic should be `"%s/led_basecolor_rgb/get",clientId`
* rgb_state_topic should be `%s/led_basecolor_rgb/get",clientId);`
* And RGB does need a state_topic in addition to rgb_state_topic. Its value would be `%s/led_enableAll/get",clientId);`

iprak:
I also noted this .. could that because `hprintf128` is formatting the string?

image

openshwprojects:
Huh, that's a great find, @iprak . I must have missed that. Fix it. % should be escaped as %%.

iprak:
It seems we are both making changes in the same area. I see your commented out brightness_value_template and color_temp_value_template which I too happened to have removed. I had also happened to make adjustments for the rgb_value_template. Are there more changes coming?

iprak:
I think everything is working. I did notice some inconsistencies and write them a separate issues .. maybe you are already know about them.

I had modified cmd_newLEDDriver to set differently styled data to fix the exception at Hass end but I see you added a template to account for that. That works but all that keeps increasing the MQTT size. I have bumped the MQTT storage to 768 from 480.

openshwprojects:
Sorry for the changes. Few days ago I made that mistake with making wrong YAML topics for some features, because I was looking at wrong YAML (posted by some user here =.= ) while implementing the code.
No more changes pending, I commit all my changes just after making them.

Good job on getting RGB CW Discovery running. Feel free to do any more changes if they are needed, but please try to do not change the current topic names and standard as they are already used by people.

Is it ready to merge?

iprak:
Yes, all my changes are in too. I do not have  real RGB but I tested it with my W600 with 5 PWM to simulate RGBCW and then with 3 RGB and then 2 RGB+relay. I wrote a couple issues (some are more of a question) about things which I encountered during dev, I will address any accidental shortcomings in my merge.

PR #261, opened 2022-10-08
Hass / removed duplicate code

iprak:
* Consolidated duplicate strings/code, fixed some build warnings
* Added condiguration_url which appears here in HomeAssistant
image

### DIFF
--- a/src/cmnds/cmd_newLEDDriver.c
+++ b/src/cmnds/cmd_newLEDDriver.c
@@ -231,7 +231,7 @@ static void sendFinalColor() {
 	byte c[3];
 
 	if(shouldSendRGB()==0) {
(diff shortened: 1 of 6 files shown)


openshwprojects:
nice, is it really to roll?

iprak:
yes

openshwprojects:
Thanks. Ok so merging, altough I would just note, that I am not sure if doing #define for each string really helps - any modern compiler should already figure out constants and make them unique.

iprak:
> Thanks. Ok so merging, altough I would just note, that I am not sure if doing #define for each string really helps - any modern compiler should already figure out constants and make them unique.

Thanks I was not completely sure about that.

Issue #259, opened 2022-10-07
Enhancement: Add discovery support for voltage sensor


iprak:
I think we (I) should be able to add discovery support for voltage sensor. I am creating this for tracking purpose.

PR #264, opened 2022-10-08
Hass discovery power

iprak:
Support for voltage/current and power sensors. [#259](https://github.com/openshwprojects/OpenBK7231T_App/issues/259)

* Introduced a test driver which can be activated by setting ENABLE_BL_TEST_DRIVER in obk_config.h
* Also added `stopDriver` command

image


openshwprojects:
Nice, the stub BL test driver is exactly something I would do if I had slightly more time. I like that style of testing.

I will be making console command return values consistent today, so tell me, is it final and ready to merge?

iprak:
Yes

PR #266, opened 2022-10-08
Calls to bk_printf with logging strings need %s

softplus:
bk_printf expects a formatting string, so if you pass it a log entry with a "%" in it, it will not show the "%" symbol and instead try to match a formatting pattern. This causes a difference in logging to serial vs logging to TCP / HTTP. Eg, this code results in different results on serial vs other logging:

```
ADDLOGF_INFO("At %d%%", 12);
```

Expected: Log shows "At 12%" - actual serial log is "At 12".

### DIFF
--- a/src/logging/logging.c
+++ b/src/logging/logging.c
@@ -184,8 +184,7 @@ void addLogAdv(int level, int feature, char *fmt, ...)
             if (tmp[strlen(tmp)-1]=='\n') tmp[strlen(tmp)-1]='\0';
             if (tmp[strlen(tmp)-1]=='\r') tmp[strlen(tmp)-1]='\0';
 
-            bk_printf(tmp);
-            bk_printf("\r\n");
+            bk_printf("%s\r\n", tmp);
             if(g_extraSocketToSendLOG) 
             {
                 send(g_extraSocketToSendLOG,tmp,strlen(tmp),0);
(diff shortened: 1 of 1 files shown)


openshwprojects:
Thanks, please just check if OTA and all stuff is still working, because the latest logging changes broke OTA and now I require all contributors to make sure to test changes on their devices before integrating them into main source tree.

softplus:
I ran it on BK7231N and BK7231T modules for >30 minutes, Wifi + MQTT connected. OTA works, memory usage is same as before.  Entering `hello 5%` into the command line results in the same text being echoed back in the serial, TCP:9000, and REST logs.

openshwprojects:
Ok, thx, that's an interesting fix and it makes sense. I have never noticed a normal print in C exhibit that behaviour, is it beken-specific?

softplus:
@openshwprojects it tries to pull the parameters & does weird things.

```
echo '#include<stdio.h>
void main() { char s[]="test 5% test"; printf(s); }' | gcc -x c -o a.out - && ./a.out

<stdin>: In function ‘main’:
<stdin>:2:47: warning: format not a string literal and no format arguments [-Wformat-security]
test 5 0.000000e+00st
```
or 

```
echo '#include<stdio.h>
> void main() { char s[]="test 5% item"; printf(s); }' | gcc -x c -o a.out - && ./a.out

<stdin>: In function ‘main’:
<stdin>:2:47: warning: format not a string literal and no format arguments [-Wformat-security]
test 5 256549304tem
---

```
echo '#include<stdio.h>
void main() { char s[]="test 5% status"; printf(s); }' | gcc -x c -o a.out - && ./a.out

<stdin>: In function ‘main’:
<stdin>:2:49: warning: format not a string literal and no format arguments [-Wformat-security]
test 5��+8�tatus
```

PR #267, opened 2022-10-09
Mqtt logging cleanup

iprak:
This probably won't help with MQTT connection issues but helps with logging confusion.

* Cleaned up logging to avoid repeated malloc/free on `mqtt_status_message`. We were allocating that to 256 char so I defined it as 256 char array.
* Default mqtt host to be empty.
* Allocate sufficient space for a local variable.

I verified the empty host by doing a full erase which gave this logging
```
Info:MAIN:Time 16, free 30296, MQTT 0(0), bWifi 1, secondsWithNoPing 0, socks 2/8<CR>
Info:MQTT:mqtt_host empty, not starting mqtt<CR>
Info:MAIN:Time 17, free 30296, MQTT 0(1), bWifi 1, secondsWithNoPing 0, socks 2/8<CR>
```

### DIFF
--- a/src/mqtt/new_mqtt.c
+++ b/src/mqtt/new_mqtt.c
@@ -57,7 +57,7 @@ static int g_timeSinceLastMQTTPublish = 0;
 static int mqtt_initialised = 0;
 static int mqtt_connect_events = 0;
 static int mqtt_connect_result = ERR_OK;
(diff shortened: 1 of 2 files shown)


openshwprojects:
I just hope that increasing buffers in the functions won't cause a stack overflow someday

PR #268, opened 2022-10-10
Logging using static buffer

iprak:
Another stability change towards logging. This change seems to resolve the hang in W600. The change is safe and reduces the continuous malloc/free operations which should help in general since we are actually logging once or more within a second.


openshwprojects:
And so the full circle is done.... @softplus  moved buffer to function, causing stack overflow.. @valeklubomir  fixed stack overflow by doing malloc for each log call... and now @iprak , buffer is static again!
Now, now, thanks for the fix, this is something I've done at the start and worked reliably for me.
It's a good change, and now guys, please let that buffer stay static.

PR #270, opened 2022-10-11
Update BL602 build info

Pieter12345:
Elaborate on necessary steps to make the build process easier for new users and contributors. It might be possible to build using Cygwin as well, but that fails on a missing `<sdk>/toolchain/riscv/CYGWIN` directory. Using Msys2, the `<sdk>/toolchain/riscv/MSYS` directory can be copied and renamed as described. This is probably something that should be changed in the SDK repo, but as far as the building information goes, this at least allows people to compile binaries without having to figure out these steps.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -38,6 +38,24 @@ Checkout [this repository](https://github.com/openshwprojects/OpenBK7231T_App) t
 Run ./build_oxr_sharedapp.sh
 
 
+# Building for BL602
+Get the SDK repo:
+https://github.com/openshwprojects/OpenBL602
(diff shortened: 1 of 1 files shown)


openshwprojects:
Thanks, altough I seem to think that you had to set some kind of global variable in the environment as well

Pieter12345:
> Thanks, altough I seem to think that you had to set some kind of global variable in the environment as well

I have not done that. I did see that a recent commit fixed a build error for the BL602 build, maybe that's what you remember? Or it might be a Linux specific thing that wasn't necessary on Windows.

PR #275, opened 2022-10-11
Add device Calex Smart Plug 900011.1

btsimonh:
Can't build and test at the moment; pls confirm it builds?


openshwprojects:
Those templates are now kinda obsolete, I'll merge it but I think the way to go isi web app Devices.json

PR #282, opened 2022-10-14
feat: migrate W600 SDK

talltechdude:
Here's a PR for migrating W600 SDK to the @openshwprojects fork if you wanted. Both repos are at same commit currently so no other changes should be required.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,4 +18,4 @@
 	url = https://github.com/openshwprojects/OpenW800.git
 [submodule "sdk/OpenW600"]
 	path = sdk/OpenW600
-	url = https://github.com/iprak/OpenW600.git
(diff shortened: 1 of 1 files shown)


openshwprojects:
@iprak  from now on, for the sake of consistency and to make it easier for new users, the SDK used to build will be my fork. If you have any further changes, just do a pull request, I will accept them happily. Don't worry, I am active almost every day.
@talltechdude thanks

openshwprojects:
@talltechdude would it be possible to include "make bins" step in OpenBL602 auto build?
image
We need those files for OTA:
image

openshwprojects:
wait I think I got it covered!

openshwprojects:
My xz file doesn't get published:
image

PR #283, opened 2022-10-15
Mqtt failure reason

iprak:
* Ensures that fields based on `CFG_GetMQTTClientId()` have more space than 64 which is what is the size of `mqtt_clientId`.
* Also added `MQTT_ready` for future use.
* Added MQTT failure reasons.
image


openshwprojects:
Seems good, the only thing to improve is MQTT_ready -> MQTT_IsReady (the "Is" conversion is important).

PR #285, opened 2022-10-15
Consumption measurement updates.

valeklubomir:
- Energy Consumption measurements updated. 
Updated MQTT topic for energy measurment.
Added caclulation for consumption per last hour, with statictics.
- Updated HA assistent. 
added new topic to configurator.
- Updated HTTP interface.
added new output for energy consumption.
added GPIO monitor for unused pins. Pin usage helper.
- Updated REST interface.
- Added new devices. 
Smart Plug, Touch button.


openshwprojects:
This is something that we really needed and I think I will merge it today, but there is one thing I'd improve.
The energyCounterMinutes array.
I wonder if array of 60 size (one per minute) is precise enough, what do you think?
I was considering something similar, but:
1. make energyCounterMinutes a pointer and alloc it only if user enables a flag in CFG
2. make energyCounterMinutes variable-sized and instead of storing samples per minute, store maybe even a sample per 10 seconds or something like that. Of course, the array size would grow much but if it's allocated only if flag is enabled, then it's okay. We can afford to allocate several kbs or more for that on many devices.

But still, your current version seems very good as well

openshwprojects:
Ah, and one futher note, before we merge.
There is a serious and unexpected issue with:
"ed current %d, voltage %d, power %d, f" on N platform. As far as I know, trying to directly logPrintf a floating number of N platform crashes.
This is why, on N platform, I am doing first a sprintf (for floating numbers) to a buffer, and then a addLogPrintf. 

I know this is very strange, and I don't know what's wrong with the VA library on N, but that's how it is so far.

valeklubomir:
Technically finer samples are possible. 
May be making it variable with two dimensions. Command with parameters: setStats <enable><rate><count>.
But sampling rates now differs BL0937 (~1sec), BL0942 (~2sec). 
I have choosen one minute as golden ratio, because I have some experience with such application from photovoltaic systems
and even with max output at 230V/16A one minute collected consumption would be ~61 Wh. 
But it depends with application used. I would still set some limits. >=10s per sample and max. 120/180 samples.
MQTT is sending the stats each time array is shifted. length of JSON struct/string could be issue if sampling precision is too high.

openshwprojects:
Thanks, is it ready to merge?

valeklubomir:
It takes time to create and fully test new features. 
I will leave it to next merge.

openshwprojects:
I meant are the already implemented features ready, but from what I understand, they are. Thx.

PR #276, opened 2022-10-12
Format/compress js/css

iprak:
So this is an attempt at making script/css more developer friendly. Writing css/script in wrapped string fields is not easy and prone to errors. It also addresses outdated "Online for" label.

**High Level Changes**
* Script/css can now be well written with meaningful comments.
* Embedded content is smaller in size, it has been minified.
* Developer will have to use [gulp](https://gulpjs.com/) after modifying js/css.
* The "Online for" is now updated on client.

** Code Changes **
* Embedded script/css have been moved out into separate files under `httpserver` folder. One can edit them, add comments.
* Minified scripts/css get injected back into `new_http.c` within marked regions. This is done using gulp package which uses [nodejs ](https://nodejs.org/en/) sub system. I use VSCode but this can done from command prompt too. After installing nodejs, one would run `npm install` to install necessary packages including gulp. All this is defined in `packages.config`.
image
* The server just sends `Time_getUpTimeSeconds` .. the js function generates the string as previously done in C.

PR #287, opened 2022-10-16
W600 - Accounting for ticks (not the deer ticks)

iprak:
So this had been bothering me since I added support for W600 that the "Online for" did not seem to match the time the device had really be on. This shortcoming was more prominent after #276 where the "Online for" was updated on client and clearly did not match the value in "Info:MAIN".

1) We use g_secondsElapsed as a metric of how long the device has been running. This is not correct in W600 since the mainLoop has `tls_os_time_delay(1000);` and  `tls_os_time_delay` actually takes how many ticks to wait for. 

   Based on code analysis, the same should hold true for W800. I unfortunately don't have W800 device yet.

2) We always wait for 1 second regardless of the time was spent in `Main_OnEverySecond`. I measured that Main_OnEverySecond takes about 5-6 ms when it does something like MQTT connection otherwise it is 1-2. 

I attempted switching to `tls_os_timer_start` and `tls_os_task_create` but the device hung in both case, so clearly there is some underlying dependency. So I decide to just adjust the delay duration to account for the time spent in `Main_OnEverySecond`.

### DIFF
--- a/src/hal/w800/hal_main_w800.c
+++ b/src/hal/w800/hal_main_w800.c
@@ -12,33 +12,34 @@
 
 #if 0
 
-void testprintf(){
+void testprintf() {
 	printf("Test 123\n\r");
 }
 
-static tls_os_timer_t *main_second_timer;
+static tls_os_timer_t* main_second_timer;
 
 void user_main(void)
 {
 	Main_Init();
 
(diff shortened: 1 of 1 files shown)


openshwprojects:
The "online for" on all platforms is not really a good time measurement, you are correct. Still, it seems you improved it for W600. Tahnks

Issue #109, opened 2022-07-05
CSE7766 driver?


belikh:
Hi, is there any chance of including the power monitor IC CSE7766 driver?

I have this device: 
https://templates.blakadder.com/deta_6920HA.html

I've managed to flash the WB2S within and the touch sensors and relays are working, but I'm not sure how to proceed with adding voltage/current sensor.

It seems supported in Tasmota and ESPHome.

(for anyone interested I needed to desolder the WB2S in order to flash.)

openshwprojects:
CSE7766 looks very simple, it uses UART interface, but I would need to get somehow one device for testing.

belikh:
Would remote web access to my device be of any help?    Otherwise I don't mind testing it.   I'm not sure how commonly the CSE7766 is used but we don't have a lot of options in Australia.

openshwprojects:
It would be very helpful, but I would have to have the remote access to openbk panel, I would need to be able to do a firmware update, and there would need to be some kind of load in the socket, maybe 10W bulb or something.

I would also need some photos from inside or at least know which pins are used by CSE7766

belikh:
I'm assuming that the chip is on the 1TX and 1RX pins, as I can't flash with those connected.   

Device is at http://110.151.54.191/ at least for now.  I'm connected via LTE temporarily.

belikh:
![20220707_115055.jpg](https://user-images.githubusercontent.com/4435491/177672970-4a4e7332-2605-44e9-88e9-71f405c6501d.jpg)

openshwprojects:
Ok let me see, but I think w should rather get a Discord or an email conversation

openshwprojects:
working on it

belikh:
Aha I have awoken to find Santa visited :p

I'll leave the port open until you say you don't need it :)

I've turned off the load though until you need it (energy crisis etc)

Thanks again!

openshwprojects:
@belikh  can you explain what exactly is inside this device, because I am not entirely sure? The power metering chip is measuring power consumption of the two sockets together?

belikh:
Correct.  It has one input,  two power outlets and two usb sockets.   I'm not sure if it's measuring usb also.  

It's a pretty tightly packed pcb and hard to tell what's going on.

I've got another I'll take some photos

openshwprojects:
@belikh  what is the voltage in your country, the exact measurement would be welcome...

Also, have you read that guide?
[...]

PR #289, opened 2022-10-16
fix: cse divide by zero and calibration factors

talltechdude:
Fixes for CSE7766 driver. I got my hands on one of the Deta Double GPOs with power monitoring similar to https://github.com/openshwprojects/OpenBK7231T_App/issues/109 so I can help with the further calibration/testing of this device and driver if required.

It looks like since https://github.com/openshwprojects/OpenBK7231T_App/commit/ce6f0aa37108b2c401c846ff876065c8acd98ff2 there are some divide by zeros happening (power usage shows as `inf`) and the calibration numbers aren't being returned from `VoltageSet` etc properly (and they are also large numbers now!). By backlog command now looks like this:
```
backlog startDriver CSE7766; VREF 383160; IREF 16520; PREF 10444800;
```

Calibrated @ 240V with 2400W / 10A resistive load.


openshwprojects:
Thanks. I will accept this merge right now as I trust your judgement.

As per CSE, the story is that I did CSE driver remotely for one user and was not able to finalize it as that user gone missing.... I was doing CSE driver on device in Australia while I'm at Poland, so that was a real remote job.

PR #290, opened 2022-10-16
chore: new built-in device for Deta GPO

talltechdude:
Device template for power monitoring edition of the DETA dual GPO with USB ports (https://www.bunnings.com.au/deta-grid-connect-smart-double-touch-power-point-with-dual-usb_p0172780)

Enables buttons linked to relays, WiFi LED and starts CSE7766 driver for power monitoring.


openshwprojects:
Thanks, but we're migrating to: https://openbekeniot.github.io/webapp/examples/devicesListDemo2.html
The new procedure is to post a detailed teardown here: https://www.elektroda.com/rtvforum/forum507.html
And update this list: https://github.com/OpenBekenIOT/webapp/blob/gh-pages/devices.json

but still, it's yet ready system, I need few days more to finish the javascript devices.json frontend and then I will update reademe

PR #293, opened 2022-10-16
Consumption measurement - configurable history

valeklubomir:
Power monitor - added configurable history model
Fixed bug: missing call of BL_stared_init
Fixed bug: htprintf128 change to use vsnprintf with buffer limit set to 256 bytes

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,9 @@ There are multiple console commands that allow you to automate your devices. Com
 | CurrentSet     | [Value] | Used for BL0942/BL0937/etc calibration. Refer to BL0937 guide for more info. |
 | DGR_SendPower     | [GroupName][ChannelValues][ChannelsCount] | Sends a POWER message to given Tasmota Device Group with no reliability. Requires no prior setup and can control any group, but won't retransmit. |
 | DGR_SendBrightness     | [GroupName][Brightness] | Sends a Brightness message to given Tasmota Device Group with no reliability. Requires no prior setup and can control any group, but won't retransmit. |
-  
+| EnergyCntReset | | Used for BL0942/BL0937/etc consumption measurement data reset |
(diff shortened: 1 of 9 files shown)


openshwprojects:
Thank you @valeklubomir , this is great. I have just one request - when doing something like "wifi state hack", can you add some comments to code why is such hack done? I am working on MCU as well so it would be good to know what's going on. Just document your changes.

valeklubomir:
I write some comments into code with all changes in the future.

openshwprojects:
@valeklubomir  there might be a problem.
Tested on: Mycket WB2S + BL0937 and also on 4 relays WB3S driver.
If BL0937 is running, the device crashes after a minute.
image
image
image
Reproduction steps:
1. get any BK7231T device (tested on two)
2. type in console: startDriver BL0937
3. NOTE: I have no MQTT connected (if it matters...)
4. Wait. It runs:
image
5. It died:
image
Device is dead, no response from buttons, no stack trace, dead

openshwprojects:
@valeklubomir  just with energy stats disabled both devices seem stable (with BL0937 on), but I will wait to make sure....
I also pushed a test function that can be run in script/timer every second to check if it's a malloc issue..
https://github.com/openshwprojects/OpenBK7231T_App/commit/abe3c6be14aece192d1b64b9e5391111a6d46353
Wait. Calling testJSON console command hangs device.

openshwprojects:
@valeklubomir  that call hangs T platform! 
image

openshwprojects:
@valeklubomir  here it crashes, I am continuing investigation:
image

openshwprojects:
Conclusion: realloc is broken on T

iprak:
I think I might have into something like this which is why I switched to `cJSON_PrintPreallocated`.

PR #297, opened 2022-10-17
Ir tester

btsimonh:
this should give you some ideas?


btsimonh:
will do a new PR which fixes online build as soon as github has confirmed it builds :).

Issue #252, opened 2022-10-06
Quibbles with light bulb autogenerated HA YAML


dwildstr:
I'm glad the feature to autogenerate YAML is working (it used to just have a blank configuration), but I noticed some discrepancies between what it generates and what I've been using on an RGBCW light ---- some of that's certainly down to personal preference in configuration, but there were a few things which seemed like they were mistakes. This is all in lines 1351–1367 of `httpserver/http_fns.c`, although similar issues seem to be in the RGB-only code in 1381–1391.

- all the lines past `name` are indented too far, with 7 spaces instead of 4.
- The key `brightness_value_template` seems unnecessary; purely numeric data shouldn't need processing to be intelligible to the MQTT light platform. The same is true of the `color_temp_value_template` key, which is already commented out.
- There's no `rgb_value_template` key, which seems like it should be necessary, since OpenBeken reports colors as hex values instead of comma-separated triples (which are what Home Assistant likes to get). I use `rgb_value_template: "{{ value[1:3] | int(base=16) }},{{ value[3:5] | int(base=16) }},{{ value[5:7] | int(base=16) }}"` in my configuration to convert it.

dwildstr:
An addendum: The `state_topic` key as currently written causes the bulbs to have their state reported as "unknown" in Home Assistant; removing it seems to make them mostly work (sometimes being a bit out of sync if HA or the bulbs have been rebooted). I  can't figure out why this would be the case. The bulbs do indeed seem to be publishing their state on MQTT correctly. This may be related to some HA issues, see https://github.com/home-assistant/core/issues/67545 and https://github.com/home-assistant/core/issues/74019, but until that's fixed, it might be safer to remove the state_topic line from autoconfig.

openshwprojects:
Hmmm I didn't think that indent affects anything, but ok. The {{ value }} stuff was just to be sure that no strange default setting breaks things. Anyway, I have applied your suggestions and I will try to check it now when I get some free time:
https://github.com/openshwprojects/OpenBK7231T_App/commit/585f633de568538ced3b56238e85e9d2ae30dc78
Can you recheck?

Also... "or the bulbs have been rebooted)." - I think there is a flag for bulb to broadcast it's state after reboot, but obviously, if you disable receiving the state then HA won't know what changed.... Maybe using the flag for saving RGB CW states for reboot can help

openshwprojects:
Wait.... shouldn't the rgb_value_template be:
[...]

PR #309, opened 2022-10-20
Discovery / publish initial values

iprak:
This PR address the issue of devices initially appearing unavailable. The MQTT publish for discovery has been adjusted to allow executing a command at the end, the command being `publishChannels`.

MQTT_PublishAll and MQTT_PublishChannels were reversed and have been addressed as well.

This should close https://github.com/openshwprojects/OpenBK7231T_App/issues/252 as well.

### DIFF
--- a/src/httpserver/http_fns.c
+++ b/src/httpserver/http_fns.c
@@ -1293,17 +1293,25 @@ int http_fn_ha_discovery(http_request_t* request) {
 	if (relayCount > 0) {
 		for (i = 0; i < CHANNEL_MAX; i++) {
 			if (h_isChannelRelay(i)) {
+				if (dev_info != NULL) {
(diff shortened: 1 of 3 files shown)


openshwprojects:
Passing a pointer to a string that has to be kept in memory is not a good practice, but... I understand that it might be necessary here. Doing a strdup or using a buffer would not be worth the effort.

The only other approach I'd consider would be just creating an enum , something like PostPublishCommand with options None, PublishAll, PublishChannels and using it instead of string command.... maybe it would be better....

because otherwise people might be tempted to pass a local char buffer[64]; as an argument and be surprised that it crashes

iprak:
Yes that is a good idea. I have revised my changes.

iprak:
@openshwprojects The BL602 build is failing. This commit (https://github.com/openshwprojects/OpenBK7231T_App/commit/716133a34ab461f1ffbcb6cbae9a7c0bf3435feb) introduced the function `add_dimmer` with no return value.

openshwprojects:
@iprak I just realized that and fixed, thx

PR #326, opened 2022-10-23
Malloctest

btsimonh:
replace ALL os_malloc() with malloc()
replace ALL os_free() with free()
Add memory/memtest.c with routines to check stack and malloc area (T only)
TCP client thread buffers allocated when needed and KEPT.  Reused on client calls.
MQTT - use 50 bytes of stack for topic if short enough.
Thread stack size changes - we need to analyse thread requirements better (use memtest!!!)


btsimonh:
ignore this one for the moment - for reference only

Issue #329, opened 2022-10-24
HomeAssistant Discovery / consumption_stats generates error


iprak:
The `consumption_stats` generates this error at HomeAssistant end

`
2022-10-24 09:35:07.592 ERROR (MainThread) [homeassistant.util.logging] Exception in async_discover when dispatching 'mqtt_discovery_new_sensor_mqtt': ({'device': {'identifiers': ['enbrighten1_8C000000'], 'name': 'enbrighten1', 'sw_version': '1.0.0', 'manufacturer': 'Beken Corporation', 'model': 'BK7231T', 'configuration_url': 'http://192.168.1.211/index'}, 'name': 'enbrighten1 consumption_stats', 'availability_topic': 'enbrighten1/connected', 'unique_id': 'enbrighten1_8C000000_sensor_5', 'qos': 1, 'device_class': 'None', 'unit_of_measurement': 'Wh', 'state_topic': 'enbrighten1/consumption_stats/get', 'state_class': 'measurement', 'platform': 'mqtt'},)
`

The call stack mentions this
`
raise er.MultipleInvalid(errors)
voluptuous.error.MultipleInvalid: expected SensorDeviceClass or one of 'apparent_power', 'aqi', 'battery', 'carbon_monoxide', 'carbon_dioxide', 'current', 'date', 'distance', 'duration', 'energy', 'frequency', 'gas', 'humidity', 'illuminance', 'moisture', 'monetary', 'nitrogen_dioxide', 'nitrogen_monoxide', 'nitrous_oxide', 'ozone', 'pm1', 'pm10', 'pm25', 'power_factor', 'power', 'pressure', 'reactive_power', 'signal_strength', 'speed', 'sulphur_dioxide', 'temperature', 'timestamp', 'volatile_organic_compounds', 'voltage', 'volume', 'weight' for dictionary value @ data['device_class']
`

I think this is because `None` is not a [valid device_class](https://www.home-assistant.io/integrations/sensor/#device-class). We probably should just not assign it in that case.

#296

valeklubomir:
I expected that None is valid entry for generic sensor. We can leave it empty. 
Question is if dev_cla is optional or mandatory.

iprak:
device_class is optional (https://www.home-assistant.io/integrations/sensor.mqtt/). It automatically sets thing like icon in the card. I can see it also being used to present value differently

PR #330, opened 2022-10-25
Removed None

iprak:
Addresses https://github.com/openshwprojects/OpenBK7231T_App/issues/329

### DIFF
--- a/src/driver/drv_main.c
+++ b/src/driver/drv_main.c
@@ -41,7 +41,7 @@ const char* counter_mqttNames[OBK_NUM_COUNTERS] = {
 const char* counter_devClasses[OBK_NUM_COUNTERS] = {
 	"energy",
 	"energy",
-	"None"
+	""
 };
 
 typedef struct driver_s {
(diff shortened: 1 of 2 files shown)


openshwprojects:
I don't have Home Assistant Discovery knowledge, I will just trust you it's okay. Please double check after release. Thanks.

iprak:
> I don't have Home Assistant Discovery knowledge, I will just trust you it's okay. Please double check after release. Thanks.

I have double checked and there is no error log. 

But I have noticed some other discrepancies and wanted to check with you.

a) Should `publishChannels/publishAll` commands publish values for sensors?

b) Should we sync up MQTT HomeAssistant Config to what is sent out by discovery or would you be okay with deprecating it? Currently, the config is outdated and does not handle sensors at all.

c) It seems that power drivers callbacks such as CSE7766_RunFrame run every second and call BL_ProcessUpdate to push updates. This eventually goes to `MQTT_PublishTopicToClient` which resets `g_timeSinceLastMQTTPublish`. If I reading it right, `MQTT_RunEverySecondUpdate` relies on `g_timeSinceLastMQTTPublish > 2` and that never happens. So invoking a command such as `publishChannels` doesn't do anything, right?

PR #342, opened 2022-10-26
Consumption counter storing functionality

valeklubomir:
Consumption data storing impemented. Added framework for other platforms.
Added calculations of consumption for actual day and previous days
Wifi/PING handler state machine updated
Added HTTP plugin for NTP
Added HTTP progress info for OTA
Resolved all compilation warnings.
New commands added.
README update
Minor bugs fixed.
OpenBK7231N SDK warnings resolved.


openshwprojects:
Whoa, this is really lots of changes! Thanks!
I will try to review each of them later today, at least briefly.
I wonder if it would be easier if next time, in future, you submitted your changes in several smaller PRs instead of one big.

openshwprojects:
@valeklubomir  may I ask what is the reason of doing **memset 0 of whole buffer before sprintf/vsnprintf etc?**
https://github.com/openshwprojects/OpenBK7231T_App/pull/342/commits/0ba61a7a81e015a476711f726eb3409c85f1a839
I am into C and C++ for about 15 years and I have never seen a need for such an operation, furthermore also documentation confirms that it is not needed.
If you want to clear a string for a strcat, you can do str[0] = 0, only set first NULL for termination char.
For sprintf/vsnprintf etc etc etc you don'e even need to clear the buffer, it only works like strcpy, always overwrites data in buffer.

openshwprojects:
@valeklubomir second thought:
https://github.com/openshwprojects/OpenBK7231T_App/pull/342/commits/6ecba9344d2e4a7404d8dc90f51742c2c04e6290
This breaks the UART2 port logging (TX2 sending log) while using monitor, is it intended?

openshwprojects:
The potential solution to:
https://github.com/openshwprojects/OpenBK7231T_App/commit/6ecba9344d2e4a7404d8dc90f51742c2c04e6290
would be to introduce a pin role that can be only set on UART pins, a LogTX, and set it by default on config creation (or on an update). That way pin monitor would know that given pin is an UART pin and would not overwrite it.
Still, maybe it would be better to hardcode the fix...

openshwprojects:
Other changes looks good, especially with the POWER update and even a readme update. Great work, thank you! I look forward to more. Just we need to resolve those two things I mentioned above.

valeklubomir:
Also do not forget to check/review request for OpenBK7231N SDK. Please.

btsimonh:
when looking at my latest device, I just polled all the GPIOs raw, and counted changes on them, but did not change configs of pins.  Then set config on a few pins manually and looked at results.  (i was looking for toggles at IR rates!).
The main issue with that approach was that if I set a pin incorrectly, and had to reboot, it was saved in flash and so I have to safe mode to get back.....  So I like the idea of a temporary pin display which clears at boot.  I'd also favour a change counter whilst monitoring a pin?

valeklubomir:
Good idea adding auto-disable during boot to this pin monitor of function. 
Forcing pins into digital input was needed because I do not not what setup was used before.
STILL IT IS DEBUG FUNCTION. Feel free to modify it As you wish. This way it helps me detecting pin assignments
without "reverse engineering" pcb.

valeklubomir:
But for incorrect configuration, SAFE mode was build which allows you to undo wrong configuration.

openshwprojects:
Thank you, is it ready to merge now?

valeklubomir:
Yes. For now. I am waiting for next failure, during last couple of days, situations where device looked like frozen, but was not occured. Just network was down, and device had trouble to reconnect.

valeklubomir:
OK. Now.

btsimonh:
@valeklubomir - although too much in one PR - well done.
I can no longer break MQTT with IR button held down - a cmd every 40ms if GAP is to be believed.  no more err: -1  aka MEM.  No more reconnect attempt....
Thankyou :) !!!!

btsimonh:
(If you know what fixed this, please do tell! ?)

valeklubomir:
This also need pull request for OpenBK7231N. MQTT could use larger ringbuff.

PR #344, opened 2022-10-27
After analysing serial output on T, it would seem that we can send to…

btsimonh:
… serial without spinning in a thread.

The changes will fill the H/W tx fifo from our logging fifo, but not wait for the hardware to be available, so we spend less tiime stuck if logs get full.


btsimonh:
Wait until I get the build working for other platforms :)

btsimonh:
good to go....

openshwprojects (approved):

Issue #354, opened 2022-10-29
W600 Device Hangs on Startup


iprak:
Something has recently changed in past few days with logging which completely hangs the device. This has happened to 2 of my devices after upgrading to 1.14.77.

I desoldered one back and then I managed to get this from serial log, just this much.

```
decompress success...
w600 Compiled @Oct 23 2022 20:38:22<CR>
Entering initLog()...
Commands registered!
initLog() done!
got serial: 127:Warn:CFG:CFG_InitAndLoad: Correct config has been loaded with 1 changes count.
Info:MAIN:Using SSID []
Info:MAIN:Using Pass [
Warn:CFG:CFG_InitAndLoad: Correct config has been loaded with 1 changes count.
Info:MAIN:Using SSID []
```

iprak:
I can confirm that this happened after the merge https://github.com/openshwprojects/OpenBK7231T_App/pull/344 @btsimonh  ?

PR #363, opened 2022-10-29
Fixed logging hang w600

iprak:
@openshwprojects
@btsimonh

This addresses https://github.com/openshwprojects/OpenBK7231T_App/issues/354 by reverting some of the changes from https://github.com/openshwprojects/OpenBK7231T_App/pull/344.

* A line used to be commented out in getSerial and was uncommented which results serial output flooding -- commented it back.
image

* Reverted and added back the `while`+ `rtos_delay_milliseconds(10);` in non-PLATFORM_BEKEN specific `og_serial_thread`. I don't understand why the loop is needed and identified this fix by reverting commits.

### DIFF
--- a/src/logging/logging.c
+++ b/src/logging/logging.c
@@ -457,7 +457,7 @@ static void getSerial2() {
 
 static int getSerial(char* buff, int buffsize) {
 	int len = getData(buff, buffsize, &logMemory.tailserial);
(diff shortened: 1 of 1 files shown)


openshwprojects:
Thank you! I hope that next time we won't be doing any more refactoring of generic shared code without testing important platforms, @btsimonh . Is it ready to merge?

PR #367, opened 2022-10-30
add driver SSDP. … run cmd "startdriver SSDP"

btsimonh:
This is NOT started by default.  run cmd "startdriver SSDP" - and you device will appear in windows networking....

We COULD in the future 'hear' notify messages from other OBK devices, so that all devices could know about each other... or even send an MSEARCH.

image


openshwprojects (approved):
Great addition!

PR #373, opened 2022-10-30
Add pin configuration for an alternative Aubess mini switch

zopieux:
Tested with 10 devices from Aliexpress, flashed with 1.14.29.


openshwprojects:
We are migrating to:
can you add it there?

PR #414, opened 2022-11-05
Shortcut icon

iprak:
This merge add a shortcut icon as the shortcut both for embedded and external webApp, it makes it look a little pretty. The icon itself came from https://ionic.io/ionicons to which I then added some color.

I noticed some partial attempt to render shortcut icon from lfs. I instead opted to have base64 embedded icon which avoid lfs and would be consistent across devices.

Additionally
* Corrected embedded webApp to render META, TITLE, SCRIPT and STYLE in HEAD.
* Simplified launch page generation for external webApp. It was assigning `obktype` which never got used. The launch page also matches the embedded webApp.
* Escaped deviceName

image
image
image


iprak:
@openshwprojects The OpenXR809 build failed here

```
cd sdk/OpenXR809/tools && wget -q "https://launchpad.net/gcc-arm-embedded/4.9/4.9-2015-q2-update/+download/gcc-arm-none-eabi-4_9-2015q2-20150609-linux.tar.bz2" && tar -xf *.tar.bz2 && rm -f *.tar.bz2
make[1]: *** [Makefile:80: sdk/OpenXR809/tools/gcc-arm-none-eabi-4_9-20[15](https://github.com/openshwprojects/OpenBK7231T_App/actions/runs/3398062094/jobs/5650782846#step:6:16)q2] Error 4
make[1]: Target 'build-XR809' not remade because of errors.
Prebuild attempt 1/3
```

Seems that toolchain is no longer available .. launchpad.net is not accessible. Can you check if you are able to access that url?

openshwprojects:
@iprak  I just requested a rebuild and it now works. IT seems it was a temporary issue.

openshwprojects:
Icon seems great, but for example here:
image
please try using sizeof( ) , it works for a constant size buffers.
Also... I wonder do we really need a buffer.... maybe it would be better/safer to create some kind of HTTP printf escaped function that directly adds to HTTP string or something...

iprak:
I think it would be quite inefficient if we created a function which I think would convert+push one char as a time. Also based on `postany`, it would require a delay between each send thus making it slow.

I actually did not modify the code which you mentioned, it was just a white spacing fix. But I have changed that and others to use `sizeof`.

iprak:
@openshwprojects It seems that the last 5 commits feb64bd305713fa8662704b2d209637aa1a3c66d - f1454d5ea81c9a51d7398c94220612e3732163d3 have been failing. 
I have seen build failures many times like this, perhaps we all should be using merge requests instead of direct commits.

openshwprojects:
@iprak  I was adding a quick fix for one of the users and forgot 0. It's now fixed.

@iprak  my main concern is that we have stack size issues and you have added a 255 bytes buffer on a stack. I think it should be possible to do an escaped print without this buffer or maybe copy this function and make an escaped version of it:
image

iprak:
I made a new function poststr_escaped like poststr, merging this logic into hprintf255 makes it complicated and the cases where escaping is needed are limited. I did adjust field generation code to use poststr_escaped since any field value might need to be escaped.

iprak:
I think maybe we need to prevent entry of characters which need to be encoded. These characters are not accepted by HomeAssistant (https://www.home-assistant.io/docs/mqtt/discovery/#discovery-topic).

openshwprojects:
@iprak  poststr_escaped escaped seems better, is this pull request ready to merge?

Please just double check or triple check, because breaking HTTP would also break OTA...

PS: Windows build is 99% ready, it is possible now to run whole obk (without MQTT) on windows, more details later.

iprak:
The code seems fine. The escaping only happens if there is any of those 5 characters are present.

But I mentioned in my last comment that names with such special char are not supported in HomeAssistant and the MQTT entity gets completely rejected.

iprak:
I have reverted the change to escape field values.

MQTT name should be restricted to `alphanumerics, underscore and hyphen`. This is followed by HomeAssistant and is considered a be industry wide best practice (https://docs.aws.amazon.com/whitepapers/latest/designing-mqtt-topics-aws-iot-core/mqtt-design-best-practices.html).

HomeAssistant validates and throws error is any other characters are found:
`2022-11-10 18:38:46.503 WARNING (MainThread) [homeassistant.components.mqtt.discovery] Received message on illegal discovery topic 'homeassistant/switch/O<p>enW600_CD2665E6_relay_0/config'. The topic contains not allowed characters. For more information see https://www.home-assistant.io/docs/mqtt/discovery/#discovery-topic`.


I have
* Reverted everything to not use `poststr_escaped` else but left the function there.
* Added a new helper function `add_label_name_field` to limit entry in ShortName and Full Name fields.
* One would see tooltip and message upon submitting with invalid value.
image


The change now is more simpler and would not accidentally prevent critical operations like OTA.

openshwprojects:
I think this is the better approach. Thanks!

PR #421, opened 2022-11-05
Added color background

iprak:
Made the color range look like this to match the same in Homeassistant along with some white space cleanup in generated html.

image


openshwprojects:
This is very, very good. Great addition! I will merge tomorrow morning.

PR #433, opened 2022-11-06
feat(new-device): add Jinvoo AC/TV Box IR Controller

lippertmarkus:


openshwprojects:
Thank you, that's a great teardown.

However, we are moving to the JSON format, could you also add it here?
The database stored in code is obsolete.

iprak:
@openshwprojects All builds for this merge are failing.

Issue #417, opened 2022-11-05
Temperature sliders in HomeAssistant and webApp don't match


iprak:
Guys can you check if the color temperature layout in our device or Hass is reversed?

It seems when I select a small value
![image](https://user-images.githubusercontent.com/6459774/200119109-b7572502-5166-4584-9f54-ebd8bcb52b34.png)
it appears like this in Hass
![image](https://user-images.githubusercontent.com/6459774/200119120-4468ca90-7052-4713-87c2-762dd9cfa867.png)

The same when value is at the other way around
![image](https://user-images.githubusercontent.com/6459774/200119227-145636d8-a2a7-4a0e-8325-bbed26d3cbe2.png)
![image](https://user-images.githubusercontent.com/6459774/200119235-1bb9e382-69a3-4899-9088-ac2ed6573e46.png)

The values in Hass and device do match.

openshwprojects:
That's strange! Give me a minute...

iprak:
Both the UIs are correct. 

We use Mired value which is 1,000,000/Kelvin_value.  In Kelvin lower values represents warm and higher cool and that results in inverted slider for us.

HomeAssistant uses Kelvin ( https://github.com/home-assistant/architecture/discussions/564).

I found this code related to color temperature and maybe that was prior to that change or the value got moved around:
```
// the slider control in the UI emits values
//in the range from 154-500 (defined
//in homeassistant/util/color.py as HASS_COLOR_MIN and HASS_COLOR_MAX).

#define HASS_TEMPERATURE_MIN 154
#define HASS_TEMPERATURE_MAX 500
```

MQTT integration only uses Mired value (https://www.home-assistant.io/integrations/light.mqtt/). 

Not sure if it is worth accepting Kelvin values in web and then converting it into Mired.

openshwprojects:
I am not aware about this. Are you saying that they have changed the standard used by HA during last several months?
My HA version:  
![image](https://user-images.githubusercontent.com/85486843/200124075-90792ec7-0dd0-4f92-b506-3ceeb454ed7f.png)

iprak:
Sorry @openshwprojects  I did not see your response while I was typing.

These are the raw values which I see in hass upon selecting a temperature
![image](https://user-images.githubusercontent.com/6459774/200125047-afb100dc-00c1-4182-b7b3-b6a5de4ad4bf.png)

[...]

PR #446, opened 2022-11-09
Kelvin temperature

iprak:
This merge changes the front end to accept kelvin value addressing https://github.com/openshwprojects/OpenBK7231T_App/issues/417

We still communicate with Homeassistant in mired. The form submit uses a hidden field to submit mired value and the conversion is done in JavaScript. The webpage generation code converts from mired to kelvin.

image


openshwprojects:
I am just wondering, would it make sense to make the setting customizable? Maybe by flag?
Is anyone using that HA range, 254 to 900 or something like that?
Maybe it would make sense to allow people to use this range on UI because it may help with scripting and testing?

openshwprojects:
Hmmm okay may it be, I will merge it for now.

PR #447, opened 2022-11-09
Improved resolution for RGB/CW color and brightness [Dev note: Will merge in few days, just need final feedback]

rogere66:
For RGB/CW the current code provides very poor color resolution at low brightness settings. This is because color and brightness is clamped to range 0-100 as integer. This pull request changes this such that color and brightness use float all the way down to the hardware level, i.e. to HAL_PIN_PWM_Update() (using BK7231 internal PWMs). This provides very good color resolution also at low brightness settings.

I also added some exponential to the brightness used in the PWM calculation. This gives reasonable changes in light intensity for each brightness step. Without it the steps are very large at low brightness and very small at high brightness.


openshwprojects:
This looks like a very good change, expecially the brightness calculation, but the problem is we can not afford to change the channel values to float. They should be integers.
Do you think it would be possible to integrate this change but keep channel values as integers?

rogere66:
When you say we can not afford to change the channel values to float, is that a speed or code size issue?
Functionally it seems to work just fine. The range is still 0-100 and conversion to integer is straight forward and done by the compiler even without explicit type-casting (could add type-casting for clarity though). For RGBCW the channel values and brightness already start out as float (baseColors[], g_brightness), but is changed to int half way trough the code (g_channelValues[]).
It should be possible to implement this while keeping channel values as integers, but would mean more code. The easiest would probably be to have two parallel channel values for float and int and use them as needed.

openshwprojects:
I mean that the whole codebase was written under assumption that channel is an integer and there are many places where that assumption is used. I am not able to list how many things may break if do that change now. We might do that change in the future, but it would require some real testing. Every MQTT publish, etc

rogere66:
I see your point about testing - it is easy to break something when changing the code. I was also assuming that channel values was handled as ints and was surprised to see that part of the code used float.
If interesting I can try to implement the code with g_channelValues[] as int, probably with some functions with both int and float versions. Otherwise I will just leave the code for reference.

openshwprojects:
Why not keep this calculation:
https://github.com/openshwprojects/OpenBK7231T_App/pull/447/commits/fa3a2c2f3d35b195c7456392907b77bf09d57ddd
but leave all remaining functions with integers? Do we really lose THAT MUCH precision because of integers?

rogere66:
I have tested this a bit and find that exponential brightness is not useful with int channel values limited to 1-100%. Lowest brightness step is 1%, next step is 2% (100% increase) then 3% (50% increase) and so on, and using exponential will not change this.
Color and brightness precision is really bad at 1% brightness with only 6 possible colors, R-RG-G-GB-B-BR. At 2% brightness it is better, but still bad. Maybe at 15-20% brightness it starts to get ok. It much depend on the actual bulb in use - if it is a very bright one it may be good to be able to dim all the way down to 0.1% brightness, while other bulbs are ok with, say 5%.
I also noticed that the code now includes an experimental option for smooth transitions and would think this function also would benefit from using float channel values, at least at the driver level.

leoshusar:
+1 for this.
I am using my bulbs in RGB mode mostly at night, usually running at around 5 percent of brightness, and those colors jump too much, as if there were only tens of colors.

The resolution at 1% brightness can only do like 60 colors (just a guess), because even this color wheel has more colors:
SmartSelect_20221116_034004_Home Assistant

When slowly swiping over it, up to 5 adjacent rectangles do not change the color at all, then it jumps to another.

rogere66:
I have updated the branch with explicit conversion from float to int channel values at some places to handle possible rounding issues. Also added a brightness control flag selecting 0.1% or 1% minimum brightness - should probably have a UI option for this if the code is merged at some point (can also be changed to use a variable to select, say 0.1-5%). With 0.1% (or even 1%) minimum brightness, some lights may get a large dead-band at the low end.
Also made a release including .rbl binaries if someone want to test this out.

WiredLife:
@rogere66 is it possible that you adapt the pr to the current version?
want to test with hsbcolor if the colors are now correct

rogere66:
Rebased the code to main and added a command to set brightness exponential mode - led_expoMode <0..4>:
          0 = Off
          1 = 1% min brightness with moderate exponential
          2 = 1% min brightness with full exponential (default)
          3 = 0.1% min brightness with moderate exponential
          4 = 0.1% min brightness with full exponential
The mode can be set by the "Change startup command text" Config command.
Also added ADC filter with MQTT publish - adc_value_pct, 0-100%.

WiredLife:
@rogere66 can you explain what ADC filter mean and does?
Is there any chance to get the colors right as on the original firmware?
For lamps with the original firmware i use this
https://github.com/TheAgentK/tuya-mqtt
And the colors are right. Maybe there is some magic in this code?

rogere66:
@WiredLife The ADC (with filter) makes it possible to connect a potentiometer to the ADC pin (P23) and use it to control something, like brightness, trough MQTT. It will not in any way influence the color.
The "magic" is likely to be in the Tuya code, probably as calibration values for the individual RGB LEDs. OBK does not currently implement RGB calibration, but it is quite possible to do so - I will look a bit more into this.

rogere66:
@WiredLife I searched some flash dumps and found a block containing "gamma_red/green/blue" which look like calibration values. The block start at a number of different addresses, probably in the range 0x1d0000 to 0x1df000, with 1kB increments, i.e. 0x1d0000, 0x1d1000 onwards. Can you check your device for this block and list the content? If you don't have a flash dump, you can use the Config/Flash Read Tool. List it in hex mode, 2400 bytes length.

WiredLife:
I have a dump of the original firmware of my LSC lamps, but i dont know if this helps.
Screenshot_20221213-012650_HEX Editor
Screenshot_20221213-012742_HEX Editor
Screenshot_20221213-012808_HEX Editor
Screenshot_20221213-012830_HEX Editor
Screenshot_20221213-012852_HEX Editor
Screenshot_20221213-012915_HEX Editor
Screenshot_20221213-012939_HEX Editor
Screenshot_20221213-013036_HEX Editor

rogere66:
@WiredLife No there is no RGB calibration data there, from address 1d9000-1e0000, but it could be on a lower address. Do you have full dump of the code, i.e 2 Mbytes? If so it may be better to just zip and post it.

WiredLife:
Ok here, just x'ed out my wifi credentials
[LSC.zip](https://github.com/openshwprojects/OpenBK7231T_App/files/10213316/LSC.zip)

WiredLife:
I have multiple lamps from different brands and they all have the same color drift with openbk
There must be a general calibration with works with many
Maybe this block here?
Screenshot_20221213-100145_HEX Editor

WiredLife:
@rogere66 take a look at
https://github.com/openshwprojects/OpenBK7231T_App/issues/516

All colors at the outer ring are ok, but if i go just one step inner, it looks like white with a touch of color, on all brightness levels.

rogere66:
That sounds like a lot more than RGB calibration. Have you checked the MQTT messages going in and out? You can also activate Flag1, "Broadcast led final color" in the config menu to check the sanity of the actual RGB values generated by the HSB commands.

rogere66:
@WiredLife I checked the code you sent and it appears that it does not use straight PWM for the RGB channels - how have you configured it? Have you tried setting Flag 8, "Alternate CW light mode" with corresponding channel settings? Do you know what RGB driver chip is used?

openshwprojects:
led_expoMode seems great, I can merge it anytime you want, just confirm it's ready.

rogere66:
@openshwprojects led_expoMode seems to work well now, but I'm not sure how useful it will be without float channel values - or do you intend to include that part as well?

openshwprojects:
@rogere66  I already provided a @WiredLife build with previous expontentional brightness (just initial version, hardcoded) and integer channels and it was tested and said to be good enough. So we will start with just brightness fix, no channel changes, and then we will see.

openshwprojects:
Ok, please do

WiredLife:
I have 4 lamps with SM2135 and one with BP1658CJ. All the same color drift.

rogere66:
The code you sent was for the bulbs using SM2135 then I presume. Could you send the code dump for de other device as well so I can confirm my setup?
led_finalcolor_rgb is working fine here. I tried sending the HSBColor command as well and all was good. Is it the same with full brightness?

WiredLife:
@rogere66 was my fault
It returns the right color values.
From the BP1658CJ one i sadly dont have a firmware dump because i directly flashed with tuyacloudcutter.

rogere66:
The firmware dump is not important so don't use effort on that.
I am trying to replicate your problem for the device using SM2135 so would like to know what channel assignments you use for SM2135/DAT and CLK, and possibly PWM for C and W?
E-mail would be ok, but I don't really want to publish my address here. Using github also have the advantage that it is available for others that may have similar problems. Maybe you can de-clutter the thread a bit by removing some screen dumps that are not useful anymore?

WiredLife:
@rogere66 
P24 (PWM4): SM2135DAT
P26 (PWM5): SM2135CLK
Thats all

rogere66:
I get some strange results with the LSC GU10 code - can you set Flag 13 "[HTTP] Show actual PIN logic level for unconfigured pins" and send over the result from the main page?

WiredLife:
@rogere66 
My SM2135 lamps are LSC GU10

PIN States
P00: NA P01: NA P02: 1 P03: 1 P04: 1 P05: 1 P06: 1 P07: 1 P08: 1 P09: 1
P10: 1 P11: 1 P12: 1 P13: 1 P14: 1 P15: 1 P16: 1 P17: 1 P18: 1 P19: 1
P20: 1 P21: 1 P22: 1 P23: 1 P24: NA P25: 1 P26: NA P27: 1 P28: 1

But the BP1658CJ from LEDVANCE (E14) has the color drift too.

rogere66:
@WiredLife No clue there. The strange thing is that the LSC GU10 code produce only PWM on all channels - no sign of I2C Clk/Dat. Thought maybe some pin need to be low to trigger I2C, but they are all identical on my setup.
But back to the color problem: included is a code that have printout on every PWM update. You can try this on the LEDVANCE bulb and use the Web App Log to check that the PWM values correspond to the color you set. Select Log Level Warn (tagged as Warn to get rid of most other messages). Brightness exponential is off, but can be set with command led_expoMode <0|1|2|3|4>

[OpenBK7231T_dev_20221214_154411.rbl.zip](https://github.com/openshwprojects/OpenBK7231T_App/files/10228801/OpenBK7231T_dev_20221214_154411.rbl.zip)

WiredLife:
But the LEDVANCE uses no PWM too i thought?
The pinout is this:
P8 (PWM2): BP1658CJ_DAT
P9 (PWM3): BP1658CJ_CLK

rogere66:
@WiredLife Oops, too quick there. Searched on Google and didn't notice that BP1638CJ came up first.
Maybe try this one then on the SM2135 device - hex PO of channels 1234 56, still as Warn messages.

[OpenBK7231T_dev_20221215_000722.rbl.zip](https://github.com/openshwprojects/OpenBK7231T_App/files/10232358/OpenBK7231T_dev_20221215_000722.rbl.zip)

WiredLife:
@rogere66 nothing from you in the log, even with set to all

rogere66:
@WiredLife Strange, it works fine here... If you want you can try to add log-printouts yourself to trace the problem - the printout I put in was in function SM2135_Write(). I don't have any SM2135 or BP1638CJ devices so don't think I can assist you much more on this part.
I intend to look a bit more into RGB calibration, but my impression is that your problem is more than just calibration. Without calibration you should still be able to produce almost any color by moving around on the color picker, but the color will not quite match what you select. That is how it works using Home Assistant and I think you should be able to do the same in openHAB.

beren12:
Have you looked into the bulb maybe not doing gamma correction internally? I know tasmota has a flag for that and it affected how brightness was scaled esp. at low levels.

https://tasmota.github.io/docs/Lights/#gamma-correction

Right after that there is a note "Internally Tasmota uses 10 bits resolution PWM to get smoother levels at low brightness. " Maybe it would be possible to do that as well.

rogere66:
Interesting point there - I wasn't aware that some bulbs have HW gamma correction. On the bulbs I have seen the stock Tuya code includes gamma correction and also some adjustment of RGB levels to correct color output, but it seems to only be a generic correction and not adapted to the actual device - the colors are still significantly off. Resolution seems to be good though, with 0.1% steps (i.e about 10 bits), but minimum brightness is like 10% which in many cases is way too bright.
Anyhow, gamma correction alone should not really change the resulting color of the RGB bulb - it merely changes the scaling of brightness.

WiredLife:
@rogere66 the current firmware situation is bad, the expo mode seems to not work with my pwm driven LSC Filament Bulbs. I cant go to the lowest brightness as it was with your patch. Maybe the float thing is needed for that? Is this included at your fork? So i'll say goodbye to the official and checkout and build your version.

rogere66:
I have updated the branch with LED gamma correction and RGB calibration functions. Calibration is straight forward: set RGB color to what you regard as white and execute the calibrate command. The branch includes a release with .rbl binary and detailed instructions on usage.

This branch still use float channel values and thus have much higher PWM resolution than the main branch, resulting in better color accuracy at lower brightness levels. The brightness exponential function is replaced with a more standard gamma correction function with adjustable gamma level and separate minimum brightness settings for RGB and CW LEDs.

WiredLife:
@rogere66 wow! Great work, i'll test it later. Is the calibration mandatory or will my lights still work right after the update?

rogere66:
@WiredLife Without calibration the lights will us default values and should work almost like the previous version.

WiredLife:
@rogere66 you are my hero! just updated my LSC Filament Bulb in the sleepingroom and it just works. Set brightness to 1% is so nice dark. thank you!
Later i will test on my rgb bulbs.

WiredLife:
@rogere66 youre a hero! it works perfectly as it should

@openshwprojects this change is a milestone
i can now calibrate my rgb bulbs to show the right color and my filament bulbs can go dark as possible
please commit to main

rogere66:
@WiredLife Thanks for the feedback - glad to hear it works.
When you say brightness 1% I assume that is the UI setting - the actual PWM setting with default gamma values will be more like 0.1% and that is not possible to achieve with the main code, with or without expo mode.

alfakenzo:
Hi, is this pr run now in master? Have same issue on the LSC GU10 and other rgbcw lights

openshwprojects:
You have convinced me. This fix is 100% necessary. I want it on main within few days.

**My only question is - would it be possible just to add a second "channel set", only for PWMs of RGBCW driver, so we don't change everything to floats?**

@rogere66  would it be possible to do something like that:

```
void Channel_Set_PWMFloat(int ch, float fVal) {
	int i;
	g_channelValues[ch] = (int)fVal;

	for (i = 0; i < PLATFORM_GPIO_MAX; i++) {
		if (g_cfg.pins.channels[i] == ch) {
			if (g_cfg.pins.roles[i] == IOR_PWM) {
				// support float here
				HAL_PIN_PWM_Update(i, fVal);
			}
		}
	}
	// no publish needed for internal PWM channels
	// Same with events, they are not needed, as far as I know?
}
```

That will replace calls here:
image

and only here, without breaking other stuff.

That would be perfect? Is there anything against that approach?

openshwprojects:
@rogere66  what is the purpose of ADC code or is it unrelated to this PR?

rogere66:
@openshwprojects  Nice that you want to include this function. It certainly is possible to implement this with separate PWM channels, but this will most likely also introduce a new set of problems. I my opinion it is cleaner and less risk in using float channel values and then closely evaluate each place the values are used. Some places will use the float value and some will use the value converted to int, in a deterministic way. I have not seen any serious issues using this approach so far.

I noticed that the functions I used for flash read/write was not available for some targets. I guess this can be fixed by including the gamma calibration values into mainConfig - what do you think?

The ADC code is not really related - I use it for controlling brightness of some lights using a potentiometer and put it in here for reference. If there is interest for this I can make a separate PR.

openshwprojects:
How big is the gamma correction table is? 

Channels are used by TuyaMCU and in many other places. For now, I will merge it with a second channel set function that works only for PWM - Channel_Set_PWMFloat - this is because nothing else is needed for those PWMs. We do not need them anywhere else.

Can you do the the Channel_Set_PWMFloat function or can I merge and then add it myself?

I got OBK running on Windows so there will be no problems with quick code changes.

In future, floating point channels are possible but right now we need this gamma feature fast for our users, so we go Channel_Set_PWMFloat route.

openshwprojects:
From what I see, the structure can be merged with main config but we would need to remove:
"char  tag[16];        // LED gamma block tag"
that takes 16 bytes unnecessary (config has a CRC etc etc so no tags needed), by default config memory is full of 0x00 so just maybe add some check based on that to see if table is not initialized

openshwprojects:
Regarding the "change channel type discussion".... I will just add here one my argument - in the current state of things, the LED driver PWM control could just totally skip channels and go directly to PWMs and it would as good as with float channels. That is why I chose Channel_Set_PWMFloat approach.

rogere66:
@openshwprojects It seems like you have a clear view on how to do this so just go ahead and merge the code. Using the Channel_Set_PWMFloat approach I guess you wold also need to look at external PWM drivers, Transition/lerp and probably a few other things.

You probably have found out already: there is no gamma table - it is calculated on the fly.
And yes, the tag is not really needed if included in main config, but the load function then need to check the parameters instead, and use defaults if not valid.

WiredLife:
@openshwprojects please dont forget to merge the code which makes my sleepingroom filament bulbs dark 😅
@rogere66 know what i mean i hope

openshwprojects:
@WiredLife  @rogere66 just to make sure, expo mode can be removed now, just as in the pull request?

WiredLife:
> @WiredLife @rogere66 just to make sure, expo mode can be removed now, just as in the pull request?

yes, can be removed because
brtMinRGB <f> - minimum brightness in % for RGB mode, range 0.0-10.0%, default = 0.1
brtMinCW <f>  - minimum brightness in % for CW mode, range 0.0-10.0%, default = 0.1
will do the job

openshwprojects:
Btw @rogere66 please do a pins merge because it seems like your commits are showing diffs related to 3click events?

rogere66:
@openshwprojects Is that in new_pins.c? All changes there (except for the ADC part) are part of float channel values so are not relevant if using the Channel_Set_PWMFloat approach.

openshwprojects:
@WiredLife  @rogere66 @alfakenzo I am merging it into my proposed channel system right now. I will do release in the following hours and will need some testers.

alfakenzo:
I'm here and can test later, when someone will tell me the cmd I can use.

openshwprojects:
@leoshusar  @esnet146  @alfakenzo  @WiredLife  @rogere66 
LED changes merged, please see here: https://github.com/openshwprojects/OpenBK7231T_App/issues/599

662KDV:
Non programmer here. Using LSC Smart Ceiling light (RGBCW, T7231N), unable to disable "exponential brightness" feature.

Controls in web interface and HA are not accurate. 

To achieve 10% warm/cold light, the slider must go up to 35%

Issue #444, opened 2022-11-08
Incorrect type for energy sensor class: The following entities have state class ‘measurement’ but ‘last_reset’ is missing:


otmezger:
**Describe the bug**
Hi, I uploaded the open beken firmware on a Aubee power metering plug. All is good, and Home Assistant sees the integration correctly via MQTT:

<img width="759" alt="image" src="https://user-images.githubusercontent.com/2758346/200566276-825090c8-f166-44c6-b12c-69e059f549e2.png">

However, when I add the "energy" sensor to the Energy dashboard of home assistant, I get the following error:

```The following entities have state class ‘measurement’ but ‘last_reset’ is missing```

<img width="384" alt="image" src="https://user-images.githubusercontent.com/2758346/200567129-3593696a-b8b3-4fcf-94d5-a0d2e428c566.png">

According to https://community.home-assistant.io/t/energy-dashboard-last-reset-missing/406733 and 
https://developers.home-assistant.io/docs/core/entity/sensor/, the sensor class should be `energy` instead of `measurement`.
Here is a quote from the forum:

> The state class for “energy” entities should be “total_increasing” 

**Firmware:**
- Version 1.14.129
- Device: Aubees power plug
- Chip/model: BK7231N
- Device config: 
<img width="592" alt="image" src="https://user-images.githubusercontent.com/2758346/200567056-e01f849a-5d1b-4cc3-9223-fb9e9d62a326.png">

otmezger:
I found this in the code, but I'm not sure how to use `total_increasing` for the energy sensor 
https://github.com/openshwprojects/OpenBK7231T_App/blob/fb8e9f129609db475c99082d172f83ac06663756/src/httpserver/hass.c#L266

openshwprojects:
Temporary solution might be adding sensors directly to YAML:
https://www.elektroda.com/rtvforum/topic3887748.html
In a meantime, @iprak , this is related to HA discovery, can you check this?

I will also try to check on my device later. My device was added manually by Yaml, it's a pretty old one.

otmezger:
Update: updated the firmware to 1.15.3 and started the MQTT discovery for Home Assistant. The same still happens.

iprak:
I will take a look.

iprak:
The documentation calls for `device_class` to be `energy`, the `state_class` can be measurement, total or total_increasing. And that is what I am seeing. `consumption_stats` is the odd ball.

[...]

PR #452, opened 2022-11-11
Energy values should be total_increasing

iprak:
This addresses https://github.com/openshwprojects/OpenBK7231T_App/issues/444


openshwprojects:
Is it complete? Which pull request is a correct fix for #444 ?

iprak:
This has the fix. 

The other merge won't work due to a couple problems in the code and unfortunately [otmezger](https://github.com/otmezger) has not been able to get it compiling so I just fixed it.

openshwprojects:
So I merge that one and other will be closed.

PR #483, opened 2022-11-19
Mqtt fixes

btsimonh:
What does this do?
It adds proxy functions for all the mqtt functions we use from LWIP.
These functions cause the real functions to be called from the tcp_thread context, which is where they must be called.
The proxy functions wait for the relevant function to have run on the other thread, and then return as though the original function had been called.

It also does not call OUR functions from the tcp_thread.  i.e. when data is received from mqtt, it is queued and serviced (currently from quick tick - could be better, like waking a thread).

### DIFF
--- a/src/mqtt/new_mqtt.c
+++ b/src/mqtt/new_mqtt.c
@@ -19,6 +19,15 @@
 #endif
 #endif
 
+// proxy functions....
+int mqtt_client_connect_proxy(mqtt_client_t *client, const ip_addr_t *ipaddr, 
+	u16_t port, mqtt_connection_cb_t cb, void *arg,
+    const struct mqtt_connect_client_info_t *client_info);
+mqtt_client_t* mqtt_client_new_proxy();
+void mqtt_disconnect_proxy(mqtt_client_t* client);
+int mqtt_client_is_connected_proxy();
+
+
 int wal_stricmp(const char* a, const char* b) {
 	int ca, cb;
 	do {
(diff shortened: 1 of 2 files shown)


valeklubomir:
This solution does not improve much, more like it is just adding more CPU processing to same task.
And probably more bugs are added. 
- post_recevied always does return same result even when publish is dropped due to full buffer.
- temp_topic and temp_data limited so 128 bytes. Everything greater is dropped. 
- Creating Proxy framework, and then you still have to wait for each publish operation completion. Kind of contra-point for multithreading.

btsimonh:
@valeklubomir 
it calls the functions within the correct thread, and so eliminates the risk.
Yes, we're limited to 128 bytes incoming.  This may be an issue for commands only?  everything else we expect is less....
(but calling our functions from tcp_thread is a VERY bad idea...)

"endless loop" - pls explain, and we can resolve?

basically the mqtt_ functions in LWIP are designed for a single threaded env.  This resolves that in the recommended way - by calling them from the single thread.

valeklubomir:
Endless loop:

	while (mqtt_****_proxy_cb_res == -10){
		rtos_delay_milliseconds(5);
		bk_printf("d");
	}


If LWIP fails to complete operation (loss of wifi/internet connection,....) and variable mqtt_****_proxy_cb_res is never changes ro != -10. Loop has no other way to exit.

What is -10? Could you define it by name?

btsimonh:
if they fail, we're screwed anyway.....
but yes, we should probably have a count, and error if it would die.
yep, -10, just a random number.

btsimonh:
@valeklubomir - added some stuff.... pls re-review.
On N without IR running, it processes MQTT receive every 50ms, resulting in mqtt publish, no failure in 2 hours.

btsimonh:
@valeklubomir - you are right, this is not the way to go.
For some reason, it just will not run with IR interrupt, and I don't know why....

btsimonh:
@valeklubomir - alternative approach:
https://github.com/btsimonh/OpenBK7231T_App/tree/mqtt_test2
not final yet (won;t build for other platforms, yet.

PR #488, opened 2022-11-21
add getcommands.js,

btsimonh:
can be run via
npm run getcommands
or
node scripts/getcommands.js
examines source for command registeration, moves stuff into comments and removes C strings..

Perfectly safe merge, as the PR does NOT change the source code!!!


btsimonh:
sorry... plus a very small fix for logging commands..... this does affect teh source, in a good way

PR #489, opened 2022-11-21
Undelayedstart

btsimonh:
modify startup on beken to start some things early, and some things after TCP is ready,

### DIFF
--- a/src/cmnds/cmd_main.c
+++ b/src/cmnds/cmd_main.c
@@ -89,19 +89,22 @@ static int CMD_Echo(const void *context, const char *cmd, const char *args, int 
 }
 
 
-void CMD_Init() {
+void CMD_Init_Early() {
     CMD_RegisterCommand("echo", "", CMD_Echo, "qqqe", NULL);
     CMD_RegisterCommand("restart", "", CMD_Restart, "Reboots the module", NULL);
     CMD_RegisterCommand("clearConfig", "", CMD_ClearConfig, "Clears all config", NULL);
 	CMD_RegisterCommand("clearAll", "", CMD_ClearAll, "Clears all things", NULL);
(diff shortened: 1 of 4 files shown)


valeklubomir:
Last changes in Pins handler caused that Button/WifiLED no longer work.

valeklubomir:
How this should work?

    uint32_t t_diff = g_last_time - g_time;
    // cope with wrap
   if (t_diff > 0x4000){
       t_diff = ((g_last_time + 0x4000) - (g_time + 0x4000));
    }

- "g_last_time - g_time" this gives always negative value, because g_last_time < g_time.
-  Logic behind 0x4000 is ??????

btsimonh:
1/ uint32_t t_diff = g_last_time - g_time; should be uint32_t t_diff = g_time - g_last_time; - corrected now.

2/ if g_time has crossed the 32 bit boundary. t_diff will be large.  So if it is large, just push both times over the boundary and re-calc.  Not sure if is required with unsigned arithmatic?

btsimonh:
button is still broken though :(.

valeklubomir:
Crossing bounadary of 32-bit value take 1193 days.
But this will work correct.

    if (g_time < g_last_time)
    { 
        t_diff = 0xFFFFFFFFUL - g_last_time;   // ms till crossing
        t_diff += g_time;                                 // ms after crossing
    } else {
        t_diff = g_time - g_last_time;
    }

btsimonh:
just pushed a fix for buttons.
@valeklubomir - check the last commit on main.  you will love the bug!!!!

valeklubomir:
debounce_cnt is only 3-bit wide. new debounce_ms == 15 > 4 bit.

PR #493, opened 2022-11-22
Change mqtt_processing to be driven by xTimerPendFunction if available.  Prep for Power Saving.

btsimonh:
@openshwprojects @valeklubomir - please look at this PR.

It implements the use of a FreeRTOS function xTimerPendFunctionCall to cause the callback in the timer thread - better than making a separate Beken timer for every time we use this technique.

To use this, add 
```
#define INCLUDE_xTimerPendFunctionCall 1
```
in FreeRTOSConfig.h

I see this as useful as we can extend `void OBK_timer_cb(void *a, uint32_t cmd)` to cover things like reading pins via GPIO edge interrupt, etc.
It will be essential to remove all short rtos_delay based loops to achieve power saving (e.g. logging, quicktick, etc. must all be made much less frequent).
Once all loops are eliminated/extended, I believe power saving is easily achieved using:

```
        bk_wlan_power_save_set_level(/*PS_DEEP_SLEEP_BIT */ PS_RF_SLEEP_BIT | PS_MCU_SLEEP_BIT);
        mcu_ps_enable();
```

but it won't sleep if looping quickly :(.

(beware of setting PS_DEEP_SLEEP_BIT - power off reset does not necessarily work when the device can survive minutes on it's capacitors!!!)

If investigating power saving, it's worth exporting some variables from the power save routine in the SDK (like actual slept ticks vs. requested) so you can know if it is actually sleeping at all.


openshwprojects:
That's perfect, but how do you define "looping quickly"? How much we can lengthen quick tick interval but still keep buttons with debouncing working? Of course we can just use on change interrupts for buttons, but are they available for each pin?

btsimonh:
this is what we need to determine.
It would need to be edge interrupt driven....
(and careful not to trigger from a very fast changing pin!!!).

valeklubomir:
Quick tick interval should be adjusted to usage. Now it is multi-purpose function timer. Everything which needed to be executed faster than 1sec, like: PINs, Output channels, LEDs, PWM and PWM smoothing, Relays,.... 
We should separate it. Basic Main Timer with reasonable operation period. 
Dedicated timers for LEDS, Buttons, PWM, IR,.... Firmware will activate specific timer is needed.

openshwprojects:
@btsimonh  any update on this? 
I think it is a very important upgrade because it seems that there are few budget devices (like relays) that do not like a large constant current drawn and just break they capacitors (or just manufacturers used a cheap ones instead of reliable low ESR...).
I already had to fix one such device in the past:  https://www.elektroda.pl/rtvforum/viewtopic.php?t=3898805&highlight=

btsimonh:
the issue is that 'quick tick' is used in multiple places.  e.g. drivers.
We need to determine what needs regular service, and what can be triggered by some external event.
e.g. IR input COULD edge trigger on interrupt to cause fast (10khz) interrupt to start.  And IF a result is detected, trigger a 'quick tick' to register the result.  (if we can;t run off edge interrupt for IR!...) The IR 20khz timer does not need to run all the time, only started on edge trigger.  But then it would need to disable edge trigger for itself until after the IR in was complete.... yet mot miss anything?

btsimonh:
DON'T MERGE.
will conflict?  with later powersave PRs.

openshwprojects:
I wonder how much value this currently has. Power Saving is already good, it has lots of room to improve but I am not sure if this PR would improve anything.

btsimonh:
yep this is now OLD.  I'm cognisant of what it contains and will try not to forget anything :).

PR #517, opened 2022-12-04
fix: setChannel should parse the client

### LINKED ISSUE #515: channelSet fails if topic has '/'
**Describe the bug**
I like to name my topics with a / to organize. 

For example, I like to set a name of 'openbk/christmas_tree_lights' for christmas tree lights, etc, so it's easier to navigate my MQTT tree.

It seems like the channel parsing of the code erroneously stops at the first '/'. So this would result in channelSet parsing '-1' when I call openbk/christmas_tree_lights/1/set=1 .

Glancing at the code, it looks like the issue is in `int channelSet(obk_mqtt_request_t* request) {`. 

I'll try to submit a PR soon.

**Firmware:**
- Version [e.g. 1.14.129]
1.15.130
- Device? 
Generic Tuya Plug
- Chip/model: BK7231T?
BK7231N
- Device config?
N/A

**To Reproduce**

Set client name with a '/'. e.g. 'openbk/light_xmas_tree'

Try publishing a message to turn on the relay. e.g. 'openbk/light_xmas_tree/1/set'

(failure) relay doesn't turn on (expected) relay turns on

** relevant logs **
```
Info:MQTT:MQTT client in mqtt_incoming_publish_cb topic openbk/light_xmas_tree/1/set
Info:MQTT:MQTT in topic openbk/light_xmas_tree/1/set
Info:MQTT:channelSet part topic light_xmas_tree/1/set
Info:MQTT:channelSet channel -1
```

no2chem:
If the mqtt client contains a '/', setChannel will fail to find the channel because it will stop at the first '/'.

This PR fixes setChannel to scan the client name instead, so it correctly skips over client. 

A new function `MQTT_RemoveClientFromTopic` is added, which skips over the topic.
In addition, the parsing in `channelSet` is greatly simplified, and and unnecessary copies are removed.

Fixes #515

### DIFF
--- a/src/mqtt/new_mqtt.c
+++ b/src/mqtt/new_mqtt.c
@@ -501,32 +501,52 @@ int MQTT_RemoveCallback(int ID) {
 	return 0;
 }
 
+/** From a MQTT topic formatted <client>/<topic>, check if <client> is present
+ *  and return <topic>.
+ *
+ *  @param topic	The topic to parse
+ *  @return 		The topic without the client, or NULL if <client>/ wasn't present
+ */
+char* MQTT_RemoveClientFromTopic(char* topic) {
+	// Get the current client ID.
+	const char* clientId;
+	clientId = CFG_GetMQTTClientId();
+
+	// TODO: cache the client ID length somewhere so we don't have to keep calculating it.
+	int clientLen = strlen(clientId);
+
+	// check if we have the client in the topic.
+	char* hasClient = strstr(topic, clientId);
+
+	topic = topic + clientLen;
+	// check if we have a '/' after the topic.
+	if (*topic != '/') {
+		hasClient = NULL;
+	}
+	
+	// If we have the client/, return the pointer, otherwise, return NULL.
+	return hasClient ? topic + 1 : NULL; 
+}
+
 // this accepts obkXXXXXX/<chan>/set to receive data to set channels
 int channelSet(obk_mqtt_request_t* request) {
-	// we only need a few bytes to receive a decimal number 0-100
-	char copy[12];
 	int len = request->receivedLen;
-	char* p = request->topic;
 	int channel = 0;
 	int iValue = 0;
+	char* p;
 
 	addLogAdv(LOG_DEBUG, LOG_FEATURE_MQTT, "channelSet topic %i with arg %s", request->topic, request->received);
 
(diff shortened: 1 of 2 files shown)


no2chem:
fixed, tested on my 7321N plug using a client with a '/'

no2chem:
I'd also recommend a better way of parsing the topic in the future, either way a lightwight regex or token parsing, this pointer math seems a bit fragile.

openshwprojects:
Will it work also for channel numbers that has two digits?

no2chem:
No (but it looks like it never did before). 

Maybe I'll take the time to actually fix the whole function 😃

no2chem:
actually i take that back, i'm pretty sure it works fine with channels with up to intmax digits, since the function calls atoi.

However, I'll just take the chance to cleanup the whole thing anyway..

no2chem:
@openshwprojects I'm not exactly sure. I'll leave the log open again for awhile but they were crashing only after I started to do a bit of debugging with the app.  Maybe it's actually related to MQTT or the like, though. Will let you know whenever I figure out how to repro it.

btsimonh:
I've been away for a couple of weeks, but I'm pretty sure my N can drop wifi (still connected, but no IP).
I can boot it from the router at a low level, and it will reconnect and work, so likely a DHCP thing.
@no2chem - would be useful if you reproduce to look at serial to see if it's actually still running.  If you don't have serial, try a router restart or kick it off the router and see if it comes back...

openshwprojects:
MQTT again? I donn't have HA always on, I'm using Devicegroups a bit more.
My Ns seems table:
image
image
image

no2chem:
So it's weird, I haven't been able to reproduce it at all. Maybe it was related to MQTT, not sure.

What would happen was the device would crash within the first 5-10 mins of startup and stop responding even locally. I'd have to unplug it and restart it.

I don't have serial on the devices in question, but will get that setup. Maybe just file it in the noise category, perhaps it was related to my setup.

I've also removed the offending `mqtt_sub_request_cb` that was causing the build to break on OpenBL602. I guess that MIPS compiler is configured a little differently than the ARM? It seems to error out on unused functions and the ARM doesn't, probably should look into that too...

openshwprojects (approved):

PR #533, opened 2022-12-10
Fix startup and toggle issues with Smooth LED transitions flag

### LINKED ISSUE #532: Default color when enabling "Smooth transitions for LED" (flag 18)
Hi,

> First I want to thank you for the awesome work on this project. It's a game changer to me and my home automation 🤩

The "Smooth transitions for LED" flag is awesome:
- It is very pleasant to change color
- When I turn on the light the fade-in effect is perfect
- The light turns on **instantly** (while fade-in)

There is only one problem: the default color on start is cold (RGBCW is set to #000000FF00).
And I haven't found a way to change it.

I use this startup command: `backlog ; BP5758D_Map 1 0 2 3 4 ; led_dimmer 100 ; led_basecolor_rgbcw #000000FFFF ; led_enableAll 1`.
On start the bulb fade-in to cold white and then the script is executed and bulb fades to cold+warm white.

Is it possible to have a way to define the default color?

Adrien

rianadon:
As I've explained in this comment: https://github.com/openshwprojects/OpenBK7231T_App/issues/532#issuecomment-1345404174, the default initialization of `finalColor` and the failure of color interpolations in the alternate cw mode to take into account whether the light is enabled will cause the light to initially be turned on after boot.

I don't have a build environment set up so I haven't tested these changes, but they should fix #532.

### DIFF
--- a/src/cmnds/cmd_newLEDDriver.c
+++ b/src/cmnds/cmd_newLEDDriver.c
@@ -51,7 +51,7 @@ int g_lightMode = Light_RGB;
 // Those are base colors, normalized, without brightness applied
 float baseColors[5] = { 255, 255, 255, 255, 255 };
 // Those have brightness included
-float finalColors[5] = { 255, 255, 255, 255, 255 };
+float finalColors[5] = { 0, 0, 0, 0, 0 };
 float g_hsv_h = 0; // 0 to 360
 float g_hsv_s = 0; // 0 to 1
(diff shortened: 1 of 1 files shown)


openshwprojects (approved):
Should be ok. I will give it a chance. Thanks.

rianadon:
Thank you! I was able to test the release, and the lights no longer turn on at startup which is what I hoped to achieve.
They still work with a startup command.

Issue #519, opened 2022-12-07
W600 LSPA9 version with BL0937 energy metering


openshwprojects:
According to this user, there is a such plug:
https://www.elektroda.com/rtvforum/viewtopic.php?p=20319760#20319760

Anyone here up to improving W600 support? @iprak  ? Remember last state, etc, etc?

iprak:
I can look into preserving the last state. Perhaps driver support could be enabled one by one as such devices are encountered.

Tryfos:
> According to this user, there is a such plug: https://www.elektroda.com/rtvforum/viewtopic.php?p=20319760#20319760
> 
> Anyone here up to improving W600 support? @iprak ? Remember last state, etc, etc?

There is also another (version of a) smart plug with W600 and BL0937, as I mention here https://www.elektroda.com/rtvforum/viewtopic.php?p=20337835#20337835   (Girier 16A JR-PM01). It comes with a C2BS chip too.

openshwprojects:
According to report of Elektroa user, it works now.

PR #535, opened 2022-12-13
W600 remember last power state

iprak:
This request starts tracking boot count and channel startup states for W600 and W800 devices.

I was only able to verify this for W600 but I am confident the solution will work for W800.

#519

### DIFF
--- a/src/hal/bk7231/hal_flashVars_bk7231.c
+++ b/src/hal/bk7231/hal_flashVars_bk7231.c
@@ -1,12 +1,12 @@
 /*
-    This module saves variable data to a flash region in an erase effient way.
+	This module saves variable data to a flash region in an erase effient way.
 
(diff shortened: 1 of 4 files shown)


openshwprojects:
IT would be better to port @btsimonh  flashvars to be multiplatform, but I think this PR is good for the start. Better than no storage at all.
Btw, can you look into HA discovery of CW lights? From what I heard, it adds CW lights as separate channels.

iprak:
@openshwprojects I had considered relocating some of that flashvars to a base level but I could not completely understand the BK7231 code in that regards.

Do you have any example or post mentioning the device setup?

openshwprojects:
Two PWMs. You can ask @Dark Man here:  https://www.elektroda.com/rtvforum/topic3866123.html

PR #539, opened 2022-12-13
Add Exponential Brightness

rogere66:
Make brightness exponential. Exponential mode can be set with command led_expoMode <0..4>:
0 = Off
1 = 1% min brightness with moderate exponential
2 = 1% min brightness with full exponential (default)
3 = 0.1% min brightness with moderate exponential
4 = 0.1% min brightness with full exponential
The mode can be set at startup by the "Change startup command text" Config command.
Modes 3 & 4 will not be very useful yet since minimum brightness is currently limited to 1%.

### DIFF
--- a/src/cmnds/cmd_newLEDDriver.c
+++ b/src/cmnds/cmd_newLEDDriver.c
@@ -11,6 +11,7 @@
 #include "cmd_local.h"
 #include "../mqtt/new_mqtt.h"
 #include "../cJSON/cJSON.h"
+#include <math.h>
 #ifdef BK_LITTLEFS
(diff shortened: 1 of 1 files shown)


openshwprojects (approved):

openshwprojects:
It seems that you need to fix compilation for BL602 platform by adding a header for pow

rogere66:
Ok. Added <math.h> and also changed pow to powf, the proper float version. The BL602 platform then apparently use stricter checks than BK7231 - maybe an idea to use stricter checks on every platform - even stricter than now.

openshwprojects:
There is still some issue but I will fix it for you after meging.

PR #540, opened 2022-12-14
Nologloops - toward power save

btsimonh:
This PR changes logging on Beken.
It removes spinning threads (serial, tcp), and instead calls a log trigger function, which pends a log send function against the timer thread.
This means that log sending will only be triggered when logs are ready to be sent.
Hence we can sleep for many ms when no logs are ready to be sent.

The TCP logging allows for up to TWO connections only on the port 9000 log port.
Additional connections will fail.
But closing the connection and re-opening is fine.

Negatives:
If the device is spinning in a timer function (e.g. MQTT gethostbyname), no logs will emerge until it is finished.
Positives:
We no longer use polling to service logs, and so avoid short rtos_delays

### DIFF
--- a/src/hal/bk7231/hal_main_bk7231.c
+++ b/src/hal/bk7231/hal_main_bk7231.c
@@ -22,6 +22,34 @@ _PTR realloc _PARAMS ((_PTR a, size_t b)) {
 #endif
 
 #define LOG_FEATURE LOG_FEATURE_MAIN
+
+// from rtos_pub.c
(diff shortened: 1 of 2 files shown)


btsimonh:
Question:
Can we use HTTP long polling for http logs?
i.e. when we get an HTTP log request, don't close the socket, but instead add it as a tcp socket that logging can send to as and when required?
This may/would allow power save when HTTP logging was active?

PR #543, opened 2022-12-15
Added basic drivers

iprak:
Basic drive support for W600 - NTP and BL0937. 

* I don't have BL0937 hardware so I faked PWM output from another chip as input. I can see power/current/voltage outputs but cannot verify their accuracy. So someone else will have to verify the actual functionality.
* I have verified that NTP driver works

Changes
* I have introduced some new definitions DRIVER_ENABLE_TUYAMCU, DRIVER_DEVICE_GROUPS_ENABLED to control what other integrations get used. We can relax this as we are able to verify those cases.
* The W600 build needs to be adjusted, the changes are in https://github.com/openshwprojects/OpenW600/pull/2.

https://github.com/openshwprojects/OpenBK7231T_App/issues/519


iprak:
@openshwprojects  Any comments on this?

openshwprojects:
It seems good in general, but I think it would be better to add ability to #define any driver ON/OFF and do not just copy-paste g_drivers array in the single #ifdef block. I will most likely merge it later.

iprak:
That is a good idea, let me do that.

openshwprojects:
Ok, merged.

Should I merge it now as well?

The final form of what we want would be:

#if PLATFORM_W600
#define DRV_ENABLE_TUYAMCU 1
#define DRV_ENABLE_BL0937 1
#endif 

etc etc and the same for Windows, BEKEN, etc...

iprak:
No, let me update that big block. It seems I also now need to resolve some conflicts.

PR #544, opened 2022-12-15
Powerlogging

btsimonh:
goes with SDK mods to log powersave

Increases quicktick onterval to 25ms - 5ms interval completely prevents PS.

Note: you will NOT get PS without the logging changes....


btsimonh:
NOTE: this increases quicktick interval to 25ms!!!
BUTTONS MUST BE RE-TESTED.

PR #545, opened 2022-12-15
W600: Implemented power save command

iprak:
Power save on W600.

This works. I am seeing the consumption fluctuate between 20 and 50 mA with it and that is while the chip was being powered by the USB Serial controller. When the setting is not set, I was consistent 80mA consumption.


openshwprojects:
That's a great news, I must check it on my W600 and W800 boards.

iprak:
@openshwprojects , any updates on this?

openshwprojects:
I am unable to test it right now but I will merge it based on your reports.

iprak:
This PR does not enable that option by default, it just provides a means to test it by running that command. We can just take it out if that doesn't help, Thanks.

PR #547, opened 2022-12-16
Move quicktick timer into user_main.c

btsimonh:
Call PIN_Ticks() from quicktick - so separating the functionality, but not changing anything.

This is mainly a cosmetic change - it does NOT change any functionality.

But it's in preparation for other changes - e.g. re-assessment of quicktick, running pins from interrupts, deep sleep...
And it provides functional separation....

### DIFF
--- a/src/new_pins.c
+++ b/src/new_pins.c
@@ -3,6 +3,7 @@
 
 #include "new_common.h"
 #include "new_pins.h"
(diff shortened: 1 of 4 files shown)


btsimonh:
Pls merge?
Next PR is based on this...

PR #556, opened 2022-12-22
Update flashing guide

kuba2k2:
This PR updates flashing guide to use bk7231tools instead of hid_download_py.

Both programs are CLI-only, so users are not losing on having a GUI vs needing to use CLI.
- bk7231tools has much better compatibility with BK7231N - it can automatically detect it (doesn't need the --unprotect flag), and can verify CRC properly
- it's faster, additionally it doesn't write all-0xFF blocks which are already there
- it's more user friendly I think

I think it's best to let people know that better tools exist - especially those that are new to Beken flashing and the entire "custom firmware" thing. People are confused about QIO/UA binaries (that's why I'd also suggest to use UA instead of QIO on BK7231N), or needing to add --unprotect or disable CRC checking. It's a shame that official Beken tools can't even check CRC properly.

Additionally, bk7231tools is also better for dumping. It doesn't produce dumps with swapped bootloader/app locations, and it also can check that CRC to verify everything's correct. Since your project has grown pretty popular, most people will first learn the flashing tools from your README and Wiki, which is why I'd like to let them know about the best available tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,18 +82,18 @@ then re-power the device (or reset with CEN by temporary connecting CEN to groun
   
 ## UART (multiplatform method, Python required)
 
-clone the repo https://github.com/OpenBekenIOT/hid_download_py
+Install [bk7231tools](https://github.com/tuya-cloudcutter/bk7231tools) using `python -m pip install bk7231tools`
   
 Use USB to TTL converter with 3.3V logic levels, like HW 597 
 
 connect the PC to TX1 and RX1 on the bk7231 (TX2 and RX2 are optional, only for log)
 
 start flash using:
-`python uartprogram <sdk folder>\apps\<folder>\output\1.0.0\<appname>_UA_<appversion>.bin -d <port> -w`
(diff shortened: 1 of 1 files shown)


kuba2k2:
**The new tool seems to also produce swapped dump files**. Since you're the author, please consider fixing it so that it reads the entire 2 MiB, both on T and N. The now obsolete instructions in the readme still point people to worse tools.

openshwprojects:
Thank you for your suggestion. According to my tests, the bootloader order is not a bug, I'd rather say it's a feature. It seems that it's not possible to read T chip bootloader directly (at 0x0 offset), it is only being read while we start reading 2MB at 0x11000 and they wrap around. That what happened for me while testing and it also explains bkWriter 1.60 behaviour.

So I'd rather treat that bootloader as a bonus to 0x11000 read and not as a bug. Of course, I could manually swap the byte sections in the resulted file, but due to the legacy purposes, I want T reads to be kept compatible with UA files (so start 0x11000) and N reads to be kept compatible with QIO files (so start at 0x0).

I am sorry, but we already have too much tutorials and materials regarding those files and it's a bit late too change standard. We do not want to confuse our users even more.

You have a point regarding N platform reading, that's why I will update hid_download_py in the following days and add a correct CRC check.

PR #558, opened 2022-12-23
W600: Basic driver support

iprak:
I redid the merge to avoid white space changes. As the project owner, at some point, you should make a decision to ensure consistent code with the goal of providing a better dev environment. That will promote more input into this project and also make it a joy to develop. The new simulation checks are in that direction. Unfortunately, formatting is a mess .. even on the same line. I tend to use VSCode because I can build directly without jumping into CygWin console and I like to make the code I touch consistent. This has been my experience with other Github project I have worked with.

But I completely understand the review effort for those unintentional formatting changes; I occasionally run into the same while reviewing changes from over enthusiastic developers :-)

====
This provides limited drive support on W600 - NTP, BL0937, HttpButton and TestDrivers.

I don't have BL0937 hardware so I tested this using PWM output from another chip. I can see power/current/voltage outputs but cannot verify their accuracy. So someone else will have to verify the actual functionality.

But I have verified that NTP driver works

I have defined a few controlling variables in `obk_config.h` to control the drivers. Some of that is mess due to inter dependency and I decided to keep changes to existing code limited.

* DRV_ENABLE_I2C has been I2C_SUPPORT; I did not see DRV_ENABLE_I2C being set anywhere.

### DIFF
--- a/src/cmnds/cmd_newLEDDriver.c
+++ b/src/cmnds/cmd_newLEDDriver.c
@@ -186,6 +186,20 @@ float led_lerpSpeedUnitsPerSecond = 200.f;
 float led_current_value_brightness = 0;
 float led_current_value_cold_or_warm = 0;
 
+void led_Save_finalRGBCW(byte* finalRGBCW) {
+#ifdef LED_DRIVER_SUPPORT
+	if (DRV_IsRunning("SM2135")) {
+		SM2135_Write(finalRGBCW);
+	}
+	if (DRV_IsRunning("BP5758D")) {
+		BP5758D_Write(finalRGBCW);
+	}
+	if (DRV_IsRunning("BP1658CJ")) {
+		BP1658CJ_Write(finalRGBCW);
+	}
+#endif
+}
+
 void LED_RunQuickColorLerp(int deltaMS) {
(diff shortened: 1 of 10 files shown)


openshwprojects:
Yes, we will need to think about that formatting. But that's not as important as many features for the users.

In general, very good job. The two cosmetic changes I'd suggest is:
1. we don't need #unfdef s, we just assume that no one defines that outside from the file
2. Naming could be slightly better , both with "Tuya" (it should be TuyaMCU so people are sure what is that) and we might need a more consistent names, maybe ENABLE_DRIVER_TUYAMCU, ENABLE_DRIVER_BL0942, etc 

Regarding IDE, I have a project in MSVC 2017 that builds the Simulator:
image
but it might be worth to add also some other project files, if possible. Would you be able to get simulator running in VScode, or do you have any other suggestions? 

I think I will PM you and we will think about it.

The simulator is not yet released fully.

iprak:
Those are good suggestions. I will make the changes soon.

iprak:
@openshwprojects , any thoughts on this merge?

openshwprojects:
Perfect, ready to run

PR #559, opened 2022-12-23
Updated OpenW600 Sub-module

iprak:
This updates the W600 submodule. Without this #558 just keeps failing to build. Including the submodule change in #558 itself was not successful.


openshwprojects:
Sorry @iprak,  I must have forgotton about it. I should have done it myself earlier, sorry again.

Btw, I have added a command to do HA discovery from command line, you can check if that's done ok if you want, but there were no significial changes.

PR #572, opened 2022-12-31
Added DHT support in W600

iprak:
My first attempt was to use rising+falling edge interrupts which would avoid the need to continuously poll for pulse detection. But there is seem to no high (micro) resolution timer so was not able to identify the pulse.
Next I attempted to use native `tls_usleep` and `tls_msleep` for delay but that seemed to lock the chip... probably because the timer might be used elsewhere.
So finally fell back to adjusting the nop delay to suit W600.

image

The DHT source files are included in https://github.com/openshwprojects/OpenW600/pull/3 and has to be merged first. Please don't forget to update the submodules after that.

### DIFF
--- a/src/driver/drv_dht_internal.c
+++ b/src/driver/drv_dht_internal.c
@@ -24,7 +24,7 @@ our DHT sensors wrapper is in drv_dht.c
 #include "../new_common.h"
 #include "../new_pins.h"
 #include "../new_cfg.h"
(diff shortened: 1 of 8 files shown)


openshwprojects:
That's great, @Indu  ! Very good work! Please don't worry about interrupts, I don't think they are required. It's just, idk, 10ms lag every 1 second (or rather every 3 seconds)? I don't think it breaks anything.

PR #575, opened 2023-01-01
Some DHT related fixes

iprak:
* We are restricting processing during OTA, this was being done using `ota_progress()==-1` check which only handled on kind of OTA and was not being used everywhere. I have created a new method `Main_GetOTAInProgress` to track any ongoing OTA.
* DHT pin definition was not setting the channelType. Due to this, the call `CHANNEL_Set > Channel_OnChanged` was not sending out any MQTT message because bCallCb was 0.
```
	if(g_cfg.pins.channelTypes[ch] != ChType_Default) {
		bCallCb = 1;
	}
```

I removed `CHANNEL_SetType` since that never saved the channel type into config and created `PIN_SetPinChannelTypeForPinIndex` as a replacement.

After this one should see something like this on status page
image

### DIFF
--- a/src/driver/drv_bl_shared.c
+++ b/src/driver/drv_bl_shared.c
@@ -200,7 +200,7 @@ commandResult_t BL09XX_ResetEnergyCounter(const void *context, const char *cmd, 
 #elif PLATFORM_W600 || PLATFORM_W800
 #elif PLATFORM_XR809
 #elif PLATFORM_BK7231N || PLATFORM_BK7231T
-    if (!Main_GetOTAInProgress())
+    if (ota_progress()==-1)
 #endif
     { 
         BL09XX_SaveEmeteringStatistics();
@@ -391,7 +391,7 @@ void BL_ProcessUpdate(float voltage, float current, float power)
 #elif PLATFORM_W600 || PLATFORM_W800
 #elif PLATFORM_XR809
 #elif PLATFORM_BK7231N || PLATFORM_BK7231T
-            if (!Main_GetOTAInProgress())
+            if (ota_progress()==-1)
 #endif
             {
                 BL09XX_SaveEmeteringStatistics();
@@ -583,7 +583,7 @@ void BL_ProcessUpdate(float voltage, float current, float power)
 #elif PLATFORM_W600 || PLATFORM_W800
 #elif PLATFORM_XR809
 #elif PLATFORM_BK7231N || PLATFORM_BK7231T
-        if (!Main_GetOTAInProgress())
+        if (ota_progress() == -1)
 #endif
         {
             lastSavedEnergyCounterValue = energyCounter;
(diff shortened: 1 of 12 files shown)


openshwprojects:
Good spot with OTA, but why a separate method? 

There is already "ota_progress", isn't? 

Maybe just use ota_progress?

Also...this is incorrect:
image
OTA exec call is non-blocking, so it will immediately set to false after setting to true.

openshwprojects:
Futhermore, you should set a channel type for Temperature_div10, so it's 21.0C and not 210C.

We already have this channel type thanks to... TuyaMCU.

openshwprojects:
Just check for DHT roles here:
image
do not set channelType without a direct request from user.
Channel_OnChannelChanged will just look for IOR_DHT_** role and if used, then do a publish.

iprak:
* `ota_progress` is only applicable to `otarequest` which only gets called for Beken platform. I am setting the new global flag in there as well.

> Channel_OnChanged

I switched to checking role but this block is very questionable then. 
```
	if(g_cfg.pins.channelTypes[ch] != ChType_Default) {
		bCallCb = 1;
	}
```

I have reverted the change for XR809. The OTA status cannot be handled from outside and I cannot verify the change at all.

I feel like the channelType setup is disjoint. One can erase the role or change channels and still retain stale channel values.

openshwprojects:
My suggestion is:
- implement ota_progress  on W600/W800 from within their SDKs
- do not set channel types automatically, just check if channel is used by DHT role, then mark bCallCB
Regaring channel types - they are a bit of 'advanced' setting and they are supposed to be users, not automatically. There is indeed only console command access for them, but maybe in future we could broke them out on GUI or maybe on web application to save flash memory

iprak:
* I have reverted all OTA changes. The OTA implementation for Beken in in this repo so I don't understand why you are asking the same for other chips to be implemented in their SDKs.
* Reverted changes for channelType as well. It would help if folks who introduce advanced features write some comments about it.

openshwprojects:
" The OTA implementation for Beken in in this repo so I don't understand why you are asking the same for other chips to be implemented in their SDKs."
Sorry, it seems that I was wrong, but still, I think you should just implement "ota_progress" for W600 and W800, most likely, when W600 manages to download OTA data (or starts downloading?) just set bool variable to 1 and do not even set it to false, because.... device reboot will reset it anyway.

iprak:
I reverted all the OTA related changes.

`ota_progress` is not callable from any SDK and this should be give more thought. We should handle ota failures as well. And iIdeally the same should be done for all SDKs or at least some stubs should be created. This will most likely require makefiles to be updated and should be done separately.

iprak:
@openshwprojects , any updates on this?

iprak:
Yes that function definitely was incorrect. I was under the impression that values will not be saved but `CFG_Save_IfThereArePendingChanges` is already called in `Main_OnEverySecond`.

openshwprojects (approved):

openshwprojects:
While some of the things were incorrect, the core DHT fix is indeed very good here and allows DHT to publish it's results. Good job.

PR #577, opened 2023-01-02
Return supportsSSDP

iprak:
Currently, webApp logs error if the chip is on older firmware or if the chip not supporting SSDP or if SSDP is not running.

image

This is because the webApp calls the end point `obkdevicelist` which is absent. We can handle this more gracefully in webApp by checking if the chip is currently supporting SSDP and that is what this PR does.

WebApp PR: https://github.com/OpenBekenIOT/webapp/pull/34


openshwprojects:
Wow, this is perfect. I am unable to check right now, but from a quick glance it seem ok so I'll merge it now. Thanks

Issue #561, opened 2022-12-24
MQTT initial reported IP address not updated


NielsPiersma:
**Describe the bug**
After running "Start Home Assistant Discovery" the CURRENT IP address is reported to HomeAssistant.
When the IP address changes, this is not reported to HomeAssistant. The change can be on the same or any other subnet.

**Firmware:**
- Version [e.g. 1.15.188 to 233/latest]
- Device LSC Smart Plug
- Chip/model: BK7231T
- Device config: n/a/

**To Reproduce**
Goto home assistant configuration
Run start home assistant discovery
move to a different subnet, but keep using the same HA server
Check home assistant MQTT device and see the ip address reported in the status page.

**Additional context**
Quick fix is deleting the device from HA and restarting HomeAssistant Discovery.

Niels

openshwprojects:
Thanks. I think you are indeed correct. IP is broadcasted on MQTT connect and not later. If it changes during runtime, it can indeed not be broadcasted to MQTT again.

I will look into it, I will also broadcast IP when it changes.

openshwprojects:
@NielsPiersma , I have added a fix, but I don't have dynamic IPs - I am using DHCP reservation for MAC.
Can you check if it works for you?
https://github.com/openshwprojects/OpenBK7231T_App/commit/adcf2ef0016b9fc5f00bb3a506d03c5281b9fcb4
It basically checks IP every frame and if it changes, it does MQTT Ip broadcast.
Thanks!

NielsPiersma:
Afternoon,

 

Not working as expected at this moment. I flashed device but only after starting the MQTT Discovery the ip addres was updated (to .99)

 

So I retested.

 

Current IP: 10.4.15.99 (from device)



 

In home assistant:



 

I updated the dhcp reservation, invalidating 10.4.15.99, so it must get a different IP address;

 



Now we have 10.4.15.102.

 

And over to HomeAssistant

 



 

Still stuck at 15.99

 

Re-run discovery from the device:



 

 

And updated in HomeAssistant…

 



 

Can we trigger an update / rediscovery? Or maybe do it on some kind of regular basis?

 

Niels

NielsPiersma:
I think all my pictures were dropped ;(.

Any way,
![image](https://user-images.githubusercontent.com/70269175/209560864-52a1b9eb-cb34-4362-942f-fea340a1affb.png)
[...]

PR #578, opened 2023-01-02
Invoke discovery on ip change

iprak:
This should avoid the need to use addEventHandler, etc. (#561).

Here is the output after the change:

```
Info:MQTT:MQTT_RegisterCallback called for bT cmnd/Device67202/ subT cmnd/Device67202/+<CR><CR>
Info:MAIN:Main_Init_After_Delay done<CR><CR>
Info:MAIN:Time 1, idle 0/s, free 30296, MQTT 0(0), bWifi 0, secondsWithNoPing -1, socks 2/8<CR><CR>
Info:MAIN:HA discovery is scheduled, but MQTT connection is not present yet<CR><CR>
Info:MAIN:Time 2, idle 0/s, free 30296, MQTT 0(0), bWifi 0, secondsWithNoPing -1, socks 2/8<CR><CR>
Info:MAIN:HA discovery is scheduled, but MQTT connection is not present yet<CR><CR>
Info:MAIN:Time 3, idle 0/s, free 30296, MQTT 0(0), bWifi 0, secondsWithNoPing -1, socks 2/8<CR><CR>
Info:MAIN:HA discovery is scheduled, but MQTT connection is not present yet<CR><CR>
Info:MAIN:Time 4, idle 0/s, free 30296, MQTT 0(0), bWifi 0, secondsWithNoPing -1, socks 2/8<CR><CR>
Info:MAIN:HA discovery is scheduled, but MQTT connection is not present yet<CR><CR>
<CR>
please wait connect net......<CR>
Info:MAIN:Time 5, idle 0/s, free 30296, MQTT 0(0), bWifi 0, secondsWithNoPing -1, socks 2/8<CR><CR>
Info:MAIN:HA discovery is scheduled, but MQTT connection is not present yet<CR><CR>
Info:MAIN:Main_OnWiFiStatusChange - WIFI_STA_CONNECTED - 4<CR><CR>
Main_ScheduleHomeAssistantDiscovery<CR>
Info:MAIN:Time 6, idle 0/s, free 30200, MQTT 0(0), bWifi 1, secondsWithNoPing -1, socks 2/8<CR><CR>
Info:MAIN:HA discovery is scheduled, but MQTT connection is not present yet<CR><CR>
Info:MAIN:Time 7, idle 0/s, free 21560, MQTT 0(0), bWifi 1, secondsWithNoPing -1, socks 3/8<CR><CR>
Info:MAIN:HA discovery is scheduled, but MQTT connection is not present yet<CR><CR>
Info:MQTT:mqtt_userName XXXX<CR><CR>
mqtt_pass XXXX<CR><CR>
mqtt_clientID Device67202<CR><CR>
mqtt_host 192.168.1.61:1883<CR><CR>
Info:MAIN:Time 8, idle 0/s, free 30200, MQTT 0(1), bWifi 1, secondsWithNoPing -1, socks 2/8<CR><CR>
Info:MAIN:HA discovery is scheduled, but MQTT connection is not present yet<CR><CR>
Info:MQTT:mqtt_connection_cb: Successfully connected<CR><CR>
Info:MQTT:mqtt_subscribed to Device67202/+/set<CR><CR>
Info:MQTT:mqtt_subscribed to cmnd/Device67202/+<CR><CR>
Error:MQTT:MQTT client "Device67202" request cb: err 0<CR><CR>
Error:MQTT:MQTT client "Device67202" request cb: err 0<CR><CR>
Info:MQTT:Publishing val Device67202 to Device67202/host retain=0<CR><CR>
Info:MQTT:[g_bPublishAllStatesNow] item -13 result 0<CR><CR>
Info:MAIN:Time 9, idle 0/s, free 30200, MQTT 1(1), bWifi 1, secondsWithNoPing -1, socks 2/8<CR><CR>
Info:MAIN:Will do request HA discovery now.<CR><CR>
Error:HTTP:HA discovery: No relay, PWM or power driver running.<CR><CR>
```


openshwprojects:
How would it behave in my case, where most of my devices were added by hand? Will it add a doubled entries?

iprak:
Yes that will happen since the manual entry most probably doesn't have the same unique id. I looked into two existing integrations (ESPHome and Tasmato) before proposing this but am still on the edge about it.

ESPHome and Tasmato are setup via custom integrations and so can adjust the HomeAssistant configuration internally.

We could use a flag or just have this case be handled via addEventHandler. 

The firmware displayed in Device page can still be incorrect even if ip address remains the same. Auto discovery (even a flag based) would be helpful if the module is reconfigured but I think that would be quite uncommon.

openshwprojects:
Hmm flag is the cheapest solution in terms of flash memory, it's just one bit.. Something like... "do autodiscovery on each startup and when IP address changes"? 
BTW, please remember - your solution will be only triggered when IP changes during runtime.
If device goes offline, and after restart gets new IP, the script event won't ever trigger.

iprak:
So maybe do discovery on startup and on ipaddress change. We could also invoke the same if any pin definition is changed.

openshwprojects:
Ok, I think it's reasonable, but please make a flag for it - so it works only if enabled by flag.

Otherwise that discovery could break users setups...

openshwprojects:
Need to check for safe mode here:
image

iprak:
Are the commands prevented in safeMode? Should I add the check to `Main_ScheduleHomeAssistantDiscovery` instead of this spot?

openshwprojects:
Safe mode forces AP so it makes no sense to attempt HA discovery there.

iprak:
> Safe mode forces AP so it makes no sense to attempt HA discovery there. At least not automatically. I'd just take that into the if(!bSafeMode) block visible on screenshot.

I understand the safe mode part. What I was asking was if commands are stopped in safe mode. If they are not, then the check can be inside Main_ScheduleHomeAssistantDiscovery instead of in the caller.

openshwprojects:
"What I was asking was if commands are stopped in safe mode. If they are not, then the check can be inside Main_ScheduleHomeAssistantDiscovery instead of in the caller."

I understand your idea, but it basically works that way:
- autoexec.bat and short startup commands are not run in safe mode
- but we can just force-disallow the user's usage of commands in safe mode. If users types command himself, then we should not block it
- there is also a "go out of safe mode button" (btw, just saying)

So I think that your idea of blocking that from within THAT function is not the best solution here.

Safe mode = blocked automatic init and execution of features. The keyword is "automatic". If user wants to do something, he should still can

I need to merge this soon because I will be adding another flag for UART on BL602.

iprak:
Thanks for the explanation. I was unable to determine if command were safe or not. I will make the change in a couple hours.

Do you think we could run out of flags?

openshwprojects:
I think we will just need to adjust our flags code to include second integer - so modif CFG_SetFlag and GetFlag
Once we reach 32, we will start using second 4 bytes int

openshwprojects:
I will need to add that flag for waiting user now, please update your code once I'm done so there are no conflicts

iprak:
@openshwprojects , any updates on this? I have DHT changes for discovery queued up behind this.

iprak:
Doing discovery in config change is just a convenience only for pro usage.

I think the most common approach is to device setup is like a wizard - 1) Setup wifi 2) Configure MQTT 3) Configure pins (either from native webApp) or use template (from new webApp) 4) Invoke discovery (from native webApp). 

So ideally there should not be a need to invoke discovery again unless a mistake was made in configuration. Hence, we would be fine even if this aspect is not included but I think it would be useful for the template part since one has to set template from newWebpp and then go back to the older one to invoke discovery ?


> This part here is not perfect: image because there is more ways to config pins, including:
> 
> * changing pins on web app (unmaintained approach)
> * applying a template on web app
> * setPinRole, setPinChannel commands etc
>   We might need to decide first if we just accept your solution and assume that other methods are well... pro-only? or figure out how to support them as well.
> 
> There is also another problem.. I saw users doing a realtime scripts like "When MQTT is offline, change pin role to X" on elektroda, and now that would redo a discovery... but maybe that's not a big issue...

That is an interesting setup. The discovery won't take place if MQTT is off and HomeAssistant will show all those devices as unavailable. Would be an issue?

PR #586, opened 2023-01-05
Add support for binary sensor (digital input) in Home Assistant discovery

Xinayder:
This change was motivated by a door/window sensor from Tuya, where I had to manually publish the discovery payload for the sensor (after converting to OpenBK) otherwise HASS would report the sensor as unavailable after a reboot or integration reload.

I had to change `get_Relay_PWM_Count`, and I found out it is also used by a Tasmota-related function, but that remains to be done as I have no idea how it would work to report or configure a binary sensor for Tasmota.

### DIFF
--- a/src/httpserver/hass.c
+++ b/src/httpserver/hass.c
@@ -244,6 +244,18 @@ HassDeviceInfo* hass_init_light_device_info(ENTITY_TYPE type) {
 	return info;
 }
 
+/// @brief Initializes HomeAssistant binary sensor device discovery storage.
+/// @param index
+/// @return
+HassDeviceInfo* hass_init_binary_sensor_device_info(int index) {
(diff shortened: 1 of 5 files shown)


openshwprojects:
Seems like a great change, from a brief glance at code it seems ok but @iprak  knows more about discovery.

@iprak , can you please take a look at this pull request and tell me if I can merge it?

iprak:
That looks great.

openshwprojects:
It seems that BL602 and W800 fails to build

iprak:
@Xinayder I think the build failures are due to `hass_init_binary_sensor_device_info` missing in the header file (hass.h). Although I am not sure why build for other chips succeeded.

openshwprojects:
Thank you!

PR #590, opened 2023-01-07
DHT discovery

iprak:
This PR adds supports for DHT sensors to HomeAssistant discovery

* The 1st commit was just enum/function renaming. `hass_init_sensor_device_info` has been renamed to `hass_init_power_sensor_device_info` because it only handles power sensors which are special. Enum names have been simplified as well.
* The 2nd commit introduces temperature and humidity sensors and this uses the new `hass_init_sensor_device_info` function. The temperature value is saved as x10 value in C and so the conversion back to correct value is handled via `value_template`. The value is rounded 2 places of decimal.
* The 3rd commit fixed discovery issues when only DHT sensor were present. The code has been simplified as well. Discovery could optional publish channels and this is now done using `MQTT_PublishChannels` which is direct and more efficient.
* 4th commit handles DHT role in `CHANNEL_HasRoleThatShouldBePublished`.
* And finally renamed DHT feature to SENSOR so that it more reusable.

My testing was to define DHT pin, enable flag 27 and restart the device. Discovery will take place on restart and channel values will be pushed out after that.

Note that temperature value is saved in C but my locale setting is F, that conversion is automatically by HomeAssistant. The temperature sensor declaration includes `unit_of_measurement` which makes this possible.

image
image


openshwprojects:
DHT rename to Sensor is perfect, I wanted to ask you about it and you did it even without me asking. We can't have separate logtype for every possible sensor.

This is also very nice:
image
that's how I wanted it to be done.

This, I understand, is a little optimization - ok:
image

Maybe call here directly?
image
that:
image

Is it checked and safe?
image

openshwprojects:
Hmm... it seems good... I will also check it on Windows, but let's give it a chance. Now TuyaMCU channels?

PR #591, opened 2023-01-07
Same shortcut icon 3 times smaller

lulol:
Replaces the web server base64 encoded shortcut icon with the same one almost 3 times smaller (264 bytes) in .gif format instead of the actual .png (748 bytes).

This is same size as [tasmota's .ico](https://github.com/arendst/Tasmota/blob/be988827487224e21a073799b1842e9f9f8f72a2/tasmota/html_uncompressed/HTTP_HEADER1_ES6.h#L6) format.


iprak:
Nice, I didn't think of gif when I added it.

PR #592, opened 2023-01-07
Added POWERSAVE status message

iprak:
While (test) powersave can be enabled via command, it is currently difficult to know if a device is in that mode. This PR appends the status to the status message.


openshwprojects:
Hm ok, maybe also... add a "PS=0" /"PS=1" on main HTTP page?
image
maybe...
PS=0, Channel 0 = 1, Channel 1 = 0? 
Smth like that?

openshwprojects (approved):

Issue #553, opened 2022-12-21
Support for WPA key as 64 hex digits


s09bQ5:
**Describe the bug**
The WPA2 standard (IEEE 802.11i) specifies a method for converting passphrases into the 256 bit PSK. The standard assumes that this method is used only for strings up to 63 characters to be able to distinguish the case where the PSK is passed directly as 64 hexadecimal digits. 

OpenBeken so far does not support passing the PSK as hexadecimal digits. If you enter more than 63 characters, the rest is silently dropped. I checked the BK7231N SDK and there appears to be code to handle the 64 hexadecimal digits case, but the intermediate arrays to hold the key are one character too short.

**Firmware:**
- Version 1.15.215
- Chip/model: BK7231N

**To Reproduce**
1. Enter a PSK as 64 hexadecimal digits
2. The device fails to connect to the network
3. Go into safe mode by quickly repowering the device several times
4. Go to the wifi config menu
5. See that the last digit in the passphrase field is missing

**Additional context**
There is no passphrase for my wifi that I could enter instead of the 64 hexadecimal digits. The PSK consists of 256 random bits. So either this is fixed in OpenBeken or I have to change the passphrase on all of my devices.

openshwprojects:
The current password field lenght is 64 bytes, which also must include the NULL terminating character for the screen.

By 64 hexadecimal digits, you mean... the hex notation? Where two chars make one bytes? FFAABB? 

Then we would need to make our field twice as large and also include space for NULL terminator....

Or do you want to pass field to be just one byte longer?

Hmmm our config has no free bytes directly after pass, we would need to make a config upgrade function to convert between old config and a new one....

@btsimonh , thoughts?

s09bQ5:
Only one byte more is needed. 256 bits = 32 bytes = 64 hexadecimal digits.
It would be possible to store the information in the existing space (by converting the hex string to binary and filling the rest with something != 0), but then the simplicity of the code is lost.

openshwprojects:
Solved with pull request submitted by user. Thanks!

PR #594, opened 2023-01-08
allow storage of WPA2 PSK as hexadecimal string

s09bQ5:
This is the main part of the fix for #553. It uses a separate statically allocated buffer for the key that is one byte larger than array in the on-flash config structure and places a terminating null byte at the end.

I have to admit that I don't know why there are os_* functions, but I guess they are supposed to be used since they exist.

This change depends on openshwprojects/OpenBK7231N#13 and openshwprojects/OpenBK7231T#18.
For BL602 the psk element in struct wifi_mgmr_connect_ind_stat_info is already 65 bytes long.
For XR809 the Doxygen comment of wlan_sta_set says that 64 digit hexadecimal keys are supported.
For W800 and W600 the disassembly of tls_wifi_connect shows that the function exits early if the key length is >= 65, so I assume 64 is ok.

### DIFF
--- a/src/new_cfg.c
+++ b/src/new_cfg.c
@@ -285,7 +285,11 @@ const char *CFG_GetWiFiSSID(){
 	return g_cfg.wifi_ssid;
 }
 const char *CFG_GetWiFiPass(){
-	return g_cfg.wifi_pass;
+	static char wifi_pass[sizeof(g_cfg.wifi_pass) + 1];
+
+	memcpy(wifi_pass, g_cfg.wifi_pass, sizeof(g_cfg.wifi_pass));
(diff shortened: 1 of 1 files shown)


openshwprojects:
This is nice. Very nice, I must admit. I didn't think about that approach. 
The ideal approach would be to update the config structure but then you'd have to write a function that converts from old format to new format...
and your approach - just works, I guess? 
I think that's a very good change, but just to be on safe side, can someone else look at that? @iprak  ? @btsimonh  ?

s09bQ5:
Ok, I see that os_memcpy and os_strlen does not work on the other platforms. Changed that to plain memcpy and strlen.

Issue #574, opened 2023-01-01
Driver Support for SHT3x


Chasbrot:
**Describe the bug**
First Happy New Year! I got a feature request. I have a super generic aliexpress temperature and humidity sensor based. Flashing and GPIOs are working, but the driver for the main sensor SHT30 is missing. I have looked at the supported devices and found two more that have the exact same sensor. 
https://www.elektroda.com/rtvforum/topic3933836.html
https://www.elektroda.com/rtvforum/topic3791330.html

**Firmware:**
- Version [1.15.251]
- Chip/model: BK7231N (CB2)

**To Reproduce**

**Screenshots**

**Additional context**

openshwprojects:
Hey, thanks for the suggestion, that's a great one!

I just bought SHT30 on ebay, lets wait for it to arrive.

Bbut are you aware that one linked device supports SHT30  but through TuyaMCU (so no direct SHT30  support) and the other one is just put on list because we also list non-WiFi devices for the sake of information?

If your device has SHT30  connected to TuyaMCU then it's most likely already supported.

Chasbrot:
I just took a quick look at it again, there is no tuyaMCU, only the two ICs. Yes, i have read both devices, but i thought they might all use the same driver. Yes, i will submit the new device. Thanks for the quick response.

catdogmaus:
I did manage to flash a very similar temperatute/humidity device with BK7231N like this https://www.elektroda.com/rtvforum/topic3874289-150.html#20181014 - that is not my post but my device is identical. This also does not have a driver, also I have no idea what to do about the pin setup. Eventually, they could have the same setup. It would be great if somebody could help.

superjedi:
> I did manage to flash a very similar temperatute/humidity device with BK7231N like this https://www.elektroda.com/rtvforum/topic3874289-150.html#20181014 - that is not my post but my device is identical. This also does not have a driver, also I have no idea what to do about the pin setup. Eventually, they could have the same setup. It would be great if somebody could help.

Apparently this device is using a CHT8305 temperature and humidity sensor: http://sensylink.com/a/products/lm2/CHT8305.html

catdogmaus:
> Apparently this device is using a CHT8305 temperature and humidity sensor: http://sensylink.com/a/products/lm2/CHT8305.html

Hopefully it will also get support at some point

superjedi:
> Hopefully it will also get support at some point

[...]

PR #607, opened 2023-01-14
Adding SHT3X Driver

Dheenhasty:


Dheenhasty:
to resolve #574 Adding SHT3X driver support maybe not so clean and could be a lot better

openshwprojects:
It seems great for a start, but depending on your coding skills, I'd suggest a few changes for a future.

1. if the I2C code is the same here and in previous CHT driver, and also in my Soft I2C driver, I'd consider merging that in the future
2. We are usually trying to avoid setting channel types where user can't see it 
image
is it required for you?

Dheenhasty:
Yes i have done that as a quick and dirty, but ideally a regroupement of SOFT I2C, CHT and SHT will be good. 

i will try to do that, but my coding skill might not help. 

for the channel set no it's was just to force the readability on web interface. but it can be removed and set at script level

openshwprojects:
@Dheenhasty ok I will merge it now and I will do I2C optimization later, ok?

Dheenhasty:
perfect, i will have some time to work on it this week end. but i will do another merge. 

Want to check the mqtt discovery for sensor too (could really help here)

openshwprojects:
@Dheenhasty DHTs are already discovered so discovery should be very easy

PR #608, opened 2023-01-15
Consolidate styles

iprak:
This PR includes changes to make styles consistent and then some fixes for incorrect.

Module
* Adjusted the message and increased its visibility (using P insted of H4 which is small).
	
Flags
* The header (H4) was not being closed, this making all flags small in size.
	
Startup
* Increased message visibility. 
* Changed layout to use LI and made some actions as links so that one can directly jump to the other section. 
* Adjusted "New start values" to be indented and use add_label_numeric_field. This also fixed the incorrect labels `for=\"boot_ok_delay\"` 
	
General
* Many places were using the same inline style e.g. style=\"display:none;\" , style=\"text-align:center; font-weight:bold; font-size:54px\". Moved them to stylesheet
* Temperature slider already had a class assigned but it also had inline style, moved the inline to stylesheet

image


openshwprojects:
Hello, was it tested also with things like a RGB WWW panel and a Power Measuring www panel?

iprak:
There are no changes for the main panel in this PR. They will display as before.

openshwprojects:
I think I will trust you on that one and merge and then just do a quick look if that's ok. Seems like an improvement. Thanks!

iprak:
Thanks. I will keep on watch on the webApp in case something doesn't look right.

PR #609, opened 2023-01-15
Win32 app / optional port

iprak:
By default the Win32 app runs on port 80 but it could be that 80 was already taken. I added an option to define `port=8080` in debugger command line options. The debugger command line options get saved into `openBeken_win32_mvsc2017.vcxproj.user` which should not be included in the repo. The argv params do get passed down to `glutInit` but I don't think that part should be impacted since we did not have any command line params to start with.

Also fixed the order of buttons which appear when Simulator is shutdown to be Yes/No/Cancel instead of Cancel/Yes/No.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ output/**
 .vscode
 node_modules
 Debug*/**
+libs_for_simulator/**
--- a/openBeken_win32_mvsc2017.vcxproj
+++ b/openBeken_win32_mvsc2017.vcxproj
@@ -83,6 +83,7 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
     <OutDir>$(SolutionDir)$(Configuration)\</OutDir>
     <IntDir>$(Configuration)\</IntDir>
+	<TargetName>openBeken_win32</TargetName>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug BL602|Win32'">
     <OutDir>$(SolutionDir)$(Configuration)\</OutDir>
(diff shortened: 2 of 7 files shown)


openshwprojects:
Very good idea, a proper Win32/linux style command args parse also would be good.
You would need a loop for that, you could also take arguments like:
-bDoUnitTests 1/0
-loadScene CfgFileName
etc

iprak:
* Do you manually copy the dependent assemblies to Output folder?
* Would you be okay changing the assembly name to just `openBeken_win32.exe` ?

openshwprojects:
_Do you manually copy the dependent assemblies to Output folder?_
No, I think output is for MCU build only. I could agree for doing a separate folder like a "bin" which will be only for simulator.

_Would you be okay changing the assembly name to just openBeken_win32.exe ?_
Ok no problem

iprak:
I am unable to get dlls like ILU.dll automatically copied to "Debug Win32" folder. I was only  able to debug after copying them manually. I feel I am missing something but I got it all working.

iprak:
Added another feature to automatically load the last project. What do you think about it?

iprak:
I think those could be useful suggestions but should be implemented as per use case. I feel that one would be mostly happy with the last config and the specified port which would be known to the consumer.

Xinayder:
Maybe instead of running on port 80 it can automatically run at a port like 8080, 9000, 5000? Or a random number. That way it will have less chances of having a conflict with port 80.

If this is the server for the webapp running on the simulator it really doesn't matter, does it?

iprak:
I don't know how the simulator uses the website but it can be used directly in browser to verify setup, actions. One would need to know the port.

openshwprojects (approved):

PR #612, opened 2023-01-15
Auto discovery activation for SHT and CHT Sensor

Dheenhasty:
Adding AutoDiscovery like the DHT sensor but for SHT and CHT Sensor

future soft I2C temp and Hum sensor just need to be added in new_pins.h in future to 

#define IS_PIN_TEMP_HUM_SENSOR_ROLE(role) (((role)==IOR_SHT3X_DAT) || ((role)==IOR_CH8305_DAT))


Dheenhasty:
Need some rework ... beginner issue :)

Dheenhasty:
Better :)

openshwprojects:
Thanks, added!

PR #637, opened 2023-01-27
Correct issue with Read battery (to rely on Relay)

Dheenhasty:
And detect if device on batery based on ADC default value.

Correct Issue with DeepSleep. (return misplaced. )


openshwprojects:
Hey, what is the purpose of relay toggle?
image

Futhermore, this will not work - pin index is different than relay index.

Dheenhasty:
Hello, 

   to be able to correctly read the ADC value we need to activate a relay pin before. 

[01-01 18:12:15 TUYA Notice][lr:0x5a017] key:,samp_type:-----val:1
[01-01 18:12:15 TUYA Notice][lr:0x5a017] key:,samp_sw_pin:-----val:17
[01-01 18:12:15 TUYA Notice][lr:0x5a017] key:,samp_pin:-----val:23

in the CBU case if on battery you need to put pin 17 in high and then mesure the pin 23 (ADC)
if not the pin23 (ADC) will always be at 4096 if battery has a suffisant level.

openshwprojects:
How have you determined that, on which board is this?

Dheenhasty:
For the determination, it's not me :) https://www.elektroda.com/rtvforum/viewtopic.php?p=20387961#20387961

For the other point, i have misunderstood something in the call then. i will rework all that. we can cancel this pull. i will submit another one with the correct way to implement this. 

Maybe a separate driver for Battery appliance could be better also than in global cmd. 

Again thank you for the help :)

openshwprojects:
So it toggles Q2 and pin 5 is ADC?
image

Dheenhasty:
Yes, but we need some delay for value to fix

Dheenhasty:
All the ADC management is include in tuya_drv_adc.c in the sdk  

static int adc_dev_control(tuya_adc_t *adc, uint8_t cmd, VOID *arg)
{
    unsigned short adc_buf = 0;
    int ret;
    
    switch(cmd) {
        case TUYA_ADC_VOLTAGE_CMD:

            ret = adc_dev_convert(adc, &adc_buf);
            if(ret != OPRT_OS_ADAPTER_OK) {
                return OPRT_OS_ADAPTER_COM_ERROR;
            }
            *(unsigned int *)arg = (adc_buf / 4096.0 * 2400); // 单位mv
            
            break;
       default:
            break;
    }
    return OPRT_OS_ADAPTER_OK;
}

But i'm unable to find a correct way to implement that ....

openshwprojects:
For now, can you merge latest changes and fix the pin index used as channel index?
I had to fix DeepSleep on the main branch because I don't know if your PR is ready or not.

Dheenhasty:
Hello, will try to do it today. but this part clearly need to be deep dive a little bit more.

Dheenhasty:
Ok finally have something consistent, i'll let the possibility to play with Some parameter as for the moment i only have  a CBU board to test. but it works fine on CBU

Dheenhasty:
So to add a little bit of explaination. the battery when new is up to 3V but the vref of the ADC is 2,403 on the CBU board so when you mesure the ADC you will have a full 4096 (who is a 12 Bits ADC not a 10 as explain in Tuya doc .....)

so you activate the divider Q5 who will divide by 2 the measurement and go below the 2,4 Ref . 

since i don't know if other board has the same behaviour i have put vref, adc bits and divider as optional parameter.

openshwprojects:
Is that delay: delay_ms(200); really necessary? It's 0.2s, long delay, as per function that can be called from quicktick

Dheenhasty:
I have done some test and no it's no needed, i added a logical test for the divider, if some board has no divider you can set it to 1 and relay will not be trigered.

Dheenhasty:
resync fork of main branch. now everything can be merge

PR #645, opened 2023-02-01
Add OffDimBright channel type for LED indicator brightness.

ianjanik:
I have a [Feit Smart Dimmer](https://www.elektroda.com/rtvforum/topic3949246.html) which has a LED indicator on the front. Its brightness level in controllable. Options are off, dim, and bright. I have added a channel type to support this.


openshwprojects:
Seems like a very simple but helpful change. I can't see anything wrong in the code, so I will merge it right away. Thanks.

PR #653, opened 2023-02-06
Add a flagged option to apply channel type multiplier (if any) when publishing MQTT value

remiserriere:
This PR adds a new flag to the list, so when a channel is set with a type _div10/_div100/_div1000 its value is divided by 10/100/1000 before publishing the value over MQTT as a float instead of an integer.


openshwprojects:
Hm, is double really needed for numbers like 0.001?

remiserriere:
> Hm, is double really needed for numbers like 0.001?

No ;) but if you run this code:
`    int iVal = 9999;
    float fVal = (float)iVal / 10;
    printf("value is %f", fVal);`
You see that the output is not 999.9 but 999.900024. It's even worse if you divide by 100, the result being 99.989998. 
It's not a big deal but I kept seeing this "issue" on during my tests. That's the reason why I went with doubles instead of floats. I leave it up to you tho :)

openshwprojects:
Ok, good job, you have convinced me

Issue #654, opened 2023-02-07
can not upgrade to newer version


thinhvuq:
**Describe the bug**

**Firmware:**
- Version: OpenW600_1.15.375
- Device: 3 Gang Wall Switch
- Chip/model: W600 - TW-03
- Device config?

Both OTA (update sofware by Wifi) and OTA (Web Application) can not upgrade to any newer version from version OpenW600_1.15.375.
Regards,
Thinh Vu

iprak:
It seems that builds 1.15.376 onwards give checksum error. 1.15.375 seems to be last good build which passes checksum.

```
magic_no=2701131679, img_type=0, zip_type=1
Error:OTA:Image header check failed
```

It seems that the image from 376 onwards overlaps with the run area. Perhaps due to new features being added to the image.

openshwprojects:
So we need to reduce the size of built binary for W600?

iprak:
I checked the make file to dump sections from the intermediate .out file using `arm-none-eabi-size` and that gave

	GNU-SIZE .output/w600/image/w600.out
	   text    data     bss     dec     hex filename
	 516884   10248  122220  649352   9e888 .output/w600/image/w600.out

c1cbcb6cabacc5f0eac8f8e3374d08a0d86e4857 (the last good version 1.15.375):

	GNU-SIZE .output/w600/image/w600.out
	   text    data     bss     dec     hex filename
	 513804   10240  122204  646248   9dc68 .output/w600/image/w600.out

After commenting out json_interface.c I got this and OTA was again operational.

	GNU-SIZE .output/w600/image/w600.out
	   text    data     bss     dec     hex filename
	 501044   10232  121948  633224   9a988 .output/w600/image/w600.out

So, yes, it seems that new code has been slowly pushing W600 image to its limit. I don't know if this was it or something else. In long term, maybe the build features should be opt-in.

iprak:
So then I commented out the content in `http_fn_ha_cfg` and the helper functions which gave this along with successful OTA. This was with the latest code.

	GNU-SIZE .output/w600/image/w600.out
	   text    data     bss     dec     hex filename
	 500740   10240  121964  632944   9a870 .output/w600/image/w600.out

openshwprojects:
What is taking so much memory? Can we disable it for W600?

I am also planning to shorten the strings and use a single function to check arguments count for commands.

iprak:
I don't know, this is just trial an error.

[...]

PR #659, opened 2023-02-10
Consolidated some code, removed white spaces

iprak:
This merge will help with https://github.com/openshwprojects/OpenBK7231T_App/issues/654 but will not resolve it.

* Removes white spaces from HTML
* Adjusts redundant code/strings in some functions: http_fn_cfg_mqtt, http_fn_cfg_webapp, http_fn_cfg_quick, http_fn_ha_cfg
* Also made a couple helper functions: poststr_h2, poststr_h4 - they seemed to help a bit

http_fn_ha_cfg can be sunset in favor of automatic discovery. Details about manual definitions can be provided via a wiki.

### DIFF
--- a/src/httpserver/http_fns.c
+++ b/src/httpserver/http_fns.c
@@ -19,6 +19,8 @@
 #include "../driver/drv_ntp.h"
 #include "../driver/drv_local.h"
 
+static char SUBMIT_AND_END_FORM[] = "<br><input type=\"submit\" value=\"Submit\"></form>";
+
 #ifdef WINDOWS
 	// nothing
 #elif PLATFORM_BL602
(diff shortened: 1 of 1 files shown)


openshwprojects:
Not bad, and very quick! Was it tested well? How much bytes it saves?

I will merge it soon, just let me read through changes.

Regarding HA discovery - yea, they should be merged, YAML gen with discovery code, but YAML generation feature should stay (just use the same code in the backend). Do you have any idea how to approach it?

openshwprojects:
Btw, would it make sense to replace flags text in flash with a script that fetches this text from github? And caches it in cookie for further use... or smth....

iprak:
The total reduction was round 300 bytes or so. After a while I lost track since I was making changes incrementally.

I had prototyped a change regarding caching/compression a while ago, which was to simply reference stylesheet content as a separate request. The content served was compressed (which reduced it down to around 40%) and was sent with a special modified flag (which is sent from client and server compares it before even sending anything). I can try to bring that stale branch up to speed and see what we can save. The saving was not in terms of bytes but also in terms of processing since the chip didn't have to repeatedly send back the same content.

Right not the embedded website uses static content. If that was changed to dynamic rendering like the webApp, then we can do a lot more optimization. That would be a big change. But we can special case flag strings and some of the more descriptive strings and I think that should reduce the size.

I think we can run into these issues with chips with 1MB flash size.

PR #661, opened 2023-02-11
Remove oneshotconfig

iprak:
This goes with this PR (https://github.com/openshwprojects/OpenW600/pull/6) to not use oneshotconfig at all.

We are not seeing OTA issues in W800 so I decided to keep the overall changes limited to W600.


openshwprojects:
Thanks, nice catch!

PR #662, opened 2023-02-11
Added Bridge Driver for polarized relays. Teardown: TONGOU TO-Q-SY1-JWT

valeklubomir:
@valeklubomir 
- Added bridge driver for polarized relays. Teardown: TONGOU TO-Q-SY1-JWT
- added logging specification for DRV
- minor warnings fixed


openshwprojects:
Thanks, but I have a question:
image
Are you sure about that change? We've been testing with users on Elektroda and there've been reports that negative peaks occurs randomly during normal power consumption. So, a device takes 50W all the time but then suddenly a -12321W peak happen. 
Our solution was to skip it, because reporting it as a 0 power consumption might give more erroneous results.

valeklubomir:
I think this is better solution, because:
- skipping occurs not only for sudden negative peak but also normally, when no load is present. probably due to phase shift between voltage and current which normally occurs due to inductive or capacitive load charcterisitcs. Here I would rather invert power value instead of setting 0 or skipping.
- skipping sample may cause wrong measurement due to larger intervals between samples and this will cause larger errors in energy metering 
- skipping samples does not show values in web page/mqtt always last value, which is problem for calibration. Values for Voltage/Current do not come as negative values.
- If this occur only for BL0942, which is serial communication, sudden peaks are probably NOT caused by wrong measurement by BL0942 but more like errors in data receiption. This requirres deep debugging, not simple patch. In this case sudden peaks could occur as positive peak, which will not be skipped.

openshwprojects:
Hmm according to the reports here:
https://www.elektroda.com/rtvforum/viewtopic.php?p=20350650#20350650
it's BL0942, but it's strange, because we're checking the checksum:
image
This is what I got from the user:
image
With your proposed solution, he would still get a "0 voltage" spike instead of constant 230V,  right?

valeklubomir:
I am not clearing all values only negative. So positive values are kept unchanged.
But you see my concerns about positive peaks which will not be filtered.

valeklubomir:
Unfortunately check is simple addition, and masked over 0xFF. And this type of checksum is known that complementary bit errors can pass through.

valeklubomir:
More sophisticated data recovery should be implemented. May be not skip samples but just replace with last valid value.

valeklubomir:
May be we can add simple calculation:

abs(power) <= (voltage * current) 

then value should pass. May be some limits for error. 
AC power measurement should always produce result:
power is always less than (voltage * current), due to phase shift between voltage and current.

openshwprojects:
I think that's a great idea, but just also multiply the power by some kind of fraction to get a safe margin, as you said.

valeklubomir:
I will use 10% as reserve.

Log output for device:

Info:EnergyMeter:BL0942 received: 55 BC 09 00 F8 8A 37 F4 3B 01 E7 FC FF C7 0B 00 FE 4D 00 01 01 00 A3
Info:EnergyMeter:Unscaled current 2492, voltage 3640056, power -793, freq 19966
Info:EnergyMeter:HEX Current: 000009BC; Voltage: 00378AF8; Power: FFFFFCE7;
Info:EnergyMeter:Real current 0.009, voltage 223.3, power -1.3
Info:EnergyMeter:Valid current 0.009, voltage 223.3, power -1.3

I will wait for 'Invalid' values.

valeklubomir:
Strange. I added another TUYA device do power meter and monitor power output.
BL042 reports always negative output:

Info:EnergyMeter:BL0942 received: 55 E9 34 00 39 9F 37 CC 3D 01 A0 EE FF CF 0B 00 EE 4D 00 01 01 00 78
Info:EnergyMeter:Unscaled current 13545, voltage 3645241, power -4448, freq 19950
Info:EnergyMeter:HEX Current: 000034E9; Voltage: 00379F39; Power: FFFFEEA0;
Info:EnergyMeter:Real current 0.053, voltage 223.6, power -7.4
Info:EnergyMeter:Valid current 0.053, voltage 223.6, power -7.4

Hmm this device uses CT transformer to measure I have to check if I have not swaped the direction. :-)

valeklubomir:
I forgot that I bought the device on aliexpress and most the device there are 2nd class and low cost.
Device probably did not pass quality check, because power measurement showed negative value.
I reversed CT direction and now it works just fine. Actually it need more precise calibration. 

Info:EnergyMeter:BL0942 received: 55 | 5C 2D 00 | B9 6D 37 | 25 40 01 | 46 0C 00 | 5E 00 00 | 06 4E 00 | 00 01 00 | 01
Info:EnergyMeter:Unscaled current 11612, voltage 3632569, power 3142, freq 19974
Info:EnergyMeter:HEX Current: 00002D5C; Voltage: 00376DB9; Power: 00000C46;
Info:EnergyMeter:Real current 0.046, voltage 222.8, power 5.2
Info:EnergyMeter:Valid current 0.046, voltage 222.8, power 5.2

openshwprojects:
How Tuya app handles negative values? Is this ready to merge?

openshwprojects:
Ok, I will merge now, and we will see how it goes. If you have any other changes, please open next pr. thanks!

PR #671, opened 2023-02-15
Bridge driver command. W600/W800 pin definitions update.

valekCN:
Bridge driver command for seting pulse length for switching relay.
Bridge OnChannelChanged bug fix.
README update.
W600 - added PA0 pin to pin list. W600-PICO module has LED there and can be used as wifi LED.
W800 - updated full list of PINS. (44 pins). 
W800 - Configuration structure changed to accomodate 48 pins. Other platforms unchanged.
Logging bug fix.


openshwprojects:
Thanks, I think I will review it tomorrow morning (in about 12h?) and merge it. Seems ok.

openshwprojects (approved):

PR #672, opened 2023-02-15
Accept empty startup command

iprak:
Allow clearing startup command.


openshwprojects:
Hello, wouldn't it clear startup command on every startup command page visit?

openshwprojects:
Thanks! Very nice!

PR #675, opened 2023-02-16
Added a Dockerized build environment

michaelkamprath:
I wanted to build OpenBK7231T for my Mac, but the current build process doesn't support that. So I created this Dockerized build environment and documented the process for using it to build. This approach would be useful on all platforms including Windows and Linux to ensure that the build environment is consistent.

### DIFF
--- a/docker/Dockerfile.build_env
+++ b/docker/Dockerfile.build_env
@@ -1,0 +1,16 @@
+FROM ubuntu:20.04
+
+ARG DEBIAN_FRONTEND=noninteractive
+ARG USERNAME=devuser
+
+RUN apt-get update \
+ && apt-get upgrade -y \
+ && apt-get install -y \
+                wget git python2 libc6-i386 \
+                make npm
+
+# build files expect "python", but it gets installed as "python2".
+# create symbolic link to fix this.
+RUN ln -s /usr/bin/python2 /usr/bin/python
+
+RUN useradd -m -s /bin/bash ${USERNAME}
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@ output/**
 node_modules
 Debug*/**
 libs_for_simulator/**
+
+# MacOS ignores
+.DS_Store
+
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -14,6 +14,7 @@ You would need `nodejs` if you plan on making JavaScript/styling changes. This i
 Gulp tasks should automatically appear in Explore pane in Visual Code. They can also be invoked from console by running `gulp`.
 
 # Building
+A Docker-ized build environment can be used to perform these builds. Instructions for doing so are [available here](./docker/README.md).
 
 ## Building for BK7231T
 
--- a/docker/Dockerfile.BK7231x_build_env
+++ b/docker/Dockerfile.BK7231x_build_env
@@ -6,11 +6,7 @@ ARG USERNAME=devuser
 RUN apt-get update \
  && apt-get upgrade -y \
  && apt-get install -y \
-                wget git python2 libc6-i386 \
+                wget git python3 python-is-python3 libc6-i386 \
                 make npm
-
-# build files expect "python", but it gets installed as "python2".
-# create symbolic link to fix this.
-RUN ln -s /usr/bin/python2 /usr/bin/python
 
 RUN useradd -m -s /bin/bash ${USERNAME}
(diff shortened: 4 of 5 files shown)


openshwprojects:
Very nice, but is it only for T platform so far?

docker/Dockerfile.build_env
```diff
@@ -0,0 +1,16 @@
...
+
+RUN apt-get update \
+ && apt-get upgrade -y \
+ && apt-get install -y \
+                wget git python2 libc6-i386 \
+                make npm
+
+# build files expect "python", but it gets installed as "python2".
+# create symbolic link to fix this.
+RUN ln -s /usr/bin/python2 /usr/bin/python
```
Xinayder, line 14: Unless you explicitly need `python2` to compile, the symlink is not required. If you read Ubuntu 20.04 release notes, they say no package is going to provide `/usr/bin/python` anymore, if you install `python3` you get the `/usr/bin/python3` binary, the same thing for `python2`.

If you need the symlink you should install `python-is-python3` package or `python-is-python2`, available on the official repositories. But as mentioned above, I believe `python3` should work as well, unless there's an explicit requirement for `python2`.

michaelkamprath: I was having some troubles using `python3` (if you look at the SDK code, you'll note they tried to wedge in `python3` support on top of original `python2` code), but I cannot replicate that at this moment, so I will switch it to `python3`. But, you do need a symlink to the python executable [because of this line](https://github.com/openshwprojects/OpenBK7231T/blob/master/platforms/bk7231t/bk7231t_os/build.sh#L82). I figured it is easier and less impactful to create the symlink than to change the original SDK code. As far as using `python-is-python3`, if you look at what it does, the `.deb` is just a symlink too. With said, I will change it to `python-is-python3` because it's the same thing.

michaelkamprath:
@openshwprojects Admittedly I only tested it with the T platform because that is what I needed. Nonetheless, I now have tested it out and can confirm I get a successful build for `BK7231T` and `BK7231N`. I cannot build `XR809` [because of this line here](https://github.com/openshwprojects/OpenXR809/blob/master/gcc.mk#L10) that default the build tools path. `BL602` requires a Windows tool to build per instructions, and I don't have that. It looks like I could make this work for `W600` by integrating the toolchain download into the docker container build, but I haven't done that yet. And `W800/W801` requires software to be downloaded from a site all in Chinese, which I cannot read to determine which one would b the Linux download.

Given that this docker image then is only useful for `BK7231T` and `BK7231N`, I will update things to reflect that.

openshwprojects:
Hmm, so how does the online github builds deal with this line:
https://github.com/openshwprojects/OpenXR809/blob/master/gcc.mk#L10
?
In general, we have online Github builds, so maybe we can use them as a reference while making the Docker system?

I will be happy to merge it as it is now, but it would be great to include other platforms

michaelkamprath:
@openshwprojects Feel free to merge the PR, as it does support two platforms. I'll see about adding more when I have time.

I did look into the Cygwin/cygdrive & "how does it build on Github actions?" mystery. The answer lies in [this line in a Makefile](https://github.com/openshwprojects/OpenBK7231T_App/blob/main/Makefile#L93). It overrides the variable set in the original line I cited. Looking at things more, I am not sure why my linux environment still tried to compile the `OpenXR809` project using the Windows tools. I will have to investigate that more.

Xinayder:
> I did look into the Cygwin/cygdrive & "how does it build on Github actions?" mystery. The answer lies in [this line in a Makefile](https://github.com/openshwprojects/OpenBK7231T_App/blob/main/Makefile#L93). It overrides the variable set in the original line I cited. Looking at things more, I am not sure why my linux environment still tried to compile the OpenXR809 project using the Windows tools. I will have to investigate that more.

Have you tried using the same commands as the GitHub Actions workflow file?

michaelkamprath:
@Xinayder That's exactly what I planed to use. Unfortunately, I won't have the mite to investigate using the GitHub actions as a template for a dockerized build for another week or two.

openshwprojects:
SO I will merge for now. Thank you, please look into expanding it soon further when you have time! Also, don't we need a Readme update?

michaelkamprath:
OK, I figured out what the GitHub actions build was doing and it's plainly not the same things as described in [the BUILDING instructions](https://github.com/openshwprojects/OpenBK7231T_App/blob/main/BUILDING.md). I can make a docker-based process that allows you to do the same thing locally. I will update things and generate a new PR.

PR #685, opened 2023-02-20
[HASS] Adding Battery driver discovery ang sensor avty flag

Dheenhasty:
add auto discovery for battery on hass when driver BATTERY is started
add a flag to ignore the avty on sensor measurement for hass to keep value 
   it does not publish status on sensor value so you had to rely on update time on HASS


Dheenhasty:
@openshwprojects > should be better regarding the mess of comit. 

you might want to take a look on flag, i tested it but unsure if i respect your method/Naming convention.

openshwprojects:
Typo: AVAILABILITY  vs AVAIBILITY

openshwprojects:
BATTERY -> Battery

in future, please do not commit coding style changes because they make it a bit harder to find out meaningful bits of code in the PR

Dheenhasty:
everything is corrected.

PR #692, opened 2023-02-24
Adding RunFrame for SHT Driver.

Dheenhasty:
Will launch on each frame a measurement.
detect automatically if periodical is activated


Dheenhasty:
@openshwprojects  > This one is ready to merge and hopefully without change in coding convention :)

openshwprojects:
Maybe also add something like
SHT_MeasureInterval , with default value 15 (every 15 seconds/frames), and command to set it to any value.

Doing measure every 1 second is an overkill. And some people might want to do it manually, like you. So "SHT_MeasureInterval 0" would turn off every second automatic measure.
And SHT_MeasureInterval 60 would do every minute (60 seconds), etc

Dheenhasty:
Done :)

openshwprojects:
Thanks!

PR #694, opened 2023-02-25
Clarification text for the pin mapping configuration.

tjkolev:
As a first time user I found the help text somewhat confusing. This is a suggestion to clarify things a bit.
* Dropped the reference to "textfield" as this is implementation specific. One day it could be one of those numeric up/down spinner controls. And I am not sure one can enter text value when index is expected to be numeric. Unless there's support to enter variables...
* Added a reference to the dropdown field, as this is what I see as "the first" field one needs to set. And it is a text value. The previous reference "first textfield" was actually referring to the second numeric field.

It would also be nice to link "role" to a documentation about what all these values mean. I could not find one. I don't know most of them and how are they used. I just took a guess.

tjk :)

### DIFF
--- a/src/httpserver/http_fns.c
+++ b/src/httpserver/http_fns.c
@@ -2079,9 +2079,9 @@ int http_fn_cfg_pins(http_request_t* request) {
 
 	http_setup(request, httpMimeTypeHTML);
 	http_html_start(request, "Pin config");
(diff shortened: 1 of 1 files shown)


openshwprojects (approved):

tjkolev:
> Hmm, not bad... I think I may merge it well. Btw, are you interested in helping us with docs?

Sure, I can help with documentation. I can't commit to tight time frames, or deadlines.

tjk :)

PR #696, opened 2023-02-26
Correct Float for SHT.

Dheenhasty:


openshwprojects:
Hey, why floor? I don't think floor is needed. Sprintf can do everything itself, without floor.

Dheenhasty:
- Correct SHT Display Value. 
- Adding possibility to deactivate autostart for all drivers on flag.

Dheenhasty:
Ready to merge

Dheenhasty:
> Hey, why floor? I don't think floor is needed. Sprintf can do everything itself, without floor.

i removed it for a int calculation, a bit overkill maybe, it's just an habit on my side to correct data before showing them.....

PR #697, opened 2023-02-26
Changing behaviour of battery driver

Dheenhasty:
To provide Zero configuration for End User :
  - 2 new pins BAT_ADC and BAT_Relay
  - Autostart of driver if pins associated
  - change vdivider default value to 2.29 Seems like default for 2AA battery sensor


openshwprojects:
Not bad, altough it would be worth to check in the future how often is that ADC+Relay mechanism used in the IoT devices, because I have personally haven't found one with that yet.
Would that also work for a device that had ADC on battery, but no relay?

Dheenhasty:
Hum it would mean that the device is powered by a 2,4V battery (vref of ADC is limited to 2,4)

but yes i can do that (just need to rely on IOR_BAT_ADC for detecting the driver need and put a vdivider has 1 if no relay detected)

PR #700, opened 2023-02-26
Adding Tasmota Mecanism to Sensor. 

Dheenhasty:
CH8305 and SHT3X supported


Dheenhasty:
Need to resync branch before

PR #710, opened 2023-03-06
Docker-based environment for building all target platforms

michaelkamprath:
This builds on my last submission to create a docker-based environment for building all target platforms. The big change here is that this version leverages this repository's `Makefile` rather than following the instructions in the [`BUILDING` documentation](https://github.com/openshwprojects/OpenBK7231T_App/blob/main/BUILDING.md), which are incomplete.

This should work on any x86 platform that supports Docker.


openshwprojects:
@michaelkamprath  would it be possible to build windows exe for simulator as well in docker?

michaelkamprath:
@openshwprojects I doubt it, as docker generally creates a linux environment. However, I am not a Windows expert, so someone else might know better.

PR #722, opened 2023-03-11
Fix for frequent restarts of S09 device 

gromdar:
I've got two S09 desktop thermometer / hygrometer / calendar + IR send/receiver devices .
On both I observed the same problem with frequent restarts of wifi every about 800s. 
I found that problem comes from MCU that resets WIFI module because it doesn't respond correctly on two commands  : 
0x2B - get network status 
0x24 - get RSSI 

after my changes they work fine and additionally RSSI is set automatically without adding commands to autoexec. 
Could you please take a look , check if it is correct or even check on other devices ex. TH06 .. 

ps. at the end one small fix for user_main.c


openshwprojects:
Very good job, that little bug with wifi state also explains some things...

I will check your changes soon, but my TH06 is not a battery powered device and TuyaMCU is not able to reset it anyway....

openshwprojects (approved):

PR #737, opened 2023-03-17
Tasmota Json correction for TasmotaAdmin

Dheenhasty:


openshwprojects:
Very good, but we need some docs. Would you like to add a Tutorial to our Smart Home Tutorials section about Tasmo Admin and OBK ? Thanks!

openshwprojects (approved):

Dheenhasty:
> Very good, but we need some docs. Would you like to add a Tutorial to our Smart Home Tutorials section about Tasmo Admin and OBK ? Thanks!

I will try to do this.
 There's a lot of work to do on tasmota compatibility.

On mqtt, tasmota use a discovery topic to declare component. And under stats/cmnd and tele it miss some pattern.

On http level there's pany value that are still static.

I'm wondering if it will not be better to implement some dedicated tasmota class cause it often push the same data

openshwprojects:
What do you mean with Tasmota class?

PR #740, opened 2023-03-18
SPI support for BK7231T (untested)

Diggar:
As I don't have hardware for testing it could work, but also could not work. Please use the code and try to fix it, if it is not working.

The code is using busy waits instead of semaphores. This could be improved, when the SPI is working.


openshwprojects:
Thanks, I replied there: https://www.elektroda.com/rtvforum/viewtopic.php?p=20497046#20497046

PR #744, opened 2023-03-21
Combined and outsourced energy calibration + Cleaned up headers/includes

Diggar:
A lot of duplicated code for calibration of power/energy sensors could be removed.
Tested calibration for BL0942 and it is working.

### DIFF
--- a/src/driver/drv_bl0937.c
+++ b/src/driver/drv_bl0937.c
@@ -7,6 +7,7 @@
 #include "../logging/logging.h"
 #include "../hal/hal_pins.h"
 #include "drv_public.h"
+#include "drv_pwrCal.h"
 #include "drv_local.h"
 #include "drv_uart.h"
 #include "../httpserver/new_http.h"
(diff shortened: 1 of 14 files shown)


Diggar:
The drv_pwrCal.c has to be added as source file for OpenW600.

openshwprojects:
Hello, how much flash memory did it save?

Seems like a great change, very valuable help, but was it tested on other devices or at least analyzed step by step?

I will try to analyze it tomorrow and maybe check on my BL0937 and BL0942 devices.... we need to make sure we won't break anything.

Diggar:
How can I see the memory/flash size? Is it this one?:
ug_file size: 623952 (before)
ug_file size: 623680 (after)
So it would be 272 Bytes less.

Yes, we have to be careful not to break anything. One problem was, that BL0937 and CSE7766 use calibration factors whereas BL0942 uses calibration dividers. I didn't change that, because else a new calibration (or a function to invert calibration factors/dividers) would have been necessary.

I checked only BL0942SPI. How do you analyze step by step?

Diggar:
@openshwprojects

Did have time to analyze the code or check with your devices? Isn't it possible with your windows port test suite? (I have only linux and don't have a virtual windows machine on my new notebook.)

Actually the changes are not so big:
- Calibration functions moved from separate drivers to drv_PwrCal file.
- Scaling is moved from separate drivers to drv_PwrCal file.
Both is tested and working with BL0942.

The only difference for the other drivers is, that they multiply the calibration value and the BL0942 divides by the calibration value.

Actually for me, reducing flash size was not the main goal, but removing duplicate code and making the code more understandable.

PR #747, opened 2023-03-25
Adding new driver for SGP30, GUVA and Generic ADC. 

Dheenhasty:


openshwprojects:
Hello, thanks, but please enable new drivers only for Beken, because on BL602 we will have reduce  binary size so OTA still works. I will update the OBK soon for that, I can save some bytes with string references.

Dheenhasty:
it's the case in drv_main, but maybe i miss other part ?  (i put them in #if defined(PLATFORM_BEKEN) || defined(WINDOWS) )

i was thinking about optimizing the whole ADC related driver with the generic one. it can save some space in memory.

openshwprojects:
I can see a little problem with proposed changes. Currently the IOR_ADC was just getting sample as integer and putting it in channel;. Now you've added a driver for that, but it basically duplicated the functionality... maybe it would be better to keep old IOR_ADC as is and maybe introduce a IOR_ADCVolts or smth like that... so they both are working... I am not sure yet.

Dheenhasty:
Hum, we can just add a flag who will switch between sample/mvolts. It's
just a formula to add in current adc function. So no need for my drv_adc.

Le dim. 26 mars 2023, 13:43, openshwprojects ***@***.***> a
écrit :

Dheenhasty:
i have done the Battery driver and the ADC driver having in mind to adapt to many chip. 
but, i have deep dive the HAL_ADC_Read function for each chip: 

BK7231  => return the sample bits
BL602 => not implemented
W800 => return the mv 
xr809 => not implemented

and seems like my driver will only work for BK Chip ..... that's not what i want to do in first time. 
you might want to uniform the ADC read for all the chip. i only have beken chip :'( so i can help on BK.
BK7231 can be easy to include conversion to mv as vref will always be 2400 for beken chip. 
if we include the mv conversion directly in hal ADC the adc driver is no more needed and the battery driver can be simplified. i can do a HAL_ADC_Read_voltage function who will be available only for BK and W800 

seems like only my driver use the READ and the generic channel read : 
```
src/hal/w800/hal_adc_w800.c:int HAL_ADC_Read(int pinNumber)
src/hal/win32/hal_pins_win32.c:int HAL_ADC_Read(int pinNumber) {
src/hal/hal_adc.h:int HAL_ADC_Read(int pinNumber);
src/hal/bk7231/hal_adc_bk7231.c://int HAL_ADC_Read(int pinNumber) {
src/hal/bk7231/hal_adc_bk7231.c:int HAL_ADC_Read(int pinNumber)
src/hal/xr809/hal_adc_xr809.c:int HAL_ADC_Read(int pinNumber) {
src/hal/bl602/hal_adc_bl602.c:int HAL_ADC_Read(int pinNumber) {
src/user_main.c:				value = HAL_ADC_Read(i);
src/driver/drv_guva.c:	g_uvvoltage = HAL_ADC_Read(g_pin_adc);
src/driver/drv_battery.c:	g_battlevel = HAL_ADC_Read(g_pin_adc);
src/driver/drv_battery.c:	g_battvoltage = HAL_ADC_Read(g_pin_adc);
src/driver/drv_adc.c:	g_adcvoltage = HAL_ADC_Read(g_pin_adc);

```

For the BL602 i have check the SDK ... and seems like the VREF and the init of ADC channel depend on many things ..... i understand why it's not yet implemented/

openshwprojects:
Yes, but IOR_ADC is already doing a HAL_ADC_Read without a driver...
image
so current version with a driver might have unexpected results. And it support multiple ADCs, not the single one, like with drv_adc.c approach.

Dheenhasty:
We say the same thing :) i will remove the drv_adc and add a volt conversion function in hal_bk7231 to simplify guva and battery driver

PR #748, opened 2023-03-26
Show more energy information on http site

Diggar:
Added following on http site (like in Tasmota):
- Apparent Power (Vrms * Irms)
- Reactive Power (Power that does no work in the load, but goes back into the grid)
- Power Factor (Active Power/Apparent Power)
- Energy Today
- Energy Yesterday

Tested and working.


openshwprojects:
Thanks. It should be okay and I'll merge it now.

PR #760, opened 2023-04-02
Increase precision of BL0942

Diggar:
BL0942 can average values for 400 ms or 800 ms. As the chip is read out every second, we get preciser values with 800 ms averaging. Additionally BL0942_UART was only read out once every two seconds.

Unfortunately I cannot test BL0942_UART as I don't have hardware. SPI is tested and working.


openshwprojects:
Very nice, I wasn't aware about such possibility. Would you be able to test with BL0942 soon if you get such device?

Diggar:
Sure, I can test it, when I have the hardware and was able to flash it. But it will arrive only in 5 weeks.

Diggar:
Finally I have a BL0942 UART device and was able to flash it with cloudcutter.

Tested and works for UART as well. (I also tested it by increasing the UART baudrate, which broke the communication and it works again, when increasing the URT baudrate to same value in Beken.)

openshwprojects:
Ok, thank you, let's give it a try.

PR #771, opened 2023-04-08
Removed old unautomated methods for power calibration + Use Tokenizer to save Flash

Diggar:
Removed old unautomated methods for power calibration.
Saving 192 Byte of Flash
Tested and working


Diggar:
Do we need to show "(changes sent 0, skipped 13008, saved 14) - BL0942SPI" in the http page? For me it sounds like debug stuff maybe for logging but not on the http page.

openshwprojects:
I actually wanted to remove IRef etc commands myself, so that's a good change.

Regarding changes sent counter - I am not sure. It is useful when you set change margines to reduce amount of MQTT publishes to HA. Thanks to this users can see that their margins changes have effect.

PR #789, opened 2023-04-14
feat: enable docker build on non amd64 platforms

no2chem:
This PR enables a docker build on non-amd64 platforms.

Previously, the docker build would error out on other platforms because i386 architecture isn't available on these platforms, which would pull a non-amd64 image by default. This PR adds an explicit request for amd64 platform so the correct image gets built.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM --platform=linux/amd64 ubuntu:22.04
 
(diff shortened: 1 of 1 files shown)


no2chem:
Apparently I failed to read the readme, but I think the tag also makes it more straightforward to run on anything other than an amd64 machine.

openshwprojects:
Thanks. I am not able to test it as I haven't looked into docker build yet, but I will merge it anyway.

Issue #810, opened 2023-04-26
Tasmota topics publish interval


serek4:
**Describe the bug**

VCPPublishIntervals command doesn't limit mqtt messages published to Tasmota topics.

**Firmware:**
- Version 1.17.29
- Chip/model: BK7231N
- BL0937 driver
- startup command: backlog startDriver NTP; ntp_timeZoneOfs 2; SetupEnergyStats 1 60 60;

1. enable: Flag 30 - [MQTT] Enable Tasmota TELE etc publishes (for ioBroker etc)
3. send command: VCPPublishIntervals 30 60
4. tele/# topics are still published in ~3s intervals

openshwprojects:
You are correct, but we will need another command for tele.

What about adding TelePeriod command? I can add this tomorrow morning.

serek4:
Are you thinking of the same implementation as VCPPublishIntervals ie. `TelePeriod [minInterval] [forcedInterval]`, or just simple `TelePeriod [interval]`?

openshwprojects:
What would fit your need better?
From what I can see currently, we do two types of TELE:
- tele/ sensor, done every 3 seconds if sensor is running (any sensor)
- tele / state, done every 120 seconds
![image](https://user-images.githubusercontent.com/85486843/234808721-4fcae983-4a66-4f18-9b45-e161789b0ed9.png)
What kind of approach would you like use to take with tele? 

Maybe do not use name "TelePeriod" because it may confuse people.
Maybe TTP (TasTelePeriod) with two arguments - first argument interval for tele/sensor, second for tele/state?

Or should be publish sensor only when it changes?

serek4:
> Maybe TTP (TasTelePeriod) with two arguments - first argument interval for tele/sensor, second for tele/state?

That sound good

> Or should be publish sensor only when it changes?

Maybe add command similar to VCPPublishThreshold?

openshwprojects:
Well, if you want a command similiar to VCPPublishThreshold, then maybe I will just plug the TELE SENSOR into the VCPPublishThreshold code block? That way VCPPublishThreshold will work for TELE.

The only issue is that I will have to add some other condition for sensors like DHT11,, but I can do that..

serek4:
Let's stick with TELE publish in intervals, adjustable by your proposed TasTelePeriod command.

openshwprojects:
Pushed: 
TasTeleInterval [SensorInterval] [StateInterval]
Can you check?
[...]

PR #816, opened 2023-04-28
fixes for bugs from #810

serek4:
fixes for bugs from [this comment](https://github.com/openshwprojects/OpenBK7231T_App/issues/810#issuecomment-1525941122)
- fixed tasmota tele/sensor sent additionally with tele/status
- ~~fixed tasmota topics sent with mqtt_broadcastInterval when flag 2 is enabled~~

plus
- tasmota topics sent on mqtt connect only when flag 10 is enabled
- fixed tasmota tele publish +1s from set interval

### DIFF
--- a/src/mqtt/new_mqtt.c
+++ b/src/mqtt/new_mqtt.c
@@ -1261,6 +1261,7 @@ OBK_Publish_Result MQTT_ChannelPublish(int channel, int flags)
 	}
 
 	MQTT_BroadcastTasmotaTeleSTATE();
+	MQTT_BroadcastTasmotaTeleSENSOR();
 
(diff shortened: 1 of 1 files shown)


serek4:
No, it were not intendent.
That line was removed to not send tasmota topics with mqtt_broadcastInterval when flag 2 is enabled.
I will fix that.

serek4:
reverted b40665ae336a9b5fa5829b34b1a41acc28b69cfb, so everything should work as before

openshwprojects:
Seems okay now to me. So, what do you say, should I merge it already?

serek4:
Yes. I'm ok with changes as well.

PR #827, opened 2023-05-14
BL0942: Increased precision by using IC energy

Diggar:
As mentioned before, the the BL0942 is now summing up the power for 800 ms, but we read out the power only every second. The IC has an build in energy counter. This commit makes use of the energy counter (read out energy) instead of calculating the energy from the power. This improves the precision.

Tested for both, UART and SPI device. For my devices the resolution of the energy is 0.18 Wh (UART) and 0.34 Wh (SPI). The precision depends on the circuit connected to the BL0942. I.e. it can be with current shunt (resistance value) and current coil (windings).

The energy counter of the IC is correlated to the power. Thus the energy counter uses the power calibration. No additional calibration is needed.

### DIFF
--- a/src/driver/drv_bl0937.c
+++ b/src/driver/drv_bl0937.c
@@ -1,5 +1,7 @@
 #include "drv_bl0937.h"
 
+#include <math.h>
+
 #include "../cmnds/cmd_public.h"
 #include "../hal/hal_pins.h"
 #include "../logging/logging.h"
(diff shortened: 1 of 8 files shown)


openshwprojects:
Hello, thanks, let me know how stable is this change and is it production ready and then I will merge it

Diggar:
Both my devices ran at least half day (they are off, while I'm not working on them). Everything worked fine. Also reboots (OTA flash), without reset of the BL0942 worked fine (the BL0942 continues measuring the energy, while the µC is rebooting).
Thus, in my opinion the commit is stable and production ready.
I also checked the code again.
Ready to merge :)

PR #832, opened 2023-05-18
Bugfixes in UART: No more corrupted packages

Diggar:
- Fixed bugs in uart_driver which caused corrupted uart reads, when the ring buffer came to its "overflow".
- Renamed two functions to clearer names:
    - UART_GetNextByte -> UART_GetByte (no index is moved to a "next" position)
    - UART_AppendByteToCircularBuffer -> UART_AppendByteToReceiveRingBuffer (there is already UART_InitReceiveRingBuffer)
- Tested with BL0942, which produced checksum errors before the commit. After commit everything works fine, and no more checksum errors occur.

### DIFF
--- a/src/cmnds/cmd_main.c
+++ b/src/cmnds/cmd_main.c
@@ -380,7 +380,7 @@ void CMD_UARTConsole_Run() {
 
 	totalSize = UART_GetDataSize();
 	while (totalSize) {
-		a = UART_GetNextByte(0);
+		a = UART_GetByte(0);
 		if (a == '\n' || a == '\r' || a == ' ' || a == '\t') {
 			UART_ConsumeBytes(1);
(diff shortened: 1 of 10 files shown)


openshwprojects:
It seems like a very good change, maybe there was indeed a bug in the ring buffer code. There are many whitespace changes in the diff, which makes it hard to follow what exactly was changed, but it seems that buffer wrapping was not correct. I will merge it right now, thanks!

Diggar:
> There are many whitespace changes in the diff
Yes, sorry for that. VS Code auto formats all changed files when doing a search and replace. I manually (CTRL + Z) changed all files back (the formatting) but seems like I forgot it in drv_uart.c. I will take more care in the future.

PR #843, opened 2023-06-01
Documentation Update

TomTheGeek:
Fixed one typo in the battery commands, also added description for Map command. Updated docs script so that commands are listed alphabetically.


openshwprojects:
Thanks, there seems to be one issue:
image
Please remove the redundant comments and just make sure there is one doc entry for Post command and then I think it's ready to merge.

Issue #895, opened 2023-08-09
Adjust HomeAssistant EntityName to confirm to changes in 2023.8.1


iprak:
**Describe the bug**
HomeAssistant https://www.home-assistant.io/blog/2023/08/02/release-20238/#breaking-changes has a breaking change in MQTT Entity Name - _The entity's name property only identifies the data point represented by the entity, and should not include the name of the device or the type of the entity. So for a sensor that represents the power usage of its device, this would be “Power usage”._

There was a PSA announcement about the same - https://community.home-assistant.io/t/psa-mqtt-name-changes-in-2023-8/598099

**Firmware:**
- Version 1.17.208
- Device? Any
- Chip/model: Any

**Screenshots**
One will see a notification like this under Settings. The message clarifies that after 2024.2.0 this can result in duplicate entities.
![image](https://github.com/openshwprojects/OpenBK7231T_App/assets/6459774/5cc50ead-d2be-43d4-92a5-ca39975eb4c4)

**Additional context**
This can be addressed by changing
`sprintf(g_hassBuffer, "%s %s", CFG_GetShortDeviceName(), CHANNEL_GetLabel(index));` 
to
`sprintf(g_hassBuffer, "%s", CHANNEL_GetLabel(index));`

Here is the MQTT payload for a fixed item which just has '0' as the entityName.
![image](https://github.com/openshwprojects/OpenBK7231T_App/assets/6459774/28240907-a410-49c1-a00d-b39d6230d462)

We can also return `None` as the name if the entity is the main device for LIGHT_PWMCW case where the entityName is the same as deviceName.

openshwprojects:
Thanks, I saw that mentioned somewhere and was not sure how to fix it. So that rename sprintf in HA discovery is all we need?

iprak:
Sort of.

If name was not assigned then HomeAssistant will auto assign it. For example, for Relay, the default name is "MQTT Switch". The default name for PWM styled entity is [MQTT LightEntity](https://github.com/home-assistant/core/blob/e9f9c7799a5b81be8ce95fcb63bdb969a5bdc3be/homeassistant/components/mqtt/light/schema_basic.py#L135). Simple Light/LED are just switch at HomeAssistant end.

I previously thought that assigning "None" to name will be sufficient for PWM case which currently uses the device shortName as the name. But that does not work, it ends up using the literal "None" as the name. None only works from within custom entities. 

[...]

PR #899, opened 2023-08-14
Remove duplicate prefix

iprak:
The merge removes the CFG_GetShortDeviceName prefix to entityName to address #895. PWM entities just have "Light" as the name. The entity_id remains unchanged. Further customization of name/icon can be done in HomeAssistant.

This has been tested on 7231T and W600 switches. I don't have other devices.
It seems that just pushing HomeAssistant discovery is not sufficient to resolve the warning, a restart at Homeassistant end might be necessary.

### DIFF
--- a/src/httpserver/hass.c
+++ b/src/httpserver/hass.c
@@ -188,61 +188,59 @@ HassDeviceInfo* hass_init_device_info(ENTITY_TYPE type, int index, char* payload
 	case LIGHT_PWM:
 	case RELAY:
 	case BINARY_SENSOR:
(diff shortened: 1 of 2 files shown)


openshwprojects:
thank you, now we can connect to the HA of the future

jenshenk:
Thanks for addressing this so quickly! Can confirm it working with BK7231N devices, too.

jenshenk:
Just finished updating all of my devices and the error message is gone! FYI: With two devices I had to manually initiate a Home Assistant discovery even after the firmware update, but after that HA stopped complaining about those ones, too.

PR #901, opened 2023-08-14
Update drivers.md

analystcmyk:
added ChargingLimit


openshwprojects:
Thanks, however, this is supposed to be done automatically, based on the comments from code. I can copy your comment to code and run docs update to make it work. Otherwise next docs rebuild would overwrite your changes.

openshwprojects:
Here is how we add desc:
https://github.com/openshwprojects/OpenBK7231T_App/commit/fae6d57d1c2eb7f9f7c118747ca5c9a15d645eaf

openshwprojects:
and then we run node scripts/getcommands.js to generate files:
https://github.com/openshwprojects/OpenBK7231T_App/commit/c88ba0f8ea93817fafd2d038a56a10bb3d0c4401

PR #906, opened 2023-08-19
KP18058: LED current setting, PWM dimming

miegl:
- ~~RGB LEDs at full power~~
- LED current now configurable (0-31)
- PWM dimming instead of analog


openshwprojects:
That's very interesting, can you say few words about how it works?

Also, is full power always preferred on all devices?

miegl:
- Previously, the RGB LED current was very low (set to 1 in fact), after looking at https://www.elektroda.com/rtvforum/topic3985037-30.html#20663929 I figured out setting the RGB LED current. It can be set to 0-31, seems like the value is in mA.
- I've set the default RGB LED current to 14, which corresponds to the red color sample in picture on forum.
- The default CW LED current of 30 also corresponds to the intercepted I2C communication on forum.
- CW LED current and RGB PWM mode referene taken from [http://www.kiwiinst.com/newsinfo/2765480.html?templateId=510705](http://www.kiwiinst.com/newsinfo/2765480.html?templateId=510705) and this [image](http://nwzimg.wezhan.cn/contents/sitefiles2052/10261492/images/30338065.jpeg).

Also, KP18058 and KP18068 share the same exact protocol, so the unfinished KP18068 driver can be probably removed.

openshwprojects:
Thank you, that's a great progress. How do you know that they share the same protocol? The linked page doesn't seem to specify it, was it tested?

miegl:
I have and tested the driver on KP18068, findings on forum for KP18058 (protocol and values) seem to correspond. Also the linked page contains this image:
obrazek

The only difference between KP18058 and KP18068 seems to be the Power factor (0.5 vs 0.7), so I think it's pretty safe to assume the protocol is exactly the same.

openshwprojects:
Love it! Thank you!

PR #917, opened 2023-09-08
Add parse header without value

avigeilpro:


openshwprojects:
Hello, can you give some explanation of logic here?

avigeilpro:
I opened an issue where i explained all. The problem come from HA, with the /config/www directory. If u put a file and access it from http://<ip>/local/myfile.xyz one of the headers (server) has no value, it generate an error (can't parse). I just add a few lines to handle when a header has empty value.

openshwprojects:
Very well. The explanation seems to be enough to merge it right now. Thank you!

PR #919, opened 2023-09-11
Update constant detections in MQTT (v2)

avigeilpro:
Solved issue with long arguments.
Made possible to put curly brackets around any constant.
self tests successfull (had issue to configure visual studio because of missing nfd_d.lib but found how to compile)
tested on BK7231N and BK7231T with success.
here some examples of possibilities :
`publish OpenBeken/test ${IP}_${CH4} 1`
`publish OpenBeken/test/${IP} ${CH4} 1`
`publish OpenBeken/test/$IP $CH4 1`
`publish OpenBeken/test ${IP}_ON 1`

if the constant is the last before a space, you can write it like this `$IP` or `${IP}` (both works)
if you have text after the constant, you must put the curly brackets like this `${IP}_ON`

token with constants are limited to 40 characters (after expand). we could make it longer but we will have to extend the size of array g_argsExpanded[MAX_ARGS][40]. I assume it needs to stay so because of memory limitation.

If you see a case i didn't cover, please say me.

### DIFF
--- a/src/cmnds/cmd_tokenizer.c
+++ b/src/cmnds/cmd_tokenizer.c
@@ -84,6 +84,7 @@ const char *Tokenizer_GetArgExpanding(int i) {
 	char *strToken = strtok(tokLine, separators);
 	while (strToken != NULL) {
 		//build tconst with ${<token>} and try find it
+		int len_tconst=0;
 		char tconst[20] = "${";
 		strcat(tconst, strToken);
 		strcat(tconst, "}");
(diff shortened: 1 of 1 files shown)


openshwprojects:
I am wondering whether we could merge it with CMD_ExpandConstantsWithinString

avigeilpro:
I took a look at CMD_ExpandConstantsWithinString, one problem that i see is that my code is more "universal", it means that on the self test "ba$CH11ha" for example, it will try find the constant named $CH11ha, it dont know that a channel can have only 2 digits, all it do is cutting with a patern. To have the same result with my code you should put "ba${CH11}ha". I made this because if somebody want send the value of CH1 followed by a "0" for example, with CMD_ExpandConstantsWithinString it's not possible (or i missed something). If this is a problem, i can try do something, but i think this solution is better for futur development, and it is not something new as it is the way we do in JS to include variables in strings.

avigeilpro:
This depend of your will. I don't know if the formatting like "baCH11ha" is already used or not. If not, i think you should migrate to "my" solution, but if it is used and is difficult to modifie, you should keep both, and slowly migrate all the function wich use the original formatting. Say me what you prefer, and i will rewrite the code in consequence.

PR #1004, opened 2023-12-20
CSE77xx fix

artin961:
Recently I purchased DIN rail mountable energy monitor based on SCE7759B and CB2S. After testing with official release i spotted two issues: 

Current measured was showing as 0.0A and after attempt to calibrate with 100w bulb and measuring using my Rigol DM3058 it started showing as "nan". After compiling my own firmware I found the issue. When the real current was rounded to 0 because it was integer. This resulted in devision by zero during calculation of the calibration constant, and also lowering the percision of the instrument. Changing raw current to float fdixed the issue.

The other issue I have spotted is that some time there were chacksum errors because the packet start was mis identifyed, because 0x5A was appearing elswere in the packet. This was easily corrected with comparing the whole header(2 bytes 0x55 0x5A according to the datasheet). 

This is tested to be working fine with CSE7759B, and shouln't break the rest. 
image_2023-12-20_16-17-19

### DIFF
--- a/src/driver/drv_cse7766.c
+++ b/src/driver/drv_cse7766.c
@@ -19,7 +19,7 @@ int CSE7766_TryToGetNextCSE7766Packet() {
 	int cs;
 	int i;
 	int c_garbage_consumed = 0;
-	byte a;
+	//byte a;
 	byte checksum;
 	int CSE7766_PACKET_LEN = 24;
 	byte header;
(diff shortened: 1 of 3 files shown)


openshwprojects:
Thanks, but are you sure that having float there will help?
We have integer value from CSE, and we cast it to float while diving it by float cal here:
image
I am just intriguided...

openshwprojects:
Or are you refering to that part?
image

openshwprojects:
I think I can see it now.... so it's a float and then get casts back to integer?

artin961:
I tested it with the prints  of raw_unscaled_power and it was 0.4 for the current.

openshwprojects:
are you sure this will not break situation where other drivers gives very big integer that cant fit in float|"?

artin961:
I cannot be 100% sure as I don't have the hardware to test. Maybe someone with the other types of IC's could give a feedback. I cannot see a reason why other drivers could give very large integers. This driver gives the current in Amps, if the others use miliamps then this one needs to be adapted to convert them without loosing percision.

openshwprojects:
Okay, I think I will merge it tomorrow and keep that change in mind. We may also just change that float to double for extra precision...

Issue #983, opened 2023-11-16
New sensor CHT8310


Tobias0110:
I want to thank you all for the great firmware you created here. I successfully installed it on many devices.

Recently I came across a temperature and humidity sensor that is currently not supported (SensyLink CHT8310).
I tried to use it with the settings for the SensyLink CHT8305 but this results in broken measurement data, so it seems the data format is different. I can get a temperature reading that changes when heating / cooling the sensor but the value is negative at room temperature (-21 is displayed at about 22°C room temperature, moves towards 0 when heated). The humidity reading jumps randomly.

Sensor is connected to:
P20 [SCK]
P22 [SDA]

Modul in use: [tuya CBU](https://developer.tuya.com/en/docs/iot/cbu-module-datasheet?id=Ka07pykl5dk4u)

Datasheet is available on the [SensyLink Website](https://www.sensylink.com/html/web//product/wsdcgq/cwsdcgq/1663024476742897666.html).

[Bought on Aliexpress](https://de.aliexpress.com/item/1005005869874007.html?spm=a2g0o.order_list.order_list_main.5.12425c5fdgkmEW&gatewayAdapt=glo2deu).

![bk_start](https://github.com/openshwprojects/OpenBK7231T_App/assets/11034331/f0f12575-ea9b-4606-a60d-88e23545e9b1)

![signal-2023-11-16-220923_003](https://github.com/openshwprojects/OpenBK7231T_App/assets/11034331/d7e1d23e-3deb-4987-8c4d-791808da22f6)

![signal-2023-11-16-220923_002](https://github.com/openshwprojects/OpenBK7231T_App/assets/11034331/cdd66b55-0e8f-408e-83a3-38db853ba34b)

openshwprojects:
The datasheet redirects me to login form , are you able to download it?

Tobias0110:
[CHT8310_Datasheet.pdf](https://github.com/openshwprojects/OpenBK7231T_App/files/13398333/CHT8310_Datasheet.pdf)

Yes, please see attached. You can create an account with random information. They don't require you to confirm the E-Mail address.

openshwprojects:
Is the datasheet only those 4 pages you attached?

Tobias0110:
Yes this is all for now. I am trying getting the full version containing more information about I2C.

openshwprojects:
Maybe just I2C address is wrong? Maybe we could try doing I2C scan?

Tobias0110:
[...]

PR #1008, opened 2023-12-23
Add cht8310 support

artin961:
Added support for the new SensyLink sendor CHT8310 and this should fix [ISSUE-983](https://github.com/openshwprojects/OpenBK7231T_App/issues/983)

I have the sensor and tested it. Compatibility with the old sensor shold be maintained, but I don't have it to test. Please confirm If you have the hardware to test, before merge. 

The new sensor has more features and the way to calculate measurement results is different. I put an if there depending on the sensor type. On the old one if can be confirmed reading a sensor ID from the IC. If it is not 8305, the code assumes it is the new one.  

I have utilised the low power mode of the IC where it only draws 50nA, and makes measurements on request. This is accomplished writing anything to the "Oneshot" register. If the sensor is the old one this is skipped, as well as the configuration on init. 

Another major difference is that if you read 4 bytes from 0x00 to get the two 16 bit registers in one go there is parity byte and 0xff as bytes 3 and 4. To read the humidity data from 0x01 a second read needs to be made moving the pointer to it and reading 2 bytes. Again if it is the new sensor the code overwrites bytes 3 and 4 with the humidity data as expected. 

Before merge needs to be confirmed that sensor identification logic is working and reading from the old sensor is not broken. I dont have the hardware to test.

### DIFF
--- a/src/driver/drv_cht8305.c
+++ b/src/driver/drv_cht8305.c
@@ -27,7 +27,7 @@ static void CHT8305_ReadEnv(float* temp, float* hum) {
 	uint8_t buff[4];
 	unsigned int th, tl, hh, hl;
 
-	if(sensor_id!=0x8305)
+	if(sensor_id==0x8215)
 	{
 		//Make dymmy write to Oneshot register to trigger a measurement, sensor is otherwise sleeping
 		Soft_I2C_Start(&g_softI2C, CHT8305_I2C_ADDR);
@@ -48,7 +48,7 @@ static void CHT8305_ReadEnv(float* temp, float* hum) {
 	Soft_I2C_Stop(&g_softI2C);
 
 	//In case we have the new sensor 8310, overwrite humidity data reading it from 0x01, as it cannot be directrly read from 0x00, there is parity
-	if(sensor_id!=0x8305)
+	if(sensor_id==0x8215)
 	{
 		Soft_I2C_Start(&g_softI2C, CHT8305_I2C_ADDR);
 		Soft_I2C_WriteByte(&g_softI2C, 0x01);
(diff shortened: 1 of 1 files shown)


openshwprojects:
Thank you! I will do a code review now, but sadly, I don't have the old sensor as well...

openshwprojects:
Hey, wouldnt it be safer to swap:
sensor_id!=0x8305
to:
sensor_id==0x8310
?

artin961:
It should be safer but unfortunately when I read the same register on the new sensor the data i got is not the sensor version and it is not  documented in the datasheet. I will try to read it as a separate register just in case it is the same as the humidity issue and that is the reason it is not giving the expected result.

openshwprojects:
Wait, I don't understand. What is the value of sensor_id in case of your new sensor? It's 0x8310 , right?

artin961:
on the old one is expected to be 8305 according to the datasheet. In the new one When reading 4 bytes from 0xfe so reading the 0xff there was unexpected data and when checked in the datasheet there is no mention of sensor version there. Now I'm trying to read 0xff as a separate register to confirm that is the case.

openshwprojects:
Ah, I see, this makes sense now,,.,,, ok please do check on your side, and otherwise I will just merge it as is and just ask users on forum to check

artin961:
Reading the two as separate register gives me: 
Info:SENSOR:DRV_CHT8305_init: ID: 59 59 82 15 

8215 should be from 0xff but doesnt match 8310 and as it is not documented in the datasheet I'm not sure how dependable it is. 


So we have a two choises: 
1. to use it even it is undocumented to identify the new sensor
2. leave it like that and verify the logic identifies the old one right. 

Maybe the second one is safer to guarantee that will not break the old devices at least.  (senor_id==0x8215).

openshwprojects:
Ok so lets try second one, so now I  can mergei t? I can fix BL602 compile error hm

artin961:
OK give me a second and I will try the second one and push changes.

openshwprojects:
ok can I merge and ask users to test?

artin961:
I tried the first option just as a POC and it prooved to be safer because I identified a mistake I have made when combining the two bytes of sensor_id. Also changed the debug print on driver init to display the actual sensor_id variable. 

Info:SENSOR:DRV_CHT8305_init: ID: 59 59 8215 

The code now should only use the new logic if the ID is 0x8215, with the exception that is reading the actual id separately. This shouldnt cause any issues. 

I saw there are more sensor on sensylink website so in long run should be easy to handle more sensors in the future. 

Now the code is tested again and should be able to be merged.

openshwprojects:
Thank you, i will merge now, let's talk futher in https://www.elektroda.com/rtvforum/viewtopic.php?p=20867267#20867267

jimbo425:
I have TH08 battery powered temperature sensor with cht8310 sensor. The log say:DRV_CHT8304 :(

jimbo425:
I have TH08 battery powered temperature sensor with cht8310. The log say:DRV_CHT8304. It is ok?

PR #1030, opened 2024-01-14
Get sunrise/sunset

ilengyel:
Get the next sunset or sunrise values as TimerSeconds to allow configuring a light's initial sate.


openshwprojects:
That's very interesting, I also saw this feature being discussed on forum. Is it ready to use? Did you run windows self tests?

ilengyel:
> That's very interesting, I also saw this feature being discussed on forum. Is it ready to use? Did you run windows self tests?

Yea, I did post a query on that topic a few weeks ago, and I was going to add a link this PR but @rickbronson closed that topic. I did get the windows self tests running (was bit of a mission as it was not obvious where to collect the libs_for_simulator files), but found the existing tests were failing due to some drift by 1 minute. Not sure what is causing the drift, will investigate a bit more.

As for implementing `addClockEvent <secondsFromMidnight>` so that calculated time values could be used more easily,  I think I will do in a separate PR, and will try to adjust this autoexe script to avoid using TimerSeconds type somehow.

openshwprojects:
https://github.com/openshwprojects/obkSimulator/tree/main/simulator

rickbronson:
During my testing I did find that the times were +-2 minutes of "real"
sunrise/sunset.  "real" being Google but if you use a different source you
might get a different result :)

  Rick

openshwprojects:
I can unlock elektroda topic if needed, can you post update there?

ilengyel:
It turns out it was all due to the timezone difference of my windows machine which resulted in various other assert fails. Please test on your machines as I updated all usages of `localtime(sourceTime)` to be `gmtime(sourceTime)` where my assumption is they both behave exactly the same on the embedded devices, but on windows machines (selftest & simulator) we should use gmtime for consistency.

PR #1043, opened 2024-01-20
Try add bl0937 for BL602

giedriuslt:
Trying to make bl0937 work on BL602


giedriuslt:
Compiles, but does not work :/

giedriuslt:
This now seems to be working, but there are two issues, these hal includes should not be there, and there is no proper way remove interrupt handler in sdk.

PR #1045, opened 2024-01-21
addClockEvent - allow TimerSeconds

ilengyel:
Or allow single integer value for the time parameter that would represent a TimerSeconds type from midnight. This would allow a script like: 

```
setChannel 10 $sunset+1800
addClockEvent $CH10 0xff 5 echo timeExpanded
```

### DIFF
--- a/src/selftest/selftest_clockEvents.c
+++ b/src/selftest/selftest_clockEvents.c
@@ -3,13 +3,24 @@
 #include "selftest_local.h"
 #include "../driver/drv_ntp.h"
 
+static void ResetEventsAndChannels(int eventsCleared) {
+	SELFTEST_ASSERT(NTP_ClearEvents() == eventsCleared);
+	SELFTEST_ASSERT(NTP_PrintEventList() == 0);
+
+	CMD_ExecuteCommand("setChannel 1 0", 0);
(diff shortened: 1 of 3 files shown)


src/selftest/selftest_clockEvents.c
```diff
@@ -119,6 +110,25 @@ void Test_ClockEvents() {
...
+
+	// Test with calculated times
+	ResetEventsAndChannels(3);
+
+	CMD_ExecuteCommand("setChannel 4 13", 0);
+	CMD_ExecuteCommand("setChannel 5 55", 0);
+	CMD_ExecuteCommand("setChannel 6 $CH4*3600+54*60+59", 0);
+
+	CMD_ExecuteCommand("addClockEvent $CH6 0xff 4 backlog addChannel 1 10; echo test event for 4", 0);
+	//CMD_ExecuteCommand("addClockEvent $CH4:$CH5 0xff 5 backlog addChannel 2 20; echo test event for 5", 0);
```
ilengyel, line 122: To allow this, it is not so simple as enabling the `TOKENIZER_ALTERNATE_EXPAND_AT_START` flag for the string tokenizer. If we wish to pursue this notation, there would need some modification in the tokenizer code.

PR #1052, opened 2024-01-28
Implement full flash read for BL602

giedriuslt:
This adds ability to read full flash from BL602. Config read wont work, because it's stored differently.
Also fixes memory leak which was probably causing issues for bk7231


openshwprojects:
Wow, great spotting, thank you! So that's why I had reboots while reading Tuya config? Well, I haven't spotted it myself.

Just to be sure - did you check it on both BK and BL602?

giedriuslt:
Just tested on bk7231n, reads config fine does not crash.

openshwprojects:
Thank you, I am very happy with your contributions, if you need any help with futher improvements, ask me on forum.

PR #1053, opened 2024-01-28
Implement reboot reason for BL602

giedriuslt:
Shows reboot reason similar to how Beken does


giedriuslt:
Yes, longest reboot reason posible is BL_RST_HARDWARE_WATCHDOG, which is 24 characters long

giedriuslt:
Tested release version, works!

PR #1054, opened 2024-01-30
Start BL0937 driver for LN882H

giedriuslt:
Does not build due to missing driver functions, not sure why

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -24,4 +24,4 @@
 	branch = master
 [submodule "sdk/OpenLN882H"]
 	path = sdk/OpenLN882H
-	url = https://github.com/openshwprojects/OpenLN882H.git
+	url = https://github.com/giedriuslt/OpenLN882H.git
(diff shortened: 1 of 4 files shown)


giedriuslt:
Changed to use normal interrupts instead of external ones (these are maybe to wake up from deep sleep, not sure), and a PR to enable building driver here https://github.com/openshwprojects/OpenLN882H/pull/5

openshwprojects:
I have accepted the SDK PR but I am still unable to update submodules on main repo, can you try? It's strange, it worked for me many times before, am I  missing something?

giedriuslt:
Did an update in my  branch, lets see if it works.

giedriuslt:
Is there a guide how to build this locally?, It's messy to do it this way.
I added few more drivers in here https://github.com/openshwprojects/OpenLN882H/pull/7
It seems unexpectedly device groups driver is required too.

openshwprojects:
it should not be required, maybe we're missing an #ifdef somewhere to disable device groups in certain setup?

openshwprojects:
pwrCall is needed for power calibration, NTP also seems to be needed, it's for today-yesterday calculations

openshwprojects:
https://github.com/openshwprojects/OpenLN882H/blob/master/english_lightningsemi_sdk_cross_build_setup.pdf

giedriuslt:
well its a mess, it needs led drivers, then tyua mcu drivers, and now uart driver

PR #1055, opened 2024-01-31
Add SM16703P to DDP

zuckschwerdt:
I have a "LSC Battletron LED Bar" with BK7231N and 16 what seems to be SK6812 addressable RGB LEDs.
It works well enough with the SM16703P driver (but some minor timing problems).

This adds the DDP capability to the SM16703P driver by sending the DDP frame data along.
The needed prototype gets added to and used from `driver/drv_public.h"`, not sure if that is the right header for this.


zuckschwerdt:
Sure, happy to share. The SPI seems to have some flaws. There is quite some jitter and worse there is a long dropout around the end of the third pixel (`2+3*3*4`=38 bytes in). Also the reset does not take. Unmodified I only get 12 of the 16 pixels on the first push, then subsequently 4 older pixel and 12 new ones in each frame.

I added a long preamble (extended the 2 bytes to 62) and postamble (4 pixels=48 bytes) which seems to fix most of the issues.

openshwprojects:
That SPI preamble idea seems good ,can you commit the code in the another PR?

openshwprojects:
I was debugging it by doing a single SPI DMA transfer every 5-10 seconds or so and I noticed that randomly there is a strange delay at the start of some frames. I am not sure if it's affected by doing two frames without delay.

If your "postpend/flush" (not prepend workaround works, can we merge this PR now and continue into another PR?

If you have a Google Chat or Discord we can also chat there, I can setup my scope again and help looking into it.

zuckschwerdt:
I have 3 commits:
- Add SM16703P to DDP (this)
- Add SK6812 support to SM16703P driver
- Fix SM16703P SPI quirks

This PR is working and can be merged. I will open two more PRs then.

zuckschwerdt:
I didn't see https://www.elektroda.com/rtvforum/topic4033333.html before, but my tests are with the same device and it works well now. (My startup is simply `backlog POWER TOGGLE; startDriver SM16703P; SM16703P_Init 16 SK6812; startDriver DDP;`)

openshwprojects:
S o let's merge it and try. Thank you!

openshwprojects (approved):

PR #1062, opened 2024-02-04
Fix OTA for BL602

giedriuslt:
Change to erasing flash in smaller chunks, this avoids issues with http connection, because erasing seems to stop everything.
Makes it faster usually too, finishes in 5.5 seconds. Tested by creating almost 800kb image and works without any issues.

There is a bug in http_rest_error function that it does not work correctly in bl602 and just resets connection, I will try to fix later.


giedriuslt:
Tester release version, works fine

PR #1065, opened 2024-02-05
Add internal temperature for Beken chipd

giedriuslt:
This uses api from beken to get internal temperature. 
However in my plugs it reports ~270. Could it be Celsius * 10?
There is these defines and checks 
`#define ADC_TEMP_VAL_MIN                            (50)`
`#define ADC_TEMP_VAL_MAX                            (700)`

Seems strange for Celsius *10, so not sure


giedriuslt:
I can, I you think it's useful. I'm not sure if it's Celsius...

openshwprojects (approved):

PR #1066, opened 2024-02-07
Make temperature readable

giedriuslt:
So if we are showing temperature now, it should be done properly with units, this fixes that.
Maybe we should move it to a driver?

### DIFF
--- a/src/httpserver/http_fns.c
+++ b/src/httpserver/http_fns.c
@@ -756,15 +756,15 @@ int http_fn_index(http_request_t* request) {
 	{
 		UINT32 temperature;
 		temp_single_get_current_temperature(&temperature);
-		hprintf255(request, "<h5>Internal temperature: %lu</h5>",
-			temperature);
+		hprintf255(request, "<h5>Internal temperature: %.1f°C</h5>",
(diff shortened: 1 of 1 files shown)


giedriuslt:
For bl0937 with LN882, the PR is still here https://github.com/openshwprojects/OpenBK7231T_App/pull/1054 (and I updated it according to the user files), so it should be good, but I did not finish the required sdk changes.

MaxineMuster:
> Btw, what about that BL0937 + LN882 combo? I have read that one user from germany managed to get it running, or should I look into it myself?

That was me ;-)

Though I have a git account, I'm still struggling in using it, so I could only upload the files I changed to the forum and not put it proper to git, as I should have done.

So, yes, its working for me, but I had to add a lot of "unwanted" drivers as well to get it compiling (starting with "drv_tuyaMCU" and "drv_ntp" and then, probably by adding theese, there were other dependencies...)
I didn't find the time yet to see, wether/how I could reduce this, to the drivers really needed. Probably just a bunch of missing `#ifdefs` but then I'm not sure, if I break anything else.

PR #1067, opened 2024-02-08
Add temperature to BL602

giedriuslt:
Adds temperature to bl602 too. Seems not very stable unfortunately.


openshwprojects:
Thanks, but I've added one little fix. I've encapsuled all temperature codes into if not safe mode, not just beken one.

PR #1073, opened 2024-02-10
fix backwards binary sensor payload on/off values

airdrummingfool:
I noticed that when I configured a binary motion sensor on my BK7231T device, it was showing up inverted in Home Assistant when I used MQTT Discovery. I checked the discovery payload and found that the `pl_on` and `pl_off` values were inverted, even though the binary motion sensor on my device was not configured to be inverted. I checked the code and it looks like it is set up backwards.

Thanks for this great project!

### DIFF
--- a/src/httpserver/hass.c
+++ b/src/httpserver/hass.c
@@ -395,8 +395,8 @@ HassDeviceInfo* hass_init_binary_sensor_device_info(int index, bool bInverse) {
 	const char *payload_on;
 	const char *payload_off;
 	if (bInverse) {
-		payload_on = "0";
 		payload_off = "1";
+		payload_on = "0";
 	} else {
 		payload_off = "0";
 		payload_on = "1";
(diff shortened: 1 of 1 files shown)


openshwprojects:
I can see how that can be a problem, but how do we handle now people who have already configurations to respect it?

airdrummingfool:
That's a good question.

I would imagine that existing users would be unaffected until a new discovery payload was pushed to MQTT, but I don't know how often that happens. If it is only when discovery is started manually, perhaps we could recognize the use of a binary sensor and show a warning in the UI? That won't work if it gets updated more often (e.g. every device power cycle or HA reboot or anything like that).

You can call it out it as a breaking change in the release changelog (maybe put it at the top or in bold letters or something), but I don't know how often people check the release notes before upgrading. Maybe you can announce it before releasing the change, e.g. have it in the changelog as an "upcoming breaking change" for perhaps for a month before merging?

rpv-tomsk:
Maybe "configuration version" value is required in FlashVars for such cases, so code could apply "migration" after upgrade ?
This one is not the only PR which requires "breaking change", but others I saw too. (For example, PR related to "energy generation metering" which adds new fields into cfg structure). 

Even for old devices what is possible to detect "0xFF" value as "old configuration version" marker, apply "migration" and save 0x01 into version to mark it as "upgraded" too.

PR #1084, opened 2024-02-14
Add SK6812 support to SM16703P driver

zuckschwerdt:
This is adds code to support SK6812 strips which swap the red and green colors with respect to SM16703P.

The init command is enhanced to accept e.g. `SM16703P_Init 16 SK6812`. Allowed chip type strings are "SK6812", "sk6812", "GRB", "grb".

### DIFF
--- a/src/driver/drv_sm16703P.c
+++ b/src/driver/drv_sm16703P.c
@@ -28,6 +28,7 @@ UINT8 *send_buf;
 struct spi_message *spi_msg;
 BOOLEAN initialized = false;
 uint32_t pixel_count = 0;
+bool format_grb = false; // swap R/G for SK6812
 
 static uint8_t translate_2bit(uint8_t input) {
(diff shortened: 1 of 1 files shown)


openshwprojects:
Thank you, is it ready to merge, as it's no longer draft? Thanks!

I will try tomorrow to get my WS2812B testing setup up again. Last time I tested it it was okaay except the SPI quirks breaking st uff.

zuckschwerdt:
Yes ready. It is working "in production" here. But you need to review if that `SM16703P_Init` change (adding a string keyword) is in line with how things usually work in this project.

The SPI fix/workaround is in #1085, but needs a rebase.

openshwprojects (approved):

openshwprojects:
I made a little change:
image

stricmp is better and let's just make a single ketyword.

In future I may change that bool to enum, but it will work good as well.

zuckschwerdt:
> In future I may change that bool to enum, but it will work good as well.

My thoughts too, but I didn't want to change so much right now :)

PR #1085, opened 2024-02-14
Fix SM16703P SPI quirks

zuckschwerdt:
This is a workaround to the SPI sending in SM16703P to support SK6812 strips on BK7231N, e.g. the Action LSC Battletron Light Bar.

Discussion is in #1055.

### DIFF
--- a/src/driver/drv_sm16703P.c
+++ b/src/driver/drv_sm16703P.c
@@ -28,6 +28,8 @@ UINT8 *send_buf;
 struct spi_message *spi_msg;
 BOOLEAN initialized = false;
 uint32_t pixel_count = 0;
+uint32_t pixel_offset = 0;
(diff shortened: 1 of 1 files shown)


openshwprojects:
Good job, I've merged previous PR, I've just slightly changed the way that argument for RGB format is parsed. Please fix merge conflicts and I will merge this PR as well.
In the future, I am planning to change that boolean to something like enum so we can support multiple formats.

openshwprojects:
It seems fine, altough I think this part could be commented (as in explained what it does in the comment):
image

openshwprojects (approved):

zuckschwerdt:
> It seems fine, altough I think this part could be commented (as in explained what it does in the comment)

I'll PR some doc comments to this file soon.

PR #1086, opened 2024-02-14
Fix DDP off-by-one memory corruption

zuckschwerdt:
This fixes an off-by-one memory corruption on received UDP frames in the DDP driver.

Received `nbytes` can be up to `sizeof(msgbuf)` thus `msgbuf[nbytes] = '\0'` will reliably crash the device on all UDP messages of 64 bytes or more.

Since we don't ever need the DDP message to be terminated I've just removed the offending line.


openshwprojects:
Hmm nice find, it makes me wonder is there the saem vulnerability in the SSDP as well..

openshwprojects (approved):

Issue #805, opened 2023-04-24
HA MQTT discovery topic is not sending daily energy stats


fokcuk:
As per the header - HA MQTT discovery topic is not sending daily energy stats
I can see that its sending MQTT data for daily, hourly etc in the main device topics, but there is no configuration to show what is actually getting sent to the Homeassistant discovery.
When I checked MQTT data, OpenBK is sending hourly stats, but not daily stats. So HA never discovers it
Is there a way to check/change what is getting to sent via HA discovery?

openshwprojects:
Thanks. Json format was wrong. It should be ok now. Can you retest?

stefan064:
I've noticed this also. When NTP is enabled then `energycounter_yesterday`, `energycounter_today `and `energycounter_clear_date` are all published on MQTT as expected and also show up on the web UI. However none of these are discovered in home assistant, which is because command `scheduleHADiscovery` doesn't include them in the MQTT `homeassistant/sensor` topic.

I tried manually publishing one of them and that fixed it: i.e. after some ~half-hour home assistant discovered the sensor and started showing its value.

Possible cause:
In hass.c -> hass_init_power_sensor_device_info() it looks like `OBK_CONSUMPTION_YESTERDAY, OBK_CONSUMPTION_TODAY, OBK_CONSUMPTION_CLEAR_DATE` are never included and therefore left out. Perhaps these are meant to be conditionally included when NTP is enabled or similar?

stefan064:
Here's my attempt at improving this: https://github.com/stefan064/OpenBK7231T_App
- added energycounter_yesterday, energycounter_today, energycounter_clear_date to home assistant discovery
- added 'friendly names' to the energy sensors (i.e. "energycounter_today" ->  "Energy Today")
- adjusted formatting of energycounter_clear_date so home assistant can interpret as a 'timestamp'.
Testing on a few BK7231N devices now with BL0937 / BL0942 energy monitoring chips

![image](https://github.com/openshwprojects/OpenBK7231T_App/assets/40447901/e701c914-881a-485b-ab95-eb3cf1317c98)

PR #1092, opened 2024-02-18
Home Assistant discovery - add energy counters today/yesterday, add clear date, add friendly names

stefan064:
Hi openshwprojects,
I'd like to offer mycontribution for energy monitoring devices.
Reason for changes:
- #805 - Energy counters for today/yesterday and energy clear date are not being discovered by home assistant

Cause of reason:
- The missing entities are not included in home assistant mqtt discovery topics

Changes made:
- Added missing items to home assistant discovery generation
- Added friendly names to the home assistant entities. e.g. energycounter_today displays as 'Energy Today'
- Added 'Energy Clear Date (i.e. energycounter_clear_date)'
- Changed formatting of 'energycounter_clear_date' so home assistant can use it as a timestamp instead of a string.

Before:
Screenshot 2024-02-18 103845
After:
Screenshot 2024-02-18 112921
Thankyou for this great project.


openshwprojects:
Thank you, it seems like a great addition. Just to be sure, is this "combo name" here correct? When is it used?
image

TurkeyMan:
Hi there, while at it, can add reporting for apparentpower, reactivepower, and powerfactor?

stefan064:
"Voltage or Current or Power" is the fall-back home assistant entity name in case the name can't be retrieved from `sensor_hassNames[]` (previously `sensor_mqttNames[]`). I saw and changed this while debugging only after finding the previous string "Power" a bit confusing because a `VCP_SENSOR` can also be a voltage or current. I don't think this string is meant to be seen in normal usage.

openshwprojects:
Can you fix W800 so I can mergei t?

stefan064:
I'm a bit held up on OpenW800 failing the checks here;
```
CC  
./bin/build/w800/lib/libuser.a(hass.o): In function `$d':
(.text.hass_init_device_info+0x2a0): undefined reference to `sensor_hassNames'
```
Something linking related specific to that build? Would appreciate a second opinion if anyone has some clues.

stefan064:
Some progress with adding power factor, reactive power, apparent power here:
[](https://github.com/stefan064/OpenBK7231T_App/tree/refactor-energy-sensors)
Testing this so far. Keen on feedback about how much home assistant reconfiguring it may or may not impose on existing setups.
image

openshwprojects:
Very nice, it seems that all compile errors are fixed. Do you think it's ready for merge?

stefan064:
Yes please it's good to go, no issues in my testing so far. I'll make a separate PR for the other additions.

TurkeyMan:
That's super awesome, thanks for doing this!

There's one odd thing to note in your screenshot though; not sure if a bug or a hardware thing... did you notice that Apparent power is ~200VA, and Power is ~260W... how did we get there?
Power factor says 0.67. Something's wrong...?

TurkeyMan:
It says above "This PR is included in version 1.17.478", I just updated to test this out, but it doesn't seem to be reporting any of those additional energy values. I'm still only getting reports for `voltage`, `power`, `current`.
Is it behind a new option or anything?

TurkeyMan:
Also I was thinking on the report options `VCPPrecision` `VCPPublishThreshold` for those values.
I reckon apparent power and reactive power can follow the same precision and threshold as power, but powerfactor might need values specified from those 2 options.

stefan064:
Home assistant discovery may need retriggering to make the new values appear, e.g. via web ui `Config -> home assistant configuration -> start home assistant discovery`, or via command scheduleHADiscovery. Not sure why thats been necessary but may relate to how home assistant interprets the order of things. I also saw that
- deleting the device from HA also deletes its discovery topics from mqtt. 
- manally deleting retained discovery topics causes HA to remove their sensors automatically
- Deleting from HA then re-adding via 'scheduleHADiscovery' brings the device back including entity history when entity ids match. HA generates entity id from device name and sensor names.
Thankyou @TurkeyMan for spotting that, they probably were wrong too as I later learnt about how hass `uniq_id` works...
I'll share further progress via your #1097

TurkeyMan:
When I click "start home assistant discovery", it just shows a message "MQTT not running".
What is this actually supposed to do?
I'm not using homeassistant... what is the goal here, or what mechanism will make it start reporting these new topics?

PR #1098, opened 2024-02-21
Warnings in OpenBK7231T_App fixed.

masj-mobica:
Compiler warnings for OpenBK7231T_App build are fixed in that pull request.


openshwprojects:
Not bad, but why is that header there:
image

masj-mobica:
> Not bad, but why is that header there: image

It looks that it was added by mistake. Removed in the latest commit.

openshwprojects:
Hmm 
image
it should be buffer size, but it seems like a mistake on my part. Duh.

PR #1099, opened 2024-02-23
Updates to drv_ir.h and drv_ir.cpp

Error-Gap:
* Migrate headers for SpoofIrSender to drv_ir.h (avoid issues between SpoofIrSender and IRSend object instantiation)
* Update code for SpoofIrSender to incorporate a pointer to myIRsend object which can be attached to SpoofIrSender instance
* Push functions through currently attached myIRsend instance for previously blank functions within SpoofIrSender


Error-Gap:
This should fix issues with IRSend with remotes that land on the "sendPulseDistanceWidthFromArray" function of SpoofIrSender (such as Kaseikyo based remotes)

openshwprojects (approved):

Issue #1097, opened 2024-02-18
Please report missing energy signals (apparentpower, reactivepower, powerfactor)


TurkeyMan:
**Describe the bug**
Device reports `voltage`, `power`, `current`, but it doesn't report `apparentpower`, `reactivepower`, `powerfactor`.
Please add reporting for those values.

If you think it's not something that most people would want, then put it behind a configuration flag?

**Firmware:**
- Version 1.17.464
- Chip/model: BK7231N
- Device: relay + energy meter
- Relay on channel 0 + BL0942

openshwprojects:
https://github.com/openshwprojects/OpenBK7231T_App/pull/1092

TurkeyMan:
Appears to be merged so I was going to close this issue, but I tested and it wasn't reporting.
Maybe there's a new option I didn't see or something like that?

openshwprojects:
Try to redo hass Discovery

stefan064:
More progress:
https://github.com/stefan064/OpenBK7231T_App/tree/refactor-energy-sensors
![image](https://github.com/openshwprojects/OpenBK7231T_App/assets/40447901/d787cb19-6452-438f-a8e9-a9b053d0bac8)
![image](https://github.com/openshwprojects/OpenBK7231T_App/assets/40447901/42b9c62b-fe35-4f4f-a461-024d07a5e5bf)
I found and included 2 additional past days energy were already being recorded. And some diagnostic sensors (via flag 2 'broadcast self state..')
I may have gotten carried away refactoring particularly in `drv_bl_shared.c`, and feel the need to give this a fairly thorough test. Keen on feedback too.

TurkeyMan:
Repost here:

When I click "start home assistant discovery", it just shows a message "MQTT not running".
What is this actually supposed to do?
I'm not using homeassistant... what is the goal here, or what mechanism will make it start reporting these new topics?

openshwprojects:
Well,you obviously need to configure mqtt first in order to use home
assistant discovery

TurkeyMan:
I'm not sure what that means though? What is home assistant discovery?
What is it to "configure MQTT'? MQTT is working fine for me, it connects to my broker and reports on the topics that it should be at the rates I expect...?

stefan064:
'Start home assistant discovery' causes the device to publish a group of messages under the mqqt topic `/homeassitant`. That's all, nothing more.
[...]

PR #1102, opened 2024-02-24
Add energy sensors to MQTT + HA discovery for active power, reactive power, power factor

stefan064:
For #1097 
And 'energy 2 days ago', 'energy 3 days ago'.
Also added discovery for diagnostic RSSI and uptime as already published on mqtt.
I'm still testing this carefully after refactoring in `drv_bl_shared.c` and don't consider it ready as yet, but keen on feedback and help with testing.


openshwprojects:
Not bad, I think I will be happy to merge it, but it needs to be tested well. Maybe I should extend the self test system. it is important not to break the other features...

stefan064:
Thank you, appreciate your concern. I've fixed a few minor things but otherwise looking good so far.

openshwprojects:
Do you think it's safe to merge?

stefan064:
Yes please I feel it's safe for all to try now.
Positive notes:
- MQTT publish thresholds and intervals appear to be respecting their settings.
- 'periodic statistics' works as before.
- Units and Wh to kWh conversions look correct, as with rounding etc.
- Changes appear to go smoothly with existing home assistant device and entity configs.

Other observations:
- Not sure how useful '2 days ago' and '3 days ago' will be for home assistant users; they came as a convenient side effect of my changes.
- I've learnt that 'Energy last hour' isn't really 'last hour' but moreso 'over the sum of periodics statistics'. I left the name as before but just noting here and in some code comments.
- I left the web UI Wh/kWh units as they were instead of making them follow mqtt settings, in case there was some reason for it being this way.

As seen in home assistant, before :
image
And After:
image
Web UI:
image
Thankyou for considering all this.

openshwprojects (approved):

PR #1107, opened 2024-02-27
better mqtt build version

openshwprojects:


src/mqtt/new_mqtt.c
```diff
@@ -12,6 +12,8 @@
 #include "../driver/drv_tuyaMCU.h"
 #include "../ota/ota.h"
 
+#define BUILD_AND_VERSION_FOR_MQTT PLATFORM_MCU_NAME " " __DATE__ " " __TIME__ " v " USER_SW_VER
```
TurkeyMan, line 15: #define BUILD_AND_VERSION_FOR_MQTT "OpenBeken v" USER_SW_VER " hw: " PLATFORM_MCU_NAME " mf: " __MANUFACTURER_STRING__ " build: " __DATE__ " " __TIME__

Something like this maybe? Still not perfect; I'm trying to order it both logically, but also where data is likely to be parsed, and I reckon the date/time is the least likely thing anyone will want to parse out.

openshwprojects:
if you want more info, then maybe send cmnd status packet and parse json?

TurkeyMan:
You mean the `tele/obk125DAED5/SENSOR` topic?

Yeah, maybe that's the way forward. Are those the raw unfiltered values? I love that it's timestamped; is that the real-time that the data was sampled from the hardware? That would be perfect if it is.
Can we add milliseconds to the time? And also `frequency` is missing from `ENERGY { ... }`.

openshwprojects:
image

TurkeyMan:
What am I looking at?

openshwprojects:
you can assign status command via http or mqtt and get reply in a form of json...

Issue #1002, opened 2023-12-18
Add 'Home' link at bottom of second level menu screens?


divadiow:
Would it be possible to put a 'Home' link at the furthest left of the bottom link section, before 'Return to cfg' on the secondary level on config pages? 

It's slightly annoying going back two steps to get to the home page of a device.

For example when on the module config, mqtt or WiFi config pages, you either have to click back twice in the browser or click return to cfg then return to menu. To make consistent, maybe 'return to menu' should be 'Home' too, as you're already on a menu page of sorts. A little clarifying perhaps?

Also, change 'Build on xxxxx' date to 'Buil**t** on xxxxx'.

Just a couple of thoughts! 

Thank you for all you do on this, it's awesome.
![IMG_20231218_054619](https://github.com/openshwprojects/OpenBK7231T_App/assets/62958974/b546ff16-c5a2-4ccc-a56c-a6781dc1d6a4)

PR #1115, opened 2024-03-07
MAIN page item added to the HTML footer.

masj-mobica:
Based on the suggestion from the issue #1002 I've added 'MAIN page' to the HTML footer. It's small change but in fact makes navigation through WebGUI more convenient.


divadiow:
excellent!!

openshwprojects (approved):

openshwprojects:
Nice idea, thank you

PR #1122, opened 2024-03-10
Hass discovery

stefan064:
Hi Openshwprojects,
I'd like to offer these improvements for home assistant discovery;
- Channels that have a name set via SetChannelLabel will also have their HA entities named the same 
- Channels excluded from mqtt via SetChannelPrivate are also excluded from HA discovery messages
- Firmware build info added to discovery under diagnostic category, using already published 'build' topic
- Energy sensor discovery bug fix re clear-date

Thank you


openshwprojects:
It seems okay, but are the changes documented anywhere? Did you also update the comments of the commands for setting channel private, etc?

stefan064:
Sure I've updated the docs for those commands, is that what you had in mind?
Example of 'build info' in HA discovery:
image
I'm afraid I didn't take before/after screenshots this time, would that help here?
It's been quite handy for tuyamcu-based devices that occupy multiple channels.

openshwprojects:
I was referring to changing the command descriptions in the docs in the code, but hmm ok let's give this PR a chance, I will look into docs later.

openshwprojects (approved):

PR #1141, opened 2024-03-24
Quick and dirty workaround for gmtime issues with LN882H

MaxineMuster:
Just replaced all occurrences of

`ltm = gmtime((time_t*)&g_ntpTime);`

with

```
time_t mytime=(time_t)g_ntpTime;
ltm = gmtime(&mytime);
```


openshwprojects:
Thank you for finding that out, here is correct soltuion:
https://github.com/openshwprojects/OpenBK7231T_App/pull/1146

PR #1150, opened 2024-03-26
Some more fixes to usage of gmtime

MaxineMuster:
After recent changes to fix usage of time_t, there were 4 other places, which should be addressed:

```
max@max:~/LN882H/20240326/OpenBK7231T_App$ find . -name "*.[ch]" | xargs grep -C 2 -n  "gmtime((time_t"
./src/driver/drv_max72xx_clock.c-71-
./src/driver/drv_max72xx_clock.c-72-	// NOTE: on windows, you need _USE_32BIT_TIME_T 
./src/driver/drv_max72xx_clock.c:73:	ltm = gmtime((time_t*)&g_ntpTime);
./src/driver/drv_max72xx_clock.c-74-
./src/driver/drv_max72xx_clock.c-75-	if (ltm == 0) {
--
./src/driver/drv_ntp_events.c-172-
./src/driver/drv_ntp_events.c-173-	// NOTE: on windows, you need _USE_32BIT_TIME_T 
./src/driver/drv_ntp_events.c:174:	ltm = gmtime((time_t*)&runTime);
./src/driver/drv_ntp_events.c-175-	
./src/driver/drv_ntp_events.c-176-	if (ltm == 0) {
--
./src/driver/drv_ntp_events.c-313-	uint8_t hour_b, minute_b;
./src/driver/drv_ntp_events.c-314-	int sunflags = 0;
./src/driver/drv_ntp_events.c:315:	struct tm *ltm = gmtime((time_t*) &ntp_eventsTime);
./src/driver/drv_ntp_events.c-316-#endif
./src/driver/drv_ntp_events.c-317-
--
./src/httpserver/json_interface.c-655-	struct tm* ltm;
./src/httpserver/json_interface.c-656-	int ntpTime = NTP_GetCurrentTime() - g_secondsElapsed;
./src/httpserver/json_interface.c:657:	ltm = gmtime((time_t*)&ntpTime);
./src/httpserver/json_interface.c-658-
./src/httpserver/json_interface.c-659-	if (ltm != 0) {
max@max:~/LN882H/20240326/OpenBK7231T_App$ 

```

The first one (in /src/driver/drv_max72xx_clock.c) is just a forgotten removal of the (now obsolete) typecast.

The second and third in _src/driver/drv_ntp_events.c_ will need some consideration:
"runtime" is a parameter ("void NTP_RunEventsForSecond(unsigned int runTime) {") so it would need to alter the function call to use "time_t runTime" or we can just use the working typecast to a time_t var here (like in the "quick and dirty solution)

An alternate "fix" could be to define "ntp_eventsTime" as time_t, too, this will resolve both issues (since "runTime" is called with "ntp_eventsTime", so we can change both to "time_t". On the other hand, all calculations with times are based on "unsigned int", so they would need attention, to.
So I used the first approach, to cast "unsigned in" to time_t, which is of same or even larger type, so casting shold be o.k.

For the third occurrence I propose something like my "solution" from issue again:
	adding a test, if NTP is synced - this avoids calling it when NTP is not synced (resulting in a negative "ntpTime")
	define ntpTime as time_t her, too (doing the casting inside the calculation
	and we can fix using the local time instead of UTC time at the same time ;-)

### DIFF
--- a/src/driver/drv_max72xx_clock.c
+++ b/src/driver/drv_max72xx_clock.c
@@ -70,7 +70,7 @@ void Clock_Send(int type) {
 	char *p;
 
 	// NOTE: on windows, you need _USE_32BIT_TIME_T 
-	ltm = gmtime((time_t*)&g_ntpTime);
+	ltm = gmtime(&g_ntpTime);
 
 	if (ltm == 0) {
 		return;
--- a/src/driver/drv_ntp_events.c
+++ b/src/driver/drv_ntp_events.c
@@ -16,6 +16,8 @@
 #define LOG_FEATURE LOG_FEATURE_NTP
 
 unsigned int ntp_eventsTime = 0;
+// to use time.h functions, we can't rely on time_t being unsigned int, so we need conversion to time_t in some cases
+static time_t tmptime;
 
 typedef struct ntpEvent_s {
 	byte hour;
@@ -171,7 +173,9 @@ void NTP_RunEventsForSecond(unsigned int runTime) {
 	struct tm *ltm;
 
 	// NOTE: on windows, you need _USE_32BIT_TIME_T 
-	ltm = gmtime((time_t*)&runTime);
+	// make sure to call gmtime with proper pointer to a time_t value
+	tmptime = (time_t)runTime;
+	ltm = gmtime(&tmptime);
 	
 	if (ltm == 0) {
 		return;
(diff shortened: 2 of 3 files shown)


openshwprojects:
Nice, but I am not sure about if for StartupUTC - what happens with Tasmota control if that field is missing from JSON?
Maybe change if to work in such a way that if value negative would be to happen, then plug 0 so it prints 1970-01-01 etc etc? So just a clamp, but still add JSON key.

openshwprojects (approved):

PR #1152, opened 2024-03-26
First try for time keeping without NTP

MaxineMuster:
Since often you might not need very exact time, or even can't connect to an NTP server, I tried an different approach for some "easy" time keeping:

If you store the epoch at startup of the device, you can get the actual time by adding "g_secondsElapsed".
Since this is quite unreliable (drifting off 1 to 2 minutes per day) I introduced a function to set it to a more reliable value derived from xTaskGetTickCount() - I hope it is present on all platforms.
ATM its only tested on the one platform I have: LN882H.

Up to now its more like an "extended proof of concept" - probably some bugs inside and even during upload I saw quite some typos ..
"Extended" because it will also respect daylight saving time - for the EU, but this can be extended.
I would be happy to get some feedback, there are quite some aspects to discuss:
E.g. how to implement DST: I have two possible solutions tested: 

1. A formula to calculate it - not really hard to implement - but how to work with different DST rules?
2. A table with pre-calculated times where switches in DST will take place 

The second solution is implemented at the moment: Takes some memory and is restricted to the calculated dates.
But it should be quite easy to change/extend it for other zones, if you only have to add dates, but no formula.
And regarding the restricted list - maybe its possible to realize an "update" process 

Now I'm at the point to find out, if this is something worth to invest more time:
 It won't make sense to put too much afford in this, if it will never make it into the project.


openshwprojects:
That's a very interesting concept, I want it merged, but yould you be able to add a #define to enable and disable that?
I am asking becaues BL602 is very limited on space and we need to be able to disable that for this platform.

MaxineMuster:
Thanks for your feedback.

Sure, this would need some effords to make it flexible and configurable.
E.g. just "emulate" a clock, no additional features, this would take only few bytes to store the time and might be included even in the "core".
Handling of timezones, DST and so an could be optional.
And maybe we should devide the actual NTP driver into parts related to NTP and more general functions:

- sunset and sunrise
- time related events 
- ...
this is not necessarily bound to NTP but just to having a running clock

MaxineMuster:
Can I somehow include changes to the Makefiles in sdk? 
In order to compile, the added file must be included, which I can do on my local clone, but theese changes are not included.
So I would need a changed version of the sdk-files, too...

openshwprojects:
I can see you made a lot of updates. Can you change the function names to fit our standard, if not NTP_GetHour then maybe Clock_GetHour, etc? Windows (uised on simulator) has names like GetMiliseconds so we need to use prefixes.

The SDK changes are problematic because they need to be done separately, but on the other hand, I can't do the in main tree because the new file created would be missing in main... maybe I can do them manually once you get the basic version working?
Is this version working? Could it have been merged soon?

Is there a correct #define to enable/disable manual timekeeping in obk_config.h ?

MaxineMuster:
No longer needed

Issue #1116, opened 2024-03-07
Unable to set static ip address


Magnus-rosenborg:
I have changed 0.0.0.0 to 192.168.200.239 but it's still using the old dhcp address.
The device have been swithed off but still the same problem.
Flags: 1120

**Firmware:**
- Version 1.17.494
- Device? mini smart switch
- Chip/model: BL602
- Device config? nothing. 

**To Reproduce**

**Screenshots**
![IMG_4263](https://github.com/openshwprojects/OpenBK7231T_App/assets/49500329/fddacca9-1482-478b-9621-346bc6909789)
![IMG_4264](https://github.com/openshwprojects/OpenBK7231T_App/assets/49500329/1a3c3d1a-688e-4baf-ab4d-e372b64e77c4)
![browser](https://github.com/openshwprojects/OpenBK7231T_App/assets/49500329/7d10611a-e55f-4a72-803a-d46728cd368c)

**Additional context**

Magnus-rosenborg:
Se also my reported issue about lfs not working on the same device.

openshwprojects:
This has been solved with PR #1155 . I am closing this issue now, but let me know if there is some other issues with static IP on BL602 platform.

PR #1155, opened 2024-03-28
Add static ip for bl602

giedriuslt:
This adds ability to use static ip for BL602 devices as requested in #1116


openshwprojects:
Very nice! Was it tested in both scenarios - static and DHCP?

giedriuslt:
Yes, works both ways

openshwprojects (approved):

PR #1163, opened 2024-04-02
Add color format option for sm16703p

geekuillaume:
Current logic for the SM16703P driver allows switching the R and G channel but doesn't allow the other combinations. This PR adds the 4 other possible combinations of R,G and B. It also updates the documentation to add this option for the command.

I made this change because my LED strip needs to use the GBR color order and this wasn't possible with the current code.


openshwprojects:
Thanks but here is one little thing to fix:
image
This part is autogenerated from:
image
Can you update the source of the docs (in C code) as well?

openshwprojects:
Ok I think I will merge it and update docs myself

openshwprojects (approved):

geekuillaume:
Thanks! Sorry, I don't have a lot of time currently and couldn't do these changes yesterday. Happy to see this change merged! Thanks for all the good work on this project

openshwprojects:
Thx, that'\s how docs should be done: https://github.com/openshwprojects/OpenBK7231T_App/commit/5d2c01045dff520418029216cfd217470233cd75

PR #1167, opened 2024-04-04
Timekeeping on OpenBeken without NTP

MaxineMuster:
I tried to add as many #if s as possible, though impact is not too much (for OpenLN882H just the clock, without handling of daylight saving time, its less than 500 bytes more compared to "OpenLN882H_1.17.534.bin").
The main part results from the Javascript extensions, which I can't "comment out" without making a separate section outside the "gulp-ed" part in "new_http.c".

I would be pleased to get some feedback if it works on other platforms!


MaxineMuster:
No longer needed

PR #1170, opened 2024-04-06
Fix RSSI for W600

giedriuslt:
Looks like rrsi is inverted for W600, otherwise looks similar to beken in similar places


giedriuslt:
NOt sure now, according to example it shuld be different, but tat doesnt work

PR #1173, opened 2024-04-07
sscanf into temp int array, it seems w600 does not support %hhu, and …

giedriuslt:
…assumes %u
This fixes issue what dns value was overwritten when submitting ip settings, reported in forum


openshwprojects:
Nice catch. we must check also for other places in code wtih this issue..

openshwprojects (approved):

giedriuslt:
There is another place I saw with same %hhu specifier in some rgb related code, currently not enabled for W600. 
It's using very old gcc tools from 2014, i think that might be the culprit here. I wonder if it's possible to update easily 🤔

PR #1175, opened 2024-04-10
fix network data log

giedriuslt:
This fixes incorrect printing on network settings for W600. 
Caused by incorrect call to wm_vsnprintf. It can't be called as normal function. va_args need to setup and so on. So using regular snprintf instead


openshwprojects (approved):

openshwprojects:
Thank you for your continued support!

Issue #85, opened 2022-05-17
MQTT connection problem


iprak:
I recently flashed a couple BK7231T WB2S devices (same devices) with the latest OpenBK7231T_UA_1.0.20.bin image. I had a hard time getting BKwriter to work but was able to make it work in the first attempt using hid_download_py.

The problem I am running into is with MQTT connection. I ahve noticed that if I leave user/password empty then MQTT is 0 in the log but if I enter something random e.g. user/password then connection succeeds. My MQTT server does not require authentication. Any suggestions for further debugging?

openshwprojects:
This is something new, I have never tested it with MQTT server with no auth required.

Can you check how the send packets looks like in Wireshark?

The original MQTT library did not support auth, but I added a support for that:
https://github.com/openshwprojects/OpenBK7231N/commit/172f1012c3a03b73441abc24a9f8ddd11e2923f1
that was required for both N and T platforms, but right now I can't see anything wrong in that code.

alex3594:
FYI, I checked that with tcpdump and can confirm when no user and password set, device is not sending any packets to the MQTT.

openshwprojects:
should be fixed
https://github.com/openshwprojects/OpenBK7231T/commit/2c3af1f719cb5d52f3a9442bc0ec4439cac1d389

PR #1176, opened 2024-04-11
Fix for Issue 808: Door/Window sensor lost data after deep sleep

bvelush:
**DoorWindowSensor driver has a security problem**, which is registered as an issue 808: if the open/close cycle happens fully before the device has been connected to MQTT broker, the event is lost. So, if an intruder opens/closes the door within about 15 seconds, this will not be registered.

The fix has been tested on 7 of Door Sensors, and it fully fixes the issue. It had been tested also with MQTT broker being down for 300 seconds to test for the queue overflow, and the memory limit is still very far.

**Proposed change leverages the MQTT queue API:**

- MQTT_QueuePublish() to queue the event while the device is still connecting to WiFi and/or MQTT,
- PublishQueuedItems() to dump all queued events when it has connected to MQTT Broker
The rest of the behavior is left intact: the device will publish events directly (without queue) every second when it's connected, until it goes to sleep again.

A few things about the queue API itself:
I don't understand why PublishQueuedItems() calls the MQTT_PublishTopicToClient() with appendGet parameter being set to false, effectively forbidding queued messages to be published to the same channel as those that are published directly. I think this is a bug, that requires a fix. For now, to work around this behavior, I am manually adding the suffix "/get" when queueing the event.

### DIFF
--- a/src/driver/drv_doorSensorWithDeepSleep.c
+++ b/src/driver/drv_doorSensorWithDeepSleep.c
@@ -104,7 +104,7 @@ void DoorDeepSleep_OnEverySecond() {
 					sprintf(sChannel, "%i/get", channel); // manually adding the suffix "/get" to the topic
 					// Explanation: I manually add "/get" suffix to the sChannel, because for some reason, 
 					// when queued messages are published through PuublishQueuedItems(), the  
-					// functionality of appendding /get is disabled (inMQTT_PublishTopicToClient()), 
+					// functionality of appendding /get is disabled (in MQTT_PublishTopicToClient()), 
 					// and there is no flag to enforce it. 
 					// There is only a flag (OBK_PUBLISH_FLAG_FORCE_REMOVE_GET) to remove 
 					// suffix, but for some reason there is no flag to add it. 
--- a/src/driver/drv_doorSensorWithDeepSleep.c
+++ b/src/driver/drv_doorSensorWithDeepSleep.c
@@ -59,8 +59,9 @@ void DoorDeepSleep_Init() {
 }
 
 void DoorDeepSleep_OnEverySecond() {
-	int i, bValue;
-	char tmp[8];
+	int i;
+	char sChannel[8]; // channel as a string
+	char sValue[8];   // channel value as a string
 
 #if PLATFORM_BK7231N || PLATFORM_BK7231T
 	if (ota_progress() >= 0) {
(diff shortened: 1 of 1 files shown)


openshwprojects:
I think this may be a good solution. I'll give it a chance.

openshwprojects:
Regarding appendGet, please open another issue and we will investigate. Thanks!

openshwprojects:
On a second tought, maybe we can improve it a bit futher. Maybe before calling MQTT_QueuePublish we can check if the value is already in queue? Because current version will add a new entry to queue every second before MQTT is on, won;t it?

bvelush:
Your suggestion to send to a queue only changes, not the current state at every second, is good. I was thinking the same. The reason I didn't implement it, was that I didn't want to change the existing behavior, where in existing code the state is published every second no matter whether it had changed or not. I think that if we want to publish changes only, we should do it in both cases, when connection has already been established (former lines #85-90) and when events are being queued. Since the code is going to be pretty much the same, I think moving it to a separate function would be great, except that in one case it's being queued, in another -- published directly to the MQTT. Is there any particular reason for it to _not_ being queued when connection is established? If not, than the code could be simplified by simply putting it in the queue in both cases. I think I can do these changes in next few days, if you don't mind. Thank you for your support of this change.

openshwprojects:
I don't have any active door sensors so take my advice with a grain of salt, I may be wrong, Lots of strangeness of the code comes from the fact that MQTT queue was added a bit later and the queue is limited to (as far as I remember) 16 or so entries, so when not needed, we just do publish directly.

Futhermore, we can just publish directly every second without getting into any potential problems and without affecting other parts of the system, while we can't just add to queue blindly, because it has limited size ot (as far as I remember) 16 and it could get clogged very easily

My suggestion for the duplicates fix would be something like adding a boolean argument to MQTT_QueuePublish , argument bool bSkipDuplicate, which will skip adding item if the last item in the queue with the same topic has the same value....

I know that lot's of stuff could be improved here, but things like door sensors  are very hard for me (as a single person) to test efficiently, that's why I'm trying to be on the safe side with the publishes.

I am not sure about using queue in all cases, but I suspect that using a queue may introduce a very little (one "every second" tick?) extra delay in comparison to direct publish call.

bvelush:
I really like your idea about de-duplication. But there is a tax to pay, even two:
1. The queue is common for all topics. What I mean is that a queue node is a structure with the topic and the payload, so if we want to dedup this, we need a separate dictionary/map with {topic: last_msg}. This could be an expensive for memory if last_msg is long in any of topics. Given that there could be many topics, it could be really expensive. 
2. Comparing each incoming message with the last message for that topic could be CPU-intensive. 
We can mitigate both taxes by storing only, say, first 63 chars (or so) and comparing to that, but it will be hard to document and prone to mis-use in general. From this perspective, I'd suggest that deduplication logic should be local to the DoorSensor, because here we have only 1 topic and the message is just one byte. 

If you decide we can pay this price, I'll implement it. 
instead of changing the interface of MQTT_QueuePublish (and potentially braking some other code that relies on it), I would propose to wrap MQTT_QueuePublish into some MQTT_QueuePublish_Dedup() {_takes same args as MQTT_QueuePublish but de-duplicates_} and promote it in dev guide? This way DoorSensor will use this new _Dedup function, and other code can gradually move to this new functionality later with some tests? 

For now, I'll implement deduplication logic locally and wait for your decision.

**About 16 msg in the queue**:
The internal mqtt message queue is implemented as a linked list without size limit (may be this is the source of the problems you have mentioned? Having this on top of adding to the queue every second when MQTT is not available could explain it :) ). I was surprised. To manage memory more efficiently, it has a flag on every node in the linked list to re-use (!!) this node if the message was published, so when the memory for the node is allocated, it stays to be reused later. 
I can attest that the queue holds at least 500 messages and then publishing them all to mqtt server in a correct order of incoming. I have tested it and can share the methodology separately, if you are interested. Of course, with dedup logic (either local or global) this event flood will be limited to changes only. 
**There is one more behavior** with the queue that you have to be aware of. I can give more details if you are interested, but in short it's that we lose the information about the time interval of when door was opened <-> closed, when we are not connected to mqtt, and also lose the exact time when it happened. This is because the queue only has events of open/close, but not the info about when it happened. So when device connects to WiFi->MQTT, all these events are dumped (in order, of course) but in a short amount of time. For my 500 messages test above, it was all these 500 events dump in ~5 seconds to Home Assistant's MQTT, and they are registered in Home Assistant all in that time, not when the door was actually opened/closed. It's something we can't change unless we travel back in time, so it might be obvious to you but I mention it anyway.

**About the 1 sec delay**:
You are right about the delay with the way the code is written now, but we can MQTT_QueuePublish_Dedup _first_ and then in the same tick of time PublishQueuedItems to the server. This way the delay will not be introduced.

In general, I am not insisting on the queueing the messages when device is mqtt-connected, we definitely can publish directly, but my tests show that at least for DoorSensor of this particular brand (board name is MC4000-A0002) the queue is very reliable (I have 8 such devices and flashed them all with the latest code).

bvelush:
Final fix for Issue 808, tested on 8 door sensors

openshwprojects:
Looks good for me. Let's see if it compiles and I will merge

openshwprojects (approved):

PR #1180, opened 2024-04-14
Fix index page performance for LN882H

giedriuslt:
Index page takes 2 seconds to refresh LN882H, due to rssi being called twice.
image
Calling it once makes performance normal for some reason

image

### DIFF
--- a/src/httpserver/http_fns.c
+++ b/src/httpserver/http_fns.c
@@ -771,7 +771,8 @@ int http_fn_index(http_request_t* request) {
 	}
 	if (Main_HasWiFiConnected())
 	{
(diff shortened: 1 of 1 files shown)


divadiow:
Ah! Interesting! Did think something was odd with home refresh. Thank you. Will try asap

openshwprojects:
Nice catch, maybe in future we should somehow cach.e... and it also may be worth to check out how is that RSSI call handled internally by LN882H, why is it slow in the fist place

openshwprojects (approved):

giedriuslt:
Internally it goes to wifi driver which has no source, so not really possible to check. Cache is now not needed it's fast enough

Issue #1156, opened 2024-03-30
Separate measuremente for negative power values


brunohoff:
Hello, 

There is the Flag 25 that enables negative values on power measurement and it works perfectly reporting current and power with positive and negative values. In my case I use a circuit break with power meter on the house and have solar panels generating energy. When the system is generating more power than the house usage, the power measurement is negative and reduces the total energy over time as expected and the opposite happens when is positive. 

The Energy Total I believe there is room for improvement because it is a more precise value than multiple snapshots from power usage over time to calculate KWh usage with integral or other methods. The suggestion is to create a separate measurement for the total amount of negative values for the day.

For example, the house consumed 10 KWh on the day and generated 9 KWh with solar panels. The OpenBK will correctly inform the energy for the day at 1KWh. For Home Assistant there is the option to inform the amount of energy generated with solar, the energy consumed by the house from the grid and the energy sent to the grid.

In this case it could be something like:
Energy Today: 1KWh
Negative Energy Today: 9KWh

Therefore it will be possible to calculate the energy of all the system.

Thank you in advance.

2Fblob:
I've been working on this for a few weeks now. See the thread:
https://www.elektroda.com/rtvforum/topic4045136.html

brunohoff:
Hello @2Fblob, that`s exactly what I am looking for. Do you need any help with the development? I can help you with the merging and pull request.

2Fblob:
Hi @brunohoff ,
You may want to try my fork. https://github.com/2Fblob/OpenBK7231T_App. The code is in a functional state. 

I'm having some problem with the WiFi routines - It will either recover itself and send the measurements on the next reporting period or it will restart. When it restarts it doesn't always connect to the router, but restaring the wifi on the router solves it. I have asked in the foruns and got no feedback if OB has any issues with the WiFi. So it would be good to have someone else testing :-)

I'm setting up another device on a different location to also run some tests.
I can make a build for you with some aditional parts of the code removed (so it only reports power measurements). 
Please ping me on the forums.

brunohoff:
Hi @2Fblob,

My only device is connected to my house main circuit breaker. I need a more stable version to flash into it. I purchased a second unit to test in a separate environment.

Did you tried without those hardcoded http request?
```
`CMD_ExecuteCommand("SendGet http://192.168.8.164/cm?cmnd=Power%20on", 0);
CMD_ExecuteCommand("setChannel 1 1", 0); 
```

[...]

PR #1184, opened 2024-04-17
Added Generation Energy Support

brunohoff:
This PR is based on the work developed by @2Fblob. -> https://github.com/openshwprojects/OpenBK7231T_App/issues/1156

Adding support to register generation energy when the Negative Power is enabled. 

 I'm currently testing on my home, and opening the PR for more people to help.


openshwprojects:
Looks good, only one thing to fix is:
image
can you just store it in the last 2 short retain channels? Just as I said on forum....  I must find the topic, I think it was there: https://www.elektroda.com/rtvforum/topic4045136.html

brunohoff:
Thank you @openshwprojects, I roll back to the original 12.

I'm debugging and facing an issue that was solved by  @2Fblob implementation. His solution was to invert the energy capture to negative when the power on that second is negative. The problem with this implementation is that can lead to large measurement errors. He also tried to identify the flow of the current but apparently, it's not working properly: https://github.com/2Fblob/OpenBK7231T_App/blob/afbbf52fbf02ee9117984b42fb747b1aca2d6663/src/driver/drv_bl0942.c#L149


Maybe you can help with this, The i_rms properly give negative values which makes negative Power correctly https://github.com/openshwprojects/OpenBK7231T_App/blob/main/src/driver/drv_bl0942.c#L130

However, the cf_cnt always returns a positive value which always makes positive values on the energy variable. https://github.com/2Fblob/OpenBK7231T_App/blob/afbbf52fbf02ee9117984b42fb747b1aca2d6663/src/driver/drv_bl0942.c#L145

Do you have any ideas on how to make this measurement more accurate? A simple solution would be to invert the value if the i_rms is negative, but it would make incorrect measurements in my understanding.

2Fblob:
The instantaneous power returns a positive or negative value, which I am using.

I did try, on early versions of the code, to use the actual flag that monitors the direction of the last energy retrieved. What I found was that on phase controlled or half wave rectified loads we were getting mixed results and the calculations werent correct, because we were both consuming and exporting within one mains cycle. But I did notice that the instantaneous power, because its averaged over a period, returned the correct sign. I didn’t note any significant errors - Do you have any data showing measurement errors?

I do have a different problem as some devices are wired backwards, so they report power as positive or negative for a given direction. We need a flag to invert this, so the output is consistent. At the moment, my code assumes positive power = consumption, negative power = generation. That is to keep compatibility with the statistics feature, where consumption is measured (original code discards negative values, so we want to keep that unchanged).

On the retain channels I havent found an easy way to save and retrieve without making significant changes to the code, because loops are used to read those memory locations, so its important they keep the memory locations unchanged. Ill write about that in more detail once I'm done with the bug changes.

brunohoff:
@2Fblob please, don't get me wrong, I'm not an expert. In my understanding, the power and energy are registered every second and the instant power can change from positive to negative during the snapshot of the measurement, and the energy be accounted to the wrong side at that moment. Does it make sense?

2Fblob:
You're not wrong, but unfortunatelly the only way to get super accurate results is to measure at each half cycle, which is not really practical. Altough as I mentioned I didnt notice any significant error in real life, because the watt register is averaged over a period of time (It was very accurate with half wave rectified loads, for example).

Here's some detail from the datasheet.

Screenshot_20240417_222817_Drive

2Fblob:
I think I recall its averaged over 800mS on the code.
Screenshot_20240417_223812_Drive

brunohoff:
Thank you for the clarification @2Fblob. In that case, the code is working stable on my device. No crashes or disconnections. Please, feel free to add or change anything to improve the code.

image

I will only make one test inverting the order on the ENERGY_METERING_DATA to see if will break history compatibility.

brunohoff:
@openshwprojects, I changed the order of the elements on the ENERGY_METERING_DATA putting TotalGeneration by last. In my testing, the TotalConsumption was preserved on the OTA update in this case. 

@2Fblob, I kept running for the entire day and indeed didn't see much difference from the other measurements like you anticipated, thank you.

rpv-tomsk:
Hi,
Can you please take a look into https://github.com/openshwprojects/OpenBK7231T_App/pull/1273 ?

I propose several structure changes in power-measurement modules, you might be interested in.
Thanks.

brunohoff:
Hello @rpv-tomsk,

The idea seems very good, congrats on the initiative. 

Now it depends on how @openshwprojects wants to proceed to merge these implementations. 

There is also #1211 working on the same implementation.

PR #1187, opened 2024-04-21
Add AHT10/AHT2X/AHT30 I2C driver

NonPIayerCharacter:
AHT20/AHT21/AHT30 chip is almost pin-compatible with CHT8305 (SDA/SCL and VDD/GND are the same, while others on AHT are NC). So this driver is for those, who wish to replace it, those with TuyaMCU, who would like to remove it and wire sensor directly to module, or just add a very cheap board to their device.
Tested it with AHT10 board, AHT20+BMP280 board, and AHT21 chip soldered to door sensor board (which is the same, as generic sensor with SHT30 with exposed pad for CHT8305).
Binary size increased by 1248 bytes in debug build.

### DIFF
--- a/src/driver/drv_aht2x.c
+++ b/src/driver/drv_aht2x.c
@@ -1,0 +1,241 @@
+#include "../new_pins.h"
+#include "../logging/logging.h"
+#include "drv_local.h"
+#include "drv_aht2x.h"
+
+#define AHT2X_I2C_ADDR (0x38 << 1)
+
+static byte g_aht_secondsUntilNextMeasurement = 1, g_aht_secondsBetweenMeasurements = 10,
+	max_retries = 20, channel_temp = 0, channel_humid = 0;
+static float g_temp = 0.0, g_humid = 0.0, g_calTemp = 0.0, g_calHum = 0.0;
+static softI2C_t g_softI2C;
+static bool isWorking = false;
+
+void AHT2X_SoftReset()
+{
+	ADDLOG_DEBUG(LOG_FEATURE_SENSOR, "AHT2X_SoftReset: Resetting sensor");
+	Soft_I2C_Start(&g_softI2C, AHT2X_I2C_ADDR);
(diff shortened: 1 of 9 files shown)


openshwprojects:
Very nice and looks almost ready to merge but I think we need obk_config.h define for that. If that driver is for custom device hacks (removing the MCU), then maybe it could be not compiled by default?

divadiow:
aht30 test

image
image

divadiow:
Info:I2C:Address 0x38 (dec 56)

divadiow:
I have not been able to test AHT20 as the sensor I have is twinned with a BMP280. The i2c for that is only showing
Info:I2C:Address 0x77 (dec 119)

image

NonPIayerCharacter:
@divadiow i have several same modules, both sensors are detected normally.
Info:I2C:Address 0x38 (dec 56)
Info:I2C:Address 0x77 (dec 119)
Have you tested it elsewhere? I had BMP280 not working on one of mine, and had to solder a spare.
image

divadiow:
> Have you tested it elsewhere?

image

I'll play more.

openshwprojects:
This looks like ready to merge from my side. @divadiow , please open an issue for your problem

openshwprojects (approved):

divadiow:
hmm. still only getting BMP i2c address detected. Have bought some more bits before I give up. what pins did you connect your combi module to @NonPIayerCharacter 

image

divadiow:
hi. was just googling around for some bits and came across this about the polling interval https://forum.arduino.cc/t/compare-different-i2c-temperature-and-humidity-sensors-sht2x-sht3x-sht85/599609/43

does this driver take into account the polling interval recommendations for each model type?

divadiow:
I couldn't work out you username on Elektroda @NonPIayerCharacter but I did make this https://www.elektroda.com/rtvforum/topic4052685.html

thanks for your efforts with the driver.

PR #1191, opened 2024-05-01
Implement LittleFS for LN882H - only first basic testing from WebApp …

MaxineMuster:
Not tested in deep, only a first basic test of creating a file, which could be accessed after reboot.

- need to add littlefs-source to sdk/OpenLN882H/project/OpenBeken/CMakeLists.txt

Remarks:
I had to move the import part of LittleFS

in various cmd-source files to nearer to the top of the files

otherwise I had very strange (false!!!) compiler errors regarding "conflicting types" for equal(!) typed prototypes and functions or even variables in .h and .c files

e.g.:

/OpenBK7231T_App/sdk/OpenLN882H/project/OpenBeken/app/src/cmnds/cmd_main.c:47:6: error: conflicting types for 'g_powersave'
   47 | bool g_powersave;
      |      ^~~~~~~~~~~
In file included from /OpenBK7231T_App/sdk/OpenLN882H/project/OpenBeken/app/src/cmnds/cmd_local.h:4,
                 from /OpenBK7231T_App/sdk/OpenLN882H/project/OpenBeken/app/src/cmnds/cmd_main.c:6:
/OpenBK7231T_App/sdk/OpenLN882H/project/OpenBeken/app/src/cmnds/cmd_public.h:33:13: note: previous declaration of 'g_powersave' was here
   33 | extern bool g_powersave;
      |             ^~~~~~~~~~~

Also I had to disable other drivers, or the OTA image would not be flashed (after rebbot old version was still present) Is there a size limit to respect for OTA? the size was well below the shown OTA-size of 0xAA000 (696320k if I calculated correct)


giedriuslt:
These errors are not false. 
Problem is most likely that application defines bool as int while littlefs uses stdbool.h and redefines bool to _Bool, and these are not the same
changing to use stdbool.h should fix the issue without a need to move anything. Might need the same for stdint.h do no remember fully.

giedriuslt:
beken does not have those defines at all likely because bool is defined to stdbool.h in sdk already

MaxineMuster:
> I think it still mises change in user_main.c which i think isr esponsible for running sctipts from files.
> 
> Also maybe it's time to change this:
> 
> ```
> #if (defined WINDOWS) || (defined PLATFORM_BEKEN) || (defined PLATFORM_BL602) || (defined PLATFORM_LN882H)
> ```
> 
> to this
> 
> ```
> #if (defined ENABLE_LITTLEFS) 
> ```

I will have another look.
I thought it was just the line after the #define you mentioned:

CMD_InitScripting();

MaxineMuster:
Added the define in user_main.c, too

If "SVM_RunThreads(g_deltaTimeMS);" is depending on / can only run if LittleFS is present, then it would be a correction to change the defindes to `#if (defined ENABLE_LITTLEFS)`  to be able to build versions for this platform without LittleFS.

But I'm not that deep in this code ...

giedriuslt:
Tried to use such script, seems there is some issues.
I works partially that is ntp driver is started, but not fully, turning of socket  after 10 min doesn't appear to work.
Same for bl602 it seems, so there might be something missing with script execution still
```
// Used channels:
// Channel 1 - relay1
// Channel 2 - Btn on device
// Channel 5 - cycles counter CH1

//Reduce used module's power
PowerSave 1

// NTP driver must be enabled for its functions to work
startDriver ntp

// It might be useful to configure a local NTP server on your LAN so that devices do not need to connect to the internet
ntp_setServer 193.204.114.232

// Set the local timezone as NTP server only provide UTC time
ntp_timeZoneOfs 3

// Setting the devices location will allow for calculating sunrise and sunset times
//ntp_setLatlong 45.4722 9.1922

// Time values are available once NTP finishes initializing
waitFor NTPState 1

addEventHandler OnClick 10 POWER 1 toggle

// this will make channel 5 save in memory
setStartValue 5 -1

alias turnoff backlog SetChannel 1 0 echo TurnOFF

//Counter number of power on
// event triggers for channel 1 changing to 1: set countdown and inc cycles' counter
backlog addChangeHandler Channel1 == 1 addRepeatingEvent 600 1 turnoff; addChannel 5 1

alias turnon backlog SetChannel 1 1 echo TurnON

//Power on device @6AM
//[TimerSeconds or Time or sunrise or sunset] [WeekDayFlags] [UniqueIDForRemoval][Command]
addClockEvent 5:58 0x3e 123 turnon

```

openshwprojects:
you need one more define
https://github.com/openshwprojects/OpenBK7231T_App/blob/main/src/obk_config.h
// parse things like $CH1 or $hour etc
#define ENABLE_EXPAND_CONSTANT		1

MaxineMuster:
I can do so.
But the more I'm thinking about it: wouldn't it be better to introduce something like `#define scripting` and `#define scripting_advanced`?
This stuff might be only possible, if littleFS is present, but its not necessary for littleFS. On the other hand, what would you use files for, if not for scripting...

openshwprojects:
LittleFS can host a custom device panel made in http and JS:
https://www.elektroda.com/rtvforum/topic3971355.html

giedriuslt:
Found the issue, there are two defines missing here https://github.com/openshwprojects/OpenBK7231T_App/blob/70fc43bd4478adc9327fc39dcc5d78b57d99fb32/src/cmnds/cmd_eventHandlers.c#L302 and here https://github.com/openshwprojects/OpenBK7231T_App/blob/70fc43bd4478adc9327fc39dcc5d78b57d99fb32/src/cmnds/cmd_eventHandlers.c#L386

They  cause waitfor to not work

giedriuslt:
Also I think here define also needed:
https://github.com/openshwprojects/OpenBK7231T_App/blob/70fc43bd4478adc9327fc39dcc5d78b57d99fb32/src/cmnds/cmd_main.c#L272

MaxineMuster:
I'll add LN882H (and BL602 to the first two) to these defines.
But I vote for a new #define, for I think all this is code segments are not platform dependent but feature dependent...

giedriuslt:
Now my test looks fine

openshwprojects:
Sure, new define sounds good, maybe in next PR? 

Is this one PR good for merge?

MaxineMuster:
LFS should be o.k. 
I only know about the issues with "hass / tele discovery" in https://www.elektroda.com/rtvforum/topic4028087-180.html#21070551
Maybe heap issues?
But I don't use MQTT...

openshwprojects:
but does this LittleFS PR really has anything to do with heap (?) issue? Didn't heap issue occur also without it?

MaxineMuster:
I just wanted to point out that I can't further investigate this, since I'm not using MQTT.
And I don't know, if the mentioned functions are possibly without LFS...

openshwprojects:
I need this feature. I will merge it, hopefully later we can just disable with with #ifdef  , just in case

openshwprojects (approved):

Issue #1207, opened 2024-05-04
addClockEvent NOT found


delay0:
``add "addClockEvent..." to autoexec.bat has no effect.

Maybe I just don't see how to do it properly. All the examples I've tried don't work.

**Firmware:**
- Version 1.17.571
- Device: unknown (Surplife Bulp)
- Chip/model: BL602 BL-M5 V1.2
  "flags": "272"

send command e.g. "addClockEvent 15:48 0xff 1 autoOn"  via Log-Console returns:

Debug:CMD:cmd [addClockEvent 15:48 0xff 1 autoOn]
Error:CMD:cmd addClockEvent NOT found (args 15:48 0xff 1 autoOn)
Info:CMD:[WebApp Cmd 'addClockEvent 15:48 0xff 1 autoOn' Result] Unknown command

autoexec.bat ...

```
PowerSave 1
startDriver ntp
ntp_timeZoneOfs 2
ntp_setServer 194.25.134.196


alias autoOn setChannel 0 1
addChangeHandler Channel0 == 1 led_enableAll 1

alias autoOff setChannel 0 0
addChangeHandler Channel0 == 0 led_enableAll 0


addClockEvent 17:36:00 0x64 0 autoOn
addClockEvent 17:37:00 0x64 0 autoOff


waitFor NTPState 1

```

openshwprojects:
It should be fixed soo nthanks to @giedriuslt

PR #1208, opened 2024-05-04
Enable calendar events for bl602

giedriuslt:
Should fix the issue reported here  #1207


openshwprojects:
.. but will OTA still work?

giedriuslt:
Yes of course, OTA is fine until 800kb size

giedriuslt:
Tested and working fine

openshwprojects (approved):

PR #1210, opened 2024-05-05
Remove strftime to save space

giedriuslt:
strftime takes a lot of space and we only use simple number printing from it. 
Saves 7kb space for ln882h, w600, and 20kb(!) for bl602


openshwprojects:
Great find!

openshwprojects (approved):

PR #1212, opened 2024-05-06
try BL602 with ENABLE_EXPAND_CONSTANT	1 - please give feedback? is OTA still working?

openshwprojects:


openshwprojects:
@giedriuslt can you check?

giedriuslt:
OTA is working, why wouldn't it. Constant also seems to be expanded in scripts

Issue #1177, opened 2024-04-11
BL602 devices crash and reboot when fingered with port scanners


divadiow:
**Describe the bug**
MAIN ISSUE: I've tried Windows NMAP, Advanced Port Scanner and Angry IP Scanner, targeting port 80 specifically and separately, all ports. This has been done on two different machines - cabled main PC and virgin laptop on wifi. Each scan will cause BL602 watchdog to eventually reboot the device. Fing Android app on mobile does not have the same effect.

SECONDARY ISSUE: ICMP ping responses from BL devices are either non existent or intermittent. NOT an issue with wifi strength - different APs tested. http gui is still available on port despite no ping response, device remains online through ping attempts.

**Firmware:**
- Version Apr 10 2024 13:59:03 version 1.17.542
- any BL602 - dev board, SM-028 Qiachip relay and LF686 RGBW LED controller tested
- Chip/model: BL602
- Device config: fresh BL flash with and without any drivers - no difference

**To Reproduce**
1. Boot up BL602 device and connect to WLAN
2. scan for ports or just port 80
3. 
4. BL602 will crash and reboot. Reboot reason: BL_RST_SOFTWARE_WATCHDOG

maybe this isn't such a big issue - just don't port scan! my LN and BK devices disconnected wifi on scan (but then reconnected themselves) but did not crash/reboot, so BL is more sensitive.

dev board uart output at time of port scan:

```
[21:36:04.930] - , free 106728, MQTT 0(14), bWifi 1, secondsWithNoPing 165, socks 2/2
[21:36:04.931] - 1
[21:36:05.932] - Info:MAIN:Time 235, idle 0/s, fr
[21:36:05.933] - ee 106728, MQTT 0(14), bWifi 1, secondsWithNoPing 166, socks 2/2
[21:36:05.934] - 1
[21:36:06.932] - Info:MQTT:mqtt_host empty, not s
[21:36:06.936] - tarting mqtt
Info:MAIN:Time 236, idle 0/s, free 106728, MQTT 0(
[21:36:06.937] - 15), bWifi 1, secondsWithNoPing
[21:36:06.943] -  167, socks 2/21
[21:36:07.939] - Info:MAIN:Time 237, idle 0/s, fr
[21:36:07.940] - ee 106728, MQTT 0(15), bWifi 1, secondsWithNoPing 168, socks 2/2
[21:36:07.940] - 1
[21:36:08.931] - Error:HTTP:TCP Client is disconn
[...]

PR #1224, opened 2024-05-10
Fix unknown method handling

giedriuslt:
This fixes crash on bl602 when unknow http method is sent, solves part of #1177, however still crashes somewhere else


giedriuslt:
with logging fix seems no longer crash

openshwprojects:
I can't really test that myself but I think Iwill merge it anyway. i really appreciate your help, @giedriuslt , keep it on! I'm just a one person and it's hard to handle all those platforms myself.

openshwprojects (approved):

PR #1229, opened 2024-05-12
allow bl602 to use constants in publishFloat ant similar

giedriuslt:
It was under platform define, should be fixed to proper define


openshwprojects:
no problem, please introduce a new define and I will accept next PR as well. Thanks!

giedriuslt:
Same define is neede for LN882 I think, there we also have expand_constants set

PR #1231, opened 2024-05-13
Add bme280 humidity

NonPIayerCharacter:
Measurement code copied from esphome.
Don't have a bme280, but bmp280 continue to work normally.
Can try to add bme680, but it is significantly different from bmx280, so the code here will not work.
So if i'm to try, am i to code it in this library, or add a new one instead?


openshwprojects:
Very good, thank you! I don't have this chip as well but maybe @divadiow  has?

Regarding 680 - well, if you say it's different, then create a separate driver, you know, separate entry in drv_main.c, then a file drv_bme680, etc. You can also ask me for help anytime and I will try my best. Just make sure to put your driver disabled by default because we are trying to avoid reaching max OTA size on some platforms

divadiow:
only BME680/BMP280 currently. will buy BME280

openshwprojects:
What is the state of this feature? Has it been already tested? I would like to merge it but i dont have required chip at hand

divadiow:
if it's any help, my BME280 should arrive this week

divadiow:
image

openshwprojects:
Thank you, I think I can merge it now

openshwprojects (approved):

DeDaMrAzR:
to initiate and read BME280 use the following in your autoexec file:


// startDriver BMP280 [ClkPin] [DatPin] [OptionalTargetChannel]
startDriver BMP280 7 8 2
BMP280_Adr 0x76

PR #1249, opened 2024-05-25
let echo return expanded arguments

MaxineMuster:
I wondered why echo would never expand expressions like "$ch10" but return the string as is (e.g.: "Info:CMD:$ch10").
I think it was just a typo, for "Tokenizer_GetArgFrom(0)" would just return g_argsFrom[0] and not expanding content like "Tokenizer_GetArg(0)".
And since we forced a single argument, this should be correct, I hope.

At least, it now works for me ;-)

### DIFF
--- a/src/cmnds/cmd_main.c
+++ b/src/cmnds/cmd_main.c
@@ -276,22 +276,11 @@ static commandResult_t CMD_Echo(const void* context, const char* cmd, const char
 #else
 	// we want $CH40 etc expanded
 	Tokenizer_TokenizeString(args, TOKENIZER_ALTERNATE_EXPAND_AT_START | TOKENIZER_FORCE_SINGLE_ARGUMENT_MODE);
-	ADDLOG_INFO(LOG_FEATURE_CMD, Tokenizer_GetArgFrom(0));
+	ADDLOG_INFO(LOG_FEATURE_CMD, Tokenizer_GetArg(0));
 #endif
 
 	return CMD_RES_OK;
 }
-
-// setChannel 1 123
-// echo $CH1
-// will print 123
-static commandResult_t CMD_Echo_Val(const void* context, const char* cmd, const char* args, int cmdFlags) {
-	// we want $CH40 etc expanded
-	Tokenizer_TokenizeString(args, TOKENIZER_ALTERNATE_EXPAND_AT_START | TOKENIZER_FORCE_SINGLE_ARGUMENT_MODE);
-	ADDLOG_INFO(LOG_FEATURE_CMD, Tokenizer_GetArg(0));
-	return CMD_RES_OK;
-}
-
 static commandResult_t CMD_StartupCommand(const void* context, const char* cmd, const char* args, int cmdFlags) {
 	const char *cmdToSet;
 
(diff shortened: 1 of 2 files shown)


MaxineMuster:
This was the situation before: Expanding worked e.g. in setChannel command, but not with "echo"
Bildschirmfoto vom 2024-05-25 18-09-35
Bildschirmfoto vom 2024-05-25 18-12-00
Bildschirmfoto vom 2024-05-25 18-12-35

Now it works like expected for me:
Bildschirmfoto vom 2024-05-25 18-14-31

openshwprojects:
I think that's not that easy.

What about "echo hello world" on your and official versions? :D 

Ye, we need to think about it..

MaxineMuster:
Hm, I don't see any difference?!?
And at least now it does, what the comment says,

```
// setChannel 1 123
// echo First channel is $CH1 and this is the test
// will print echo First channel is 123 and this is the test
```

 at least for me, it would write with the original version:
 "First channel is $CH1 and this is the test"

MaxineMuster:
Ah, I see, using some longer text, it will get truncated..

MaxineMuster:
What do you think about

```
static commandResult_t CMD_Echo(const void* context, const char* cmd, const char* args, int cmdFlags) {

#if 0
	ADDLOG_INFO(LOG_FEATURE_CMD, args);
#else
	// we want $CH40 etc expanded
	Tokenizer_TokenizeString(args, TOKENIZER_ALTERNATE_EXPAND_AT_START | TOKENIZER_FORCE_SINGLE_ARGUMENT_MODE);
	char *line = CMD_ExpandingStrdup(Tokenizer_GetArgFrom(0));
	ADDLOG_INFO(LOG_FEATURE_CMD, line);
	free(line);
#endif

	return CMD_RES_OK;
}

```

This seems to work o.k.:
Bildschirmfoto vom 2024-05-25 19-31-03
Bildschirmfoto vom 2024-05-25 19-31-26

openshwprojects:
That's a better, I think I can accept it, however I always have doubts about using strdup (malloc) and free too eagerly, but maybe I am overthinking.... futhermore echo is not so standard command I guess.

openshwprojects:
PS: Btw I am sometimes a bit slow to check Github, so if you have some suggestions next time, you can also consider asking on forum, I am usually checking forum first, then Git.

MaxineMuster:
I also think it's used not too much.
I only (tried) to use it once before, when trying to find the value of some time functions (like "$day") and it didn't work.

Another possible way would be to introduce "echo_value" (just "copy" my first approach to a new name)
Yes, this can only display 40 chars, but who cares, if its just about checking the value of $sunrise or similar
I could live with the fact it can only display the value and no long text ;-) 
And we would avoid some additional malloc/free

MaxineMuster:
> PS: Btw I am sometimes a bit slow to check Github, so if you have some suggestions next time, you can also consider asking on forum, I am usually checking forum first, then Git.

What do you mean by "slow"? I wasn't even ready with appending some screenshots when you answered and directly pointed out the catch of my "correction" ;-)

openshwprojects (approved):

openshwprojects:
Hm I think we can just merge echoval with echo and allow the behaviour with strdup in echo

MaxineMuster:
Yes, I know it's demanding, but may I ask for another possible change ?!?

Is there anything else I didn't take into account in my first approach beside that it will only convert 40 chars (limited by "g_argsExpanded[0]" being 40 chars?

If that is the only problem, I ask to consider the following:
```

	if (flags & TOKENIZER_FORCE_SINGLE_ARGUMENT_MODE) {
		g_args[g_numArgs] = g_buffer;
		g_argsFrom[g_numArgs] = g_buffer;
		// some hack, but we fored to have only have one arg, so we can extend the string over array bondaries.
		// probably better: introducing an union containing g_argsExpanded[][] and one sole string in the same memory area ...
		CMD_ExpandConstantsWithinString(g_buffer,g_argsExpanded,sizeof(g_argsExpanded)-1);
		g_numArgs = 1;
		return;
	}

```
As far as I see, TOKENIZER_FORCE_SINGLE_ARGUMENT_MODE is only used in echo (but it should not be a problem in other cases, too).
So we force all arguments into only one argument (leading to the problem when expanding it).
But since we will never have other arguments than argument 0, we can (mis-)use the space reserved for expanded other arguments for our "large" argument 0.
Since the GetArg() function will check, if an argument is already expanded:
```

const char *Tokenizer_GetArg(int i) {
	const char *s;

	if (i >= g_numArgs)
		return 0;

	if (g_argsExpanded[i][0] != 0) {
		return g_argsExpanded[i];
	}

``` this will work. 

It's not nice to ignore the array boundaries and would be "nicer" to make this clear e.g. as a union, but it works as requested:
We can "leave" echoval (now showing even very long lines) or just converting echo to use the expanded version

openshwprojects:
I must have forgot about it, but you have a point. So, remove echoval and fix echo with the solution you proposed.

Maybe it would also make sense someday to do somtehing like:
char *g_args[MAX_ARGS]
char argsBuffer[4096];
and then dynamically put args in argsBuffer, but not now

openshwprojects (approved):

Issue #1246, opened 2024-05-23
Not connecting to SSID2


xjikka:
I am trying to use SSID2 with no success. It always connects to SSID1 only (CFG_GetWiFiSSID() and CFG_GetWiFiPass()).
I couldn't find code in the source code that uses CFG_GetWiFiSSID2() and CFG_GetWiFiPass2(). 
Is SSID2 solved somehow differently, or is it not yet implemented?

MaxineMuster:
It's not yet implemented, as you found out correctly.
Maybe a hint in GUI would be helpful

xjikka:
> It's not yet implemented, as you found out correctly. Maybe a hint in GUI would be helpful

I agree. Maybe some change like this:
![image](https://github.com/openshwprojects/OpenBK7231T_App/assets/38142124/6a677620-fe7f-40f3-88c7-5e551fc1a011)

[http_fns.zip](https://github.com/openshwprojects/OpenBK7231T_App/files/15429381/http_fns.zip)
(not tested, just suggestion)

xjikka:
Anyway, I need SSID2 and will make some simple changes in the code to make it working somehow. If it's at least somewhat usable, I'll post it here.

xjikka:
I have SSID2 simple mod done. It is tested on BK7231N, but with these simple changes it should work on any platform. 

How does it work:
It tries g_SSIDSwitchAfterTry times to connect to the SSID and if it fails it switches to SSID2 (if set) and so on after 3 more failed attempts back to SSID and so on.

Here are the differences:
https://github.com/xjikka/OpenBK7231T_App/compare/main...ssid2mod

[user_main.zip](https://github.com/openshwprojects/OpenBK7231T_App/files/15447126/user_main.zip)

xjikka:
If You will find it OK, I can submit these changes here.

MaxineMuster:
Thanks for your code!

My personal approach would be a bit different, I would make the first SSID "preferred" so in pseudo code (I will "rename" SSID to SSID1 to get it clearer ;-)):
Every time I try to connect to WiFi, I first try SSID1 and increment a counter
If counter > threshold, also try SSID2

If WiFi is connected, reset counter.


But I hope it's o.k. if I give some feedback on your code?
Because I think it could be simplified in some points:

```
// static int g_SSIDactual = 0;			// 0=firsttime 1=ssid1 2=ssid2		

[...]

PR #1254, opened 2024-05-31
SSID2 autoswitch mod

xjikka:
How does it work:
It tries g_SSIDSwitchAfterTry times to connect to the SSID and if it fails it switches to SSID2 (if set) and so on after 3 more failed attempts back to SSID and so on.

Discussed here: https://github.com/openshwprojects/OpenBK7231T_App/issues/1246

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ OpenBeken features:
 - NTP time from network (can be used with [TH06](https://www.elektroda.com/rtvforum/topic3942730.html) and other TuyaMCU devices), can run any script on selected weekday hour:minute:second
 - Dedicated [TuyaMCU support](https://www.elektroda.com/rtvforum/topic4038151.html) with extra TuyaMCU analyzer tool for decoding new devices ([tutorial here](https://www.elektroda.com/rtvforum/topic3970199.html), code [repository here](https://github.com/openshwprojects/TuyaMCUAnalyzer))
 - support for [TuyaMCU Battery Powered devices protocol](https://www.elektroda.com/rtvforum/topic3914412.html) (TuyaMCU enables WiFi module only to report the state, eg. for door sensors, water sensors)
-- [RGBCW LED lighting control](https://www.youtube.com/watch?v=YQdR7r6lXRY) compatible with Home Assistant (including PWM LEDs, and SM2135, BP5758, SM15155 etc )
+- [RGBCW LED lighting control](https://www.youtube.com/watch?v=YQdR7r6lXRY) compatible with Home Assistant (including PWM LEDs, and SM2135, BP5758, [SM15155](https://www.elektroda.com/rtvforum/topic4060227.html) etc )
 - LittleFS integration for scripts and large files (you can [write scripts there](https://www.youtube.com/watch?v=kXi8S12tmC8), you can host a page there with [REST interface control](https://www.elektroda.com/rtvforum/topic3971355.html) of device)
 - Command line system for starting and configuring drivers, for controlling channels, etc
 - Short startup command (up to 512 characters) storage in flash config, so you can easily init your drivers (eg. BL0942) without LittleFS


openshwprojects:
I think this should be enabled by default. Is ALLOW_SSID2 enabled in this PR?

@divadiow , would you be able to test that? I have just a single router...

xjikka:
ALLOW_SSID2 id disabled for W800 and W600 in the new_pins.h
for the other platforms is enabled.

new_pins.h:
...
#if PLATFORM_W600 || PLATFORM_W800
#define ALLOW_SSID2 0
#define ALLOW_WEB_PASSWORD 0
	char initCommandLine[512];
#else
#define ALLOW_SSID2 1
#define ALLOW_WEB_PASSWORD 1
...

divadiow:
Hi. I'm away but will be back tomorrow evening and will test then or Monday

divadiow:
OK. With 1254_merge_689545a11ddf

Two SSIDs setup on same router, OBK1 and OBK2. Same security profile so identical pw and encryption types. Password was super simple with no special characters.

image
image

From downing of SSID1 (no power cycle):
BK-T took ~ 5 mins to connect to OBK2
BK-N, BL, LN didn't connect to OBK2 (waited 20 mins)

From power-on with SSID1 down:
BK-N & T took ~3mins 45s to connect to OBK2
LN882H took ~6 minutes to connect to OBK2
BL602 did not connect (waited 20 minutes). (BL602 did reconnect quickly to OBK1 when that was re-broadcasting)

From downing of SSID2 (with N, T &LN connected) and reinstatement of SSID1 (no power cycle):
BK-N & T connected to OBK1 after a few minutes (lost track - 3-6 mins?)
LN882H did not connect to OBK1 (waited 20 mins)

MaxineMuster:
To speed up the times I would suggest to lower the seconds to wait inside Main_OnWiFiStatusChange() (g_connectToWiFi is used as a timer here). 
Especially 120 seconds for WIFI_STA_CONNECTING is very long. For testing setting all to 10 or 15 seconds would help I think.

divadiow:
> From downing of SSID1 (no power cycle):
> BK-T took ~ 5 mins to connect to OBK2
> BK-N, BL, LN didn't connect to OBK2 (waited 20 mins)

I'm going to rerun this now

divadiow:
From downing of SSID1 with all connected (no power cycle):

BK-T - 4 min 48s to connect to OBK2
BK-N - no connection after 20 mins
LN882H - "
BL602 - "

xjikka:
BK7231N - I'm using it and it switches to SSID2 in both cases on my router (both SSIDs are simple SSID + simple password)
bk7231n-no-SSID1-on-poweron.log - SSID1 not available at power on
bk7231n-SSID1-on-poweron-disabled-after-connect.log - SSID1 available at power on, in short time later disabled
[bk7231n-no-SSID1-on-poweron.log](https://github.com/user-attachments/files/15539409/bk7231n-no-SSID1-on-poweron.log)
[bk7231n-SSID1-on-poweron-disabled-after-connect.log](https://github.com/user-attachments/files/15539410/bk7231n-SSID1-on-poweron-disabled-after-connect.log)

xjikka:
I'll try to recompile it and I'll check again on the BK7231N

divadiow:
I'll redo N flash, factory reset and test again in the morning. I was expecting N to behave like T and your experience with N so hopefully just a crappy test from me.

xjikka:
Thanks for the info.
I also re-compiled the App again, tested it again on the BK7231N with the same result (switching SSID 1 and 2 in both scenarios).
As a bonus, I tested it using hidden SSIDs and it works as expected.

divadiow:
cool, so I think in summary it feels like BK-N/T is good but the rest need more work?

xjikka:
I have no other devices available, only BK-T and BK-N. 
So I suggest to modify the definition so that SSID2 is offered (in the webconfig) only for Beken so:

#ifdef PLATFORM_BEKEN
#define ALLOW_SSID2 1
#else
#define ALLOW_SSID2 0
#endif
#if PLATFORM_W600 || PLATFORM_W800
#define ALLOW_WEB_PASSWORD 0
char initCommandLine[512];
#else
#define ALLOW_WEB_PASSWORD 1
...

MaxineMuster:
I'm not at home at the moment, so I can't check the logs, what e.g. LN882H will do when WiFi is down.
It works if connection isn't possible due to SSID or password mismatch. I didn't test a "vanishing" SSID yet.

As I said, it only works, when Main_OnWiFiStatusChange() is called in these cases (or we would need another mechanism to set `g_connectToWiFi` to a value > 0).

divadiow:
lmk if/when you want me to redo LN test on multiple reset devices to confirm current state on LN

xjikka:
I have added/changed ssid2mod to src/new_pins.h

#ifdef PLATFORM_BEKEN
#define ALLOW_SSID2 1
#else
#define ALLOW_SSID2 0
#endif

openshwprojects:
@xjikka please double check if you are not changing config layout on non-Beken platform, if you hide SSID data from config, you will break configs on those platforms. So take care, do not change #ifdef in the config struct

xjikka:
> @xjikka please double check if you are not changing config layout on non-Beken platform, if you hide SSID data from config, you will break configs on those platforms. So take care, do not change #ifdef in the config struct

As I was going through the source code I know that CFG_SetWiFiSSID2() and CFG_GetWiFiPass2() haven't been used anywhere so far.
But you're right that someone could theoretically have passwords/SSIDs stored there in thede fields SSID2 as a backup even when not in use.
Storing via wifi_pass2 and wifi_ssid2 will be preserved, however it won't be editable.
This has to be decided by someone else. The other option is to edit the text for SSID2 in WebInterface to make it clear that it only works on BEKEN

openshwprojects:
I think you can easily add #ifdef conditional for PLATFORM_BEKEN to add some extra HTML text on the page. Please try, will it be then ready to merge?

PR #1256, opened 2024-06-02
First attempt to turn device in an access point with WPA password.

MaxineMuster:
Tested only with LN882H
Since it alters wifi code, I would advice to use it only if UART flashing is possible during test.
It's just a test up to now, for I can only test on LN882H device.
Actually you can only do it on WiFi config page:
Bildschirmfoto vom 2024-06-02 21-51-45

IP is not altered from AP code - it's using 192.168.4.1 as the open AP of OBK

### DIFF
--- a/src/hal/bk7231/hal_wifi_bk7231.c
+++ b/src/hal/bk7231/hal_wifi_bk7231.c
@@ -273,7 +273,8 @@ void HAL_DisconnectFromWifi()
     bk_wlan_stop(STATION);
 }
 
-int HAL_SetupWiFiOpenAccessPoint(const char* ssid)
+// setup openAP with key=NULL, WPA2 AP with key !="" 
+int HAL_SetupWiFiAP(const char* ssid, const char* key)
 {
 #define APP_DRONE_DEF_NET_IP        "192.168.4.1"
 #define APP_DRONE_DEF_NET_MASK      "255.255.255.0"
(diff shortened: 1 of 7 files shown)


divadiow:
image

divadiow:
column H & I are with both AP and STA broadcasting

MaxineMuster:
Just got my BK7231N module, and it's working with this code (I simply forgot to copy the key).
This was due to the fact that everything was first inserted in a struct, which was later only used to copy its content to the struct needed to setup the AP.
I got rid of this step now.

I only have the problem, that I can only connect one client to the AP, but this seems to be the same with the open AP of release firmware.

I found one possible problem in W800 code, maybe that fixes the issue with non functional password...

MaxineMuster:
@divadiow if you find some time could you please give the new version a try?
Especially the W600/W800 (I hope BK7231T will work like the N module of mine).

Thanks

divadiow:
Ha. Was just coming here to say I can test more today and you posted an hour ago.

Yes. I'll do more and update shortly

divadiow:
Success with BK now

image

sadly W600/W800 remain as above in table

MaxineMuster:
Thanks again!
I'll try to find out how to increase the number of clients with BK (CFG_STA_MAX alone won't help, I'll look into DHCP now)

Could you test, if on BL602 more than one client can connect (e.g. test with phone or tablet additionally)?

Regarding W600/W800: I'll take another look, and will try to get my hands on one myself.
W8001 should work like W800, right ? There are some development boards on ali with W801/W806 and W801 seems to be the WiFi version ...

divadiow:
> W8001 should work like W800, right ? There are some development boards on ali with W801/W806 and W801 seems to be the WiFi version ...

I guess so. Mine is W800-C400 dev board.

I'll check BL602.

divadiow:
two clients can connect to BL602 but no more than 2

MaxineMuster:
From all I see W600/W800 code looks like the code in demo_ap..

And also I didn't succeed in allowing more than one client to BK device. 
After setting `#define CFG_STA_MAX 5` in   `platforms/bk7231n/bk7231n_os/beken378/ip/lmac/src/rwnx/rwnx_config.h`   my phone no longer shows that no more stations are allowed on that SSID, but connection isn't successful :-(. 
The mechanism of adding a station with some message is just too much for me to follow/debug.

So all in all I was not too successful today.

But I Just ordered some W801 dev boards, being able to set and see some debug output will hopefully help here...

divadiow:
re LN882H AP mode, saw this a while ago. just FYI if relevant ever when working on LN code

`7.Fixed the bug that in softAP mode, when multiple devices are connected at the same time, they may not be able to connect`

https://gitee.com/lightningsemi/ln882h/commit/b2d811378d07960f3800d8b3f0ac3145f961f127

MaxineMuster:
Thanks, I'll take a look. 
As usual you will only get a new blob, a new library without knowledge, what they changed in detail.

MaxineMuster:
Just got my W801 dev boards. For me, the last build works, I could connect at least phone and PC at the same time.
But only very few seconds - then the WiFi client / STA code kicks in, and AP disappears and device connects to my AP.
It works if I start from open AP

So at least I know where do look some further.

But the chip seems to get very hot and I had some crashes, not really sure, why.

As I'm very busy ATM, it will take some time to investigate further, but I wanted to give a short feedback now

MaxineMuster:
No longer needed, newer PR present

Issue #1270, opened 2024-06-12
Incorrect calculations for BL0937 power meter pulses


rpv-tomsk:
https://github.com/openshwprojects/OpenBK7231T_App/blob/59fcf19bbdee45dee56f23e74ac3b8e1cacca5da/src/driver/drv_bl0937.c#L354-L361

I'm sorry if I'm wrong, but for me this code  looks incorrect. 
If we count more pulses due to more `ticksElapsed` since previous counter reset, the measured value (Voltage,Current, Power) should be less, not greater. Division and multiplication are swapped here.

giedriuslt:
Could this incorrect calculation cause wrong values with powersave? I need to try

PR #1280, opened 2024-06-24
BL0937: Fixed swapped division and multiplication in measured value correction

rpv-tomsk:
value = pulses / time

More pulses for same time - bigger value.
More time for same pulses count - smaller value.
Twice more pulses for twice time - same value.

ticksElapased * tick_period == elapsed time in ms
ticksElapased * tick_period / 1000 == elapsed time in s

final = final * expected_time / elapsed_time; //expected time is 1 second.
final = final * expected_ticks / elapsed_ticks; //expected ticks for 1 second: (1000.0ms / portTICK_PERIOD_MS)

Closes: #1270

### DIFF
--- a/src/driver/drv_bl0937.c
+++ b/src/driver/drv_bl0937.c
@@ -293,7 +293,6 @@ void BL0937_RunEverySecond(void) {
 		bNeedRestart = true;
 	}
 
-	ticksElapsed = (xTaskGetTickCount() - pulseStamp);
 
 #if PLATFORM_BEKEN
 	GLOBAL_INT_DECLARATION();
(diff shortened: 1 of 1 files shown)


openshwprojects:
I will need to... think about it. I am not the author of that division for BL0937. Would it be possible to somehow... make an experiment to verify which version is correct? Hm, I could try ticking drivers every 2 seconds, not 1.... and then see..

rpv-tomsk:
I don't think experiment is needed, because of clear principle of work: the frequency is proportional to the value.
More the value - more pulses per fixed unit of time. Pure math.

rpv-tomsk:
> Hm, I could try ticking drivers every 2 seconds, not 1.... and then see..

Yes, that should produce 4x greater measurement. 2x - twice the counted pulses  and 2x - from `ticksElapsed`.
But correct implementation should leave result unchanged and independent from ticks period.

rpv-tomsk:
Another issue is `PwrCal_Scale()` design. We have to use current/voltage measurement unchanged from previous cycle, but implementation does 'correction'. Due to `ticksElapsed` mismatch between cycles, that produces additional error in measurement.

giedriuslt:
@openshwprojects could we merge this?
The math here is correct, anfd this bug probably causes some occasional spikes I see.

Here is small demonstration I made by modifying period on which Main_OnEverySecond runs. The calibration is not modified.
If period increases to 1.1s for example voltage reported wrongly increases to \~280. which is because over 1.1s 1.1 times more pulses is counted and current firmware then multiplies it (**wrongly**) by 1.1 again for 1.21 increase (230*1.21\~280)
This fix makes calculation correct.
di9zG3RUG1

openshwprojects:
Ok, you have convinced me. I'll merge it now. I just needed someone else to confirm on important changes.

openshwprojects (approved):

PR #1288, opened 2024-07-10
DS18x20 driver

divadiow:


MaxineMuster:
Maybe you want to take a look here:

https://github.com/MaxineMuster/OpenBK7231T_App/tree/DS1820

It's not the whole thing, but if you only want to use one device, it sort of works for me.

When I saw your try, I finally got the energy to fix my approach (the timing didn't work before, so I only read no values at all or only crap).

Since I don't have a logic analyzer I didn't see, what was wrong.

Finally I came up with the idea to give out an info about how long a sleep of XXX usecs was in ms (derived from xTicks) if reset or device failed, that did the trick to show, how wrong my expectations where.

This is just a proof of concept, showing the temp every 5 seconds in log ...

PR #1299, opened 2024-07-30
Change "Build" to "Built" in the build date text

divadiow:
Reason for the Change:
The current text "Build on xxxxx" is grammatically incorrect in this context. The word "build" is a verb in its base form, which is used to describe the action of constructing or creating something. However, in the context of displaying the date and time when the program was constructed, the past participle "built" should be used. This is because we are referring to an action that has been completed in the past.


divadiow:
yay cool thanks. 

image

yes, even on holiday I have a single CBU SPI LED device running OBK/PixelAnim that I can flash!

PR #1300, opened 2024-08-05
Update cmd_main.c - powersave description update

divadiow:
https://github.com/openshwprojects/OpenBK7231T_App/pull/1186

### DIFF
--- a/src/cmnds/cmd_main.c
+++ b/src/cmnds/cmd_main.c
@@ -686,7 +686,7 @@ void CMD_Init_Early() {
 	//cmddetail:"examples":""}
 	CMD_RegisterCommand("DeepSleep", CMD_DeepSleep, NULL);
 	//cmddetail:{"name":"PowerSave","args":"[Optional 1 or 0, by default 1 is assumed]",
-	//cmddetail:"descr":"Enables dynamic power saving mode on BK and W600. You can also disable power saving with PowerSave 0.",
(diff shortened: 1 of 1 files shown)


openshwprojects (approved):

MaxineMuster:
Hi,

iirc, for LN882H it was "ln_pm_sleep_mode_set(LIGHT_SLEEP)" which rendered results from BL0937 unusable.
So we introduced "PowerSave 2" to set this mode, leaving `PowerSave 1`  to work with  BL0937 present.

So, if I'm not mistaken, the text should be:

On LN882H PowerSave >1 (eg `PowerSave 2`) should **not** be used if BL0937 metering is present.

Or: 

On LN882H only `PowerSave 1` should be used if BL0937 metering is present.

Regards

divadiow:
right OK. I was reading it the other way round.

so on LN `powersave 1` is a deeper sleep than `powersave >1`?

and the lighter sleep makes BL0937 unusable whereas the deeper sleep is fine with BL0937?

MaxineMuster:
Nope.

1 is "lighter" sleep (and doesn't seem to impact pulse counting)
2 is "deeper" sleep (and pulse counting is not working correct)

So, you should not use >1 if you want a correct reading.

divadiow:
lol. that makes more sense. I'll do PR to fix. ta

MaxineMuster:
Just to have everything in one place:

PowerSave 1 uses only `wifi_sta_set_powersave(WIFI_MAX_POWERSAVE)`, this seems to have no impact on interrupts, which are needed for counting ...

divadiow:
excellent

PR #1310, opened 2024-08-11
CHT831X alert add humidity

NonPIayerCharacter:
Added humidity to alert configuration.
On driver init, will read status register and log what triggered the alert. (https://github.com/NonPIayerCharacter/OpenBK7231T_App/blob/7c5fd07de0011c88eb8293b7ad91987b2ecda2ee/src/driver/drv_cht8305.c#L257)
Update docs, taken from #1302


openshwprojects:
Thank you, however, I had to downgrade switch to if to support older compilers, can you solve merge issues?
Once done, I think I can merge it.

NonPIayerCharacter:
Decided to use switch (without case x ... y:), and if input value is wrong - just print a warning and select default. Should be a little faster and more lightweight than if/else.

PR #1316, opened 2024-08-15
Changes / improvements for DS1820

MaxineMuster:
Changes for DS1820:
- change timing for BEKEN - works with powerSave 1 now 
- introduce  "critical" sections for reading/writing bits as proposed by @giedriuslt in openshwprojects#1294 
- add DS1820 driver for BL602 - tested in  openshwprojects#1294

Tested on BK7231N (with powersave 0 and 1) 
LN882H (with powersave 0/1/2)
W800 (still not working out of the box for missing support for "drivers" here)

To make it work on W800:
```
diff --git a/src/httpserver/http_fns.c b/src/httpserver/http_fns.c
index 385518d5..c47e66b1 100644
--- a/src/httpserver/http_fns.c
+++ b/src/httpserver/http_fns.c
@@ -720,6 +720,8 @@ int http_fn_index(http_request_t* request) {
        poststr(request, "</table>");
 #ifndef OBK_DISABLE_ALL_DRIVERS
        DRV_AppendInformationToHTTPIndexPage(request);
+#elif (PLATFORM_W800 && ENABLE_DRIVER_DS1820)
+       DRV_AppendInformationToHTTPIndexPage(request);
 #endif
 
        if (1) {
diff --git a/src/user_main.c b/src/user_main.c
index b0b5ff3d..d8336cdf 100644
--- a/src/user_main.c
+++ b/src/user_main.c
@@ -533,8 +533,11 @@ void Main_OnEverySecond()
 #if defined(PLATFORM_BEKEN) || defined(WINDOWS) || defined(PLATFORM_BL602)
        UART_RunEverySecond();
 #endif
+#elif (PLATFORM_W800 && ENABLE_DRIVER_DS1820)
+       DRV_OnEverySecond();
 #endif
 
+
 #if WINDOWS
 #elif PLATFORM_BL602
 #elif PLATFORM_W600 || PLATFORM_W800
@@ -1040,6 +1043,8 @@ void Main_Init_BeforeDelay_Unsafe(bool bAutoRunScripts) {
        g_unsafeInitDone = true;
 #ifndef OBK_DISABLE_ALL_DRIVERS
        DRV_Generic_Init();
+#elif (PLATFORM_W800 && ENABLE_DRIVER_DS1820)
+       DRV_Generic_Init();
 #endif
 #ifdef PLATFORM_BEKEN
        int bk_misc_get_start_type();

```


openshwprojects:
Looks very good, I can merge it right away, but the proposed patch for drivers on next paltform is not correct - we should rather fix old conditions instead of introducing new ones like:
"+#elif (PLATFORM_W800 && ENABLE_DRIVER_DS1820)"

MaxineMuster:
> Looks very good, I can merge it right away, but the proposed patch for drivers on next paltform is not correct - we should rather fix old conditions instead of introducing new ones like: "+#elif (PLATFORM_W800 && ENABLE_DRIVER_DS1820)"

Shure, thats more a note to myself...
The "problem" is, W800 is built with "OBK_DISABLE_ALL_DRIVERS", so all the mechanisms for handling drivers won't work here, so the W800 build process needs rework.

PR #1326, opened 2024-08-20
fix flicker in DDP (fixes #1325)

### LINKED ISSUE #1325: RGB LED Strip flickering during effects (with DDP)
While displaying color changing effects, the RGB LEDs flicker constantly while the colors change.

Since we have no effects (yet) on device, I'm sending some color changes and color wheels via ledfx or xLights over the DDP protocol. That works as intended. But there is a very annoying flicker going on constantly.

- Firmware Version 1.17.663
- Device: [LSC LED Strip 5m 3203632](https://www.elektroda.com/rtvforum/topic4034357.html)
- Chip/model: BK7231N

1. Enable the SM16703P driver
2. Enable the DDP driver
3. Send some dynamic effects with xLights or ledfx (like a color wheel)
4. See error, the RGB LEDs flicker while displaying the effect

h3ndrik:
The DDP_Parse() function should write the incoming DDP data to the LEDs. But it did it twice per incoming packet. First with LED_SetFinalRGB() which set all LEDs to a single color(?) and then again with SM16703P_setMultiplePixel() to the correct values. Resulting in flickering LEDs.

This fixes that and writes the pixels with either the SM16703P function OR the other one.

fixes https://github.com/openshwprojects/OpenBK7231T_App/issues/1325

### DIFF
--- a/src/driver/drv_ddp.c
+++ b/src/driver/drv_ddp.c
@@ -12,6 +12,10 @@
 #include "lwip/inet.h"
 #include "../httpserver/new_http.h"
 
+#if ENABLE_DRIVER_SM16703P
+#include "drv_spiLED.h"
+#endif
+
 static const char* group = "239.255.250.250";
 static int port = 4048;
 static int g_ddp_socket_receive = -1;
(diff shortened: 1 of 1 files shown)


h3ndrik:
I think this can be merged. I've tested this for a day now and it completely removes the flicker.

openshwprojects:
wait a sec, what if you have SM16 etc driver included in build, but it's not run, and you have RGB bulb?

h3ndrik:
> wait a sec, what if you have SM16 etc driver included in build, but it's not run, and you have RGB bulb?

Ah right, I didn't think of that. Can we somehow make the decision on runtime? Maybe with `spiLED.ready`? I mean as is it's also not correct...

openshwprojects:
there is DRV_IsRunning but it seems it may be slow... it iterates an array...

do we have some "ready" variable in that SPI LED driver? Maybe it would be good just to use it directly, just as you said

h3ndrik:
So. I've force pushed a new fix. Taking into account both cases. I've tested it briefly but just for my use case with the driver compiled in and the addressable LEDs loaded.

openshwprojects:
sooo only bl602 fix is left

h3ndrik:
alright, next try. thanks for the patience.

openshwprojects:
great thx

PR #1328, opened 2024-08-22
LN882H PWM

NonPIayerCharacter:
Requires https://github.com/openshwprojects/OpenLN882H/pull/15
Needs further testing. I tested it only on 1 channel.

### DIFF
--- a/src/hal/ln882h/hal_pins_ln882h.c
+++ b/src/hal/ln882h/hal_pins_ln882h.c
@@ -1,49 +1,54 @@
 #ifdef PLATFORM_LN882H
 
-
 #include "../../new_common.h"
 #include "../../logging/logging.h"
 #include "../../new_cfg.h"
 #include "../../new_pins.h"
 // LN882H header
 #include "hal/hal_gpio.h"
+#include "hal/hal_adv_timer.h"
+#include "hal/hal_clock.h"
+
+static int g_active_pwm = 0b0;
 
 typedef struct lnPinMapping_s {
 	const char *name;
 	uint32_t base;
 	gpio_pin_t pin;
+	int8_t pwm_cha;
 } lnPinMapping_t;
 
 lnPinMapping_t g_pins[] = {
-	{ "A0", GPIOA_BASE, GPIO_PIN_0 },
-	{ "A1", GPIOA_BASE, GPIO_PIN_1 },
-	{ "A2", GPIOA_BASE, GPIO_PIN_2 },
-	{ "A3", GPIOA_BASE, GPIO_PIN_3 },
-	{ "A4", GPIOA_BASE, GPIO_PIN_4 },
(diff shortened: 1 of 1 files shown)


openshwprojects:
thanks! SDK merged, can you update submodules and retry?

NonPIayerCharacter:
Maybe remove pins A13-B2 altogether? They are present only on QFN40, and used for QSPI. I believe they should not be used for gpio at all.

divadiow:
basic question. a decent way to test this would be RGBCW separate LEDs on 5 GPIOs on LN, set PWM1-5 as appropriate per GPIO in OBK, then fiddle with the colour picker/dimmer and web app RGBCW output tools?

divadiow:
it's probably my test but with these set currently only R, C and W light

RED A6 PWM1
GREEN A1 PWM2
BLUE A11 PWM3
COOL B8 PWM4
WARM B9 PWM5

thing is you can then re-assign blue to say B3 and it'll start working, but then warm or cool will stop. Individually all LEDs work. Could well me choosing wrong combination of pins, I know some are default pull-up, so the LED lights dimly when connected as a default state.

These are the options I have with MQ-12F module

image

as of right now with RCW working the warm/cool slider behaves as expected 

image
image
(green only dimly lit because A1 is default PU)

divadiow:
ok. with these set BEFORE A REBOOT

image

Red works
Green always lit
Blue works
Cool not work
Warm works

AFTER REBOOT

Red not work
Green works
Blue not work
Cool works
Warm works

every reboot after that is the same as second list above.

apologies if I'm confusing things and my test is rubbish.

NonPIayerCharacter:
I'm away for several days, so can't test it for now. Will try to connect an rgb led and play with channels, maybe figure out what's wrong. I have a LN-02 module, it should have enough free pins to test. Possibly problem is with ADV_TIMER_x_BASE, you can try to reduce amount of pwm channels in code, so each channel have separate TIMER_BASE.

divadiow:
cool. LN-02 options

image

I don't know why I haven't tried with a reduced number of PWMs set. I could just remove C+W to see if RGB are then fine. I will

> you can try to reduce amount of pwm channels in code, so each channel have separate TIMER_BASE.

I may do if I can work out what this means for the code!

divadiow:
interesting. with C+W unset G and B toggle OK. If you unset B too, leaving only R and G, only G will light BUT using the "100% WARM" button

divadiow:
with only R - PWM 1 - set the green is permanently lit but red does toggle

NonPIayerCharacter:
try https://github.com/openshwprojects/OpenBK7231T_App/pull/1331
max channels - 5

divadiow:
oh cool. will do. ChatGPT and I are playing too https://github.com/openshwprojects/OpenLN882H/pull/16/files

Issue #767, opened 2023-04-05
An error occurred while following the instructions to create a Docker build.


toneyche:
I am using Ubuntu 22.04 to build firmware with Docker. The first step was completed successfully with the command "docker build -t openbk_build --build-arg USERNAME=$USER .". However, I encountered an error when executing the second step. The error message is as follows:

root@DockerHost:/home/hatel/OpenBK7231T_App/docker# docker run -it -v "$(pwd)/..":/OpenBK7231T_App  openbk_build
\\/build_target_platforms.sh: line 3: shell: command not found
****************************************
****************************************

Building OpenBeken for the following platforms: OpenBK7231T OpenBK7231N OpenXR809 OpenBL602 OpenW800 OpenW600
Building OpenBeken with version "dev_20230405_204440"

****************************************
Building platform: OpenBK7231T
****************************************

make -C sdk/OpenBK7231T/platforms/bk7231t/bk7231t_os APP_BIN_NAME=OpenBK7231T_App USER_SW_VER=dev_20230405_204440 clean
make[1]: *** sdk/OpenBK7231T/platforms/bk7231t/bk7231t_os: No such file or directory.  Stop.
make: *** [Makefile:140: clean] Error 2
make APP_NAME=OpenBK7231T TARGET_PLATFORM=bk7231t SDK_PATH=sdk/OpenBK7231T APPS_BUILD_PATH=../bk7231t_os build-BK7231
make[1]: Entering directory '/OpenBK7231T_App'
git submodule update --init --recursive --remote
error: could not lock config file .git/config: Permission denied
error: could not lock config file .git/config: Permission denied
fatal: Failed to register url for submodule path 'sdk/OpenBK7231N'
make[1]: *** [Makefile:34: submodules] Error 128
make[1]: Leaving directory '/OpenBK7231T_App'
make: *** [Makefile:74: OpenBK7231T] Error 2
................

openshwprojects:
Hmm I haven't used docker build, it is very easy to use online github builds, they are working also in forks and in the pull requests.
But maybe the Docker support contributor can shed some light on to issue - @michaelkamprath  can you help?

michaelkamprath:
[...]

PR #1343, opened 2024-09-09
Add UID of user when building to avoid permission issue

memiks:
Hi,

When I tried to compile OpenBeken with docker locally, I had the same issue as #767 .

This was due to the UID of user used by the docker container. The UID was 1000 in docker and not the one of my Host user.

To correct this, I added the current UID as an argument to build the container image.


openshwprojects:
Thanks. I am not able to check it but I guess it won't harm. Merged.

PR #1346, opened 2024-09-12
Simple temperature graph on main page

MaxineMuster:
Store the last 100 internal CPU temperatures (if present) and display them in a simple SVG:

Graph_LN_BK_W800


divadiow:
cool 
image

lmk if you want it seen on any other CPU temperature supporting platforms

openshwprojects:
This could be merged if put in separate driver

openshwprojects:
I've checked PRs history and I think that LOTS of your @MaxineMuster PRs could be merged , but I would need your help, open topic on our forum, give us minimal working demo and include the changed functionality in self tests, and also if feature is not in obk core, just give it disabled by default in obk_config.h

PR #1369, opened 2024-09-29
Try authentication on W600 and W600

MaxineMuster:


divadiow:
I believe this is ready to merge?

MaxineMuster:
Two Points:
First: It has the limitation of 30 chars ("rest of the world" has 32).

Then: It worked o.k. for me on my W800, but made a "reset" of my config.
I think this is reasonalbe, since:
If I read the code correctly, W800 will use g_cfg->version=5.

So `"configSize = sizeof(mainConfig_t)"` will be used.

The password is added behind the last entry in `struct mainConfig_s` (for W600 and W800):
```
	//char initCommandLine[512];
#if PLATFORM_W600 || PLATFORM_W800
#define ALLOW_SSID2 0
#define ALLOW_WEB_PASSWORD 1
	char initCommandLine[512];
	// ofs 0x000007E0 (dec 2016)
	char webPassword[30];    // <<---- added this for PR
#else
#define ALLOW_SSID2 1
#define ALLOW_WEB_PASSWORD 1
	char initCommandLine[1568];
	// offset 0x00000C00 (3072 decimal)
	char wifi_ssid2[64];
	// offset 0x00000C40 (3136 decimal)
	char wifi_pass2[68];
	// offset 0x00000C84 (3204 decimal)
	char webPassword[33];
	// offset 0x00000CA5 (3237 decimal)
	char unused[347];
#endif
} mainConfig_t; 
```

Then for W800 (when there was no "unused" space defined after `char initCommandLine[512];` before this PR) configSize will be changed from this extension.

MaxineMuster:
No longer needed after pushing W600/W800 to config V4/V5

Issue #1376, opened 2024-10-03
Timestamps on charts showing wrong time


ehorvat1:
Charts are a great feature, thanks again.

Timestamps on x axis show incorrect time if NTP driver is used with an offset.
Since NTP driver delivers already correct unix timestamp values the drv_charts.c file could be changed.

Exchanging:
Chart_AddTime(g_chart, NTP_GetCurrentTime()); 
with:
Chart_AddTime(g_chart, NTP_GetCurrentTimeWithoutOffset());

in line 628 of drv_charts.c would be my proposed fix.

I will send a pull request for this .
Rg, Bert

Issue #1375, opened 2024-10-03
Brighter color for charts axis and ticks would improve readability


ehorvat1:
Charts are a great feature, thanks.

A brighter color for charts axis and ticks would improve readability
Currently the text/labels for axes and ticks are quite dark and therefore hard to read on a dark canvas.

This could be improved by adding : Chart.defaults.color = '#099' to the drv_charts.c file.
I will send a pull request for this .
Rg, Bert

Issue #99, opened 2022-06-27
Power monitoring supported?


TurkeyMan:
I have some DIN relays with power monitoring.
It doesn't look like power monitoring is supported yet. I have 3-4 (similar) different brand devices here I can flash, is there a way I can help test?
I need power monitoring in addition to relay control.

PeterPolacek:
For me is bk7231n from NAS-WR07W powerplug, thank you

openshwprojects:
I have this plug, it has BL0942 or BL0937 so it's supported I think.
![image](https://user-images.githubusercontent.com/85486843/175972690-e1325b83-73ce-486d-b312-001cfe9510d0.png)
![image](https://user-images.githubusercontent.com/85486843/175972772-b96bd80c-0991-4022-9436-36ee8be3aa89.png)

PeterPolacek:
Thank you, working now .. but voltage value is not correct .. i have 240V measured with volt meter in grid but plug shows me 257.9 .. how can i correct it? Thank you

PR #1377, opened 2024-10-03
Fix issue #1375 and #1376

ehorvat1:
Fix issue #1375  .....better visibility of text in charts by adding Chart.defaults.color = '#099'
Fix issue #1376 .....wrong time in charts when using NTP with an offset


openshwprojects:
I can merge it, but... you need to exclude obk_config.h changes

openshwprojects (approved):

ehorvat1:
OK. Reverted now all files which where unintentionally commited because I am a beginner in Github contributions. Now drv_charts.c only contains fixes  #1377  which is the only changed file remaining.. 
So just pull commit 31e5ea86106ae497f9d36b2fc725a172dbcbf360 and 47f76a41d13225dd37076f988f5dba1bbeaac7a6

openshwprojects:
Great, thank you

PR #1383, opened 2024-10-07
Update obk_config.h for custom build

jkwim:
#define ENABLE_DRIVER_CHARTS	1


jkwim:
You can ignore this pull request.

I already found that

Update obk_config.h #1374 - add charts driver to OBK for beken

will help me.

PR #1406, opened 2024-10-29
Update workflow to avoid deprecated commands

MaxineMuster:
Update workflow.yaml:
- change actions/up-download to v4
- use "environment" instead of "set-output"

see

https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

and

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -54,10 +54,10 @@ jobs:
           fi
           echo "new_release: ${{ steps.semantic.outputs.new_release_published }}"
           echo "version: ${VERSION}"
-          echo "::set-output name=new_release::${{ steps.semantic.outputs.new_release_published }}"
-          echo "::set-output name=short_ref::${BRANCHTRANSLATED}"
-          echo "::set-output name=sha_short::SHA-$(git rev-parse --short=12 HEAD)"
-          echo "::set-output name=version::${VERSION}"
+          echo "new_release=${{ steps.semantic.outputs.new_release_published }}" >> $GITHUB_OUTPUT
+          echo "short_ref=${BRANCHTRANSLATED}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


DeDaMrAzR:
@MaxineMuster 

There are more complications than it meets the eye in there :)

upload-artifact@v4 is implicitly creating ZIP archives for every artifact so there has to be a extract and then merge so we get the exact same output for publishing or we must change the releaserc.yaml to accommodate the changes. 

See some of my experiments here - https://github.com/DeDaMrAzR/OpenBK7231T_App/actions/runs/11516737538 

You can test there as much as you want so we can perfect the solution and then move it over to main.

DeDaMrAzR:
I've also updated the latest for simulator build so no warning are now present but I don't have enough time to finish that - nor knowledge 😊

MaxineMuster:
> upload-artifact@v4 is implicitly creating ZIP archives for every artifact so there has to be a extract and then merge so we get the exact same output for publishing or we must change the releaserc.yaml to accommodate the changes.

Yes, I'm aware of that, and I know this way we will get one zip file per platform during build, e.g. in PRs, instead of one single zip including all files. But to be honest, I think can be seen both an advantage and a disadvantage.

And I won't pretend I totally understand all this, but as far as I understood, there is no need to change releaserc.yaml, for download-artifact seems to be aware of that.

So, as I said, only tested once, a release seems to work exactly as before:

https://github.com/Max4OBK/OpenBK7231T_App/releases/tag/1.0.1

MaxineMuster:
You're welcome ;-)  ;-)!
Sorry for doing so, it was just curiosity, and I learned something new again, I would never have looked at before...

Thanks for bringing this into a working state!

PR #1407, opened 2024-10-29
Update workflow.yaml

DeDaMrAzR:
Final version of workflow file with deprecated actions and calls updated and fixed - thanks to Max for hints and inspiration!!!

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -6,7 +6,7 @@ on:
   pull_request:
     branches: 
       - '**'
-
+      
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


Max4OBK:
Would you support my request to keep the platform zips?

workflow_keep_platforms

Just don't delete the files (`delete-merged: false`).

I really would appreciate to be able to download "my" platforms file instead of the complete zip with all files.

I tried this here: https://github.com/Max4OBK/OpenBK7231T_App/releases/tag/1.0.5
and the release page looks o.k. (same as before).

Additionally even without a "release" you can select which platform you want to download - beside the full zip with all files as before, as e.g. in this branch:
https://github.com/Max4OBK/OpenBK7231T_App/actions/runs/11589617160
grafik

PR #1414, opened 2024-11-04
LN882H WS2811 Driver

jasperro:
This PR is related to https://www.elektroda.com/rtvforum/topic4083817.html

It implements WS2811 support for the LN882H platform.

I will still be adding some documentation and the ability to change the data pin using a pin role, after which I will unmark this as draft.


jasperro:
I think this one is about ready. Before merging, it is needed that https://github.com/openshwprojects/OpenLN882H/pull/19 is merged. The .gitmodules file and modules can then point to the main OpenLN882H repo again.

jasperro:
> Nice, WS2811 on LN88H, altough it would be better to have it togeter with current SPI LED for Beken, instead as separate driver. Btw, I don't think you can add pin enumeration inside enum without breaking people configs... also , why pins header change?

The pin header change was because LN882H pins are weird. They have a base and then the pin. By splitting out the pin definitions into a separate header file, I could use this to call LN882H hal functions (which require base and pin) with OpenBK7231T_App pin index.
I'm not sure if this is the best way, but it is the one I could make work.

The reason it does not work like the BK SPIDMA/SPILED is because LN882H has direct HAL for WS2811 in the SDK. This way, I could just easily use the already existing code. I think the LN882H has some built-in WS2811 circuitry that doesn't use SPI at all, only DMA. So it would not make sense to use SPI if this is actually not supposed to be used.

Or do you mean something else? I think it's a good idea is to have a header file for SM16703P-like drivers, and then implement them based on the platform?

So we would have a drv_SM16703P.h, that is implemented by (for example) drv_SM16703P_LN882H.c and drv_SM16703P_BK7231.c. Then the commands would be the same on both platforms, and would make it possible to use PixelAnim without any more work.

Also, with "Btw, I don't think you can add pin enumeration inside enum without breaking people configs...", do you mean you store the enum backing int in the config instead of the string? So if I add it in the middle it does not work for existing configs? (but in the end shouldn't break it)

It would be very nice to get this merged eventually.

NonPIayerCharacter:
Alternative using SPI + DMA is available since 1.18.151
Any pin that is not internal flash is allowed (flash - A13 to B2 inclusive).

jasperro:
It works great on my LED strip! I just flashed version 1.18.175. All effects working as well (including the new theater ones). Closing the PR.

PR #1429, opened 2024-11-16
Add commands to check if actual time is in DST or not

MaxineMuster:
New command `clock_calcDST nthWeekEnd monthEnd dayEnd hourEnd nthWeekStart monthStart dayStart hourStart `
to set the DST start and end rule.

Examples:
 	Australia / Sydney		`CLOCK_calcDST 1 4 1 3 1 10 1 2`
		Amsterdam (EU)			`CLOCK_calcDST 0 10 1 3 0 3 1 2`

`$isDST` will return 1 if DST is active, 0 else

It's enabled with

`#define ENABLE_NTP_DST                  1`


giedriuslt:
careful with this, I see it's already to big for bl602, probably because of strftime use

MaxineMuster:
> probably because of strftime use

That could be easily solved, it's only used for informational purposes.
It's time my BL602 arrives, so I can do my own tests...

openshwprojects:
I think I can merge it, but can you disable it by default, since it seems to raise flash uasge a bit?

MaxineMuster:
> I think I can merge it, but can you disable it by default, since it seems to raise flash uasge a bit?

@openshwprojects:  
Switched to "ctime()" for logging to save some space and disabled feature for all platforms but Windows, since I also added testing code ...

MaxineMuster:
> careful with this, I see it's already to big for bl602, probably because of strftime use

Should be better now @giedriuslt, tried to remove all additional functions from time.h
Does it fit for BL602 now? It's down to approximately 1.5k additional file size now.

giedriuslt:
Should be fine now. Firmware partition is 800kb in bl602 with devcube 1.4.8, we are quite close to that.

openshwprojects (approved):

PR #1435, opened 2024-11-22
Uart2

divadiow:


xjikka:
I'm working on "upgrading" the uart to be able to use both uarts at the same time (separate buffers and functions). The changes are in drv_uart.c and drv_uart.h (backwards compatibility is maintained). I'm already in the testing phase, but it will take longer.
Aren't you doing something major on the uart now?

divadiow:
ah most of the above is removal of some customisations from where I branched this new PR. the latest commits were me trying to add your code to make a build. I didn't complete. I'll close PR.

PR #1437, opened 2024-11-23
Uart2bufmod

xjikka:
I added new routines to enable simultaneous use of both UARTs on Beken. Each UART has its own Ring Buffer. It is allocated only when used. The new routines use the uartindex parameter - specifying the port and also the buffer.

The goal is to use a measuring cube with two BL0942, each on one UART /I am working on the modification/.

Backward compatibility with UART switching in the config using the OBK_FLAG_USE_SECONDARY_UART flag has been preserved.

Currently only for the Beken platform, but it should not be a problem to add other platforms.


MaxineMuster:
Just for I'm curious: Did you take into account, that actually the second UART is used for device logging? 
Did you maybe already change this?

xjikka:
> Just for I'm curious: Did you take into account, that actually the second UART is used for device logging? Did you maybe already change this?
> 
> Oh, sorry, I just now found the Elektroda thread about all this. Forget my comment ...

Of course, for using UART2, it is good to first turn off the log by using "logtype none" in autoexec.

Anyway /and this surprised me/ it can also work in parallel. Ie. log on UART2 + BL0942. Only with the fact that the log is also at 4800 baud which slows down the system, but it works fine. BL0942 ignores the other data on port. 

But I recommend turning off the log when using it.

PR #1443, opened 2024-11-27
Uart2bufmod

xjikka:
I added new routines to enable simultaneous use of both UARTs on Beken. Each UART has its own Ring Buffer. It is allocated only when used. The new routines use the uartindex parameter - specifying the port and also the buffer.

The goal is to use a measuring cube with two BL0942, each on one UART /I am working on the modification/.

Backward compatibility with UART switching in the config using the OBK_FLAG_USE_SECONDARY_UART flag has been preserved.

Currently only for the Beken platform, but it should not be a problem to add other platforms.

20241127 merged changes from main

### DIFF
--- a/src/driver/drv_pwrCal.c
+++ b/src/driver/drv_pwrCal.c
@@ -1,5 +1,6 @@
 #include "drv_pwrCal.h"
 
+#include <math.h>
 #include "../cmnds/cmd_public.h"
 #include "../logging/logging.h"
 #include "../new_cfg.h"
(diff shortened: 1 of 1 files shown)


xjikka:
@openshwprojects Can you please take a look to this PR before there are a lot of other changes on main so I don't have to resync source again? I'm running the whole switchboard on this PR and it's OK. There are no breaking changes, just separate buffers for both UARTs. Thanks.

PR #1448, opened 2024-12-01
SSID retain with GUI

MaxineMuster:
try for GUI setting of the new retain SSID feature from XJIKKA

See discussion here:
https://www.elektroda.com/rtvforum/topic4088302.html


MaxineMuster:
Bildschirmfoto vom 2024-12-01 18-40-27

PR #1451, opened 2024-12-02
bk7238 arduino sdk

NonPIayerCharacter:
There is now a normal sdk, so i will just upload this branch on arduino sdk for testing purposes.


NonPIayerCharacter:
Better alternative in https://github.com/openshwprojects/OpenBK7231T_App/pull/1527

PR #1462, opened 2024-12-12
Move some functions to hal, update workflow

NonPIayerCharacter:
Moved delay us and wdt init/reset to HAL (will move uart later, likely in next port).
Added generic platform, with empty function weakly linked for easier initial porting.
Adjusted makefile, so that now only SDK required for build will be updated (cuts some time for workflows).
Fix workflow deprecation warnings.
Workflow for pull request will now generate an archive for each platform, instead of a single one with all platforms inside. Still, merge is performed for compatibility with release script.


MaxineMuster:
Can't test, but that's a great idea to consolidate these functions to HAL!
Thanks so much for your engagement for this project!

divadiow:
nice. maybe one day I'll make something of the LN8825x SDK https://www.elektroda.com/rtvforum/viewtopic.php?p=21344461#21344461

not that anyone has a device with that chip in

NonPIayerCharacter:
Continues in #1468, additionally moved uart

PR #1463, opened 2024-12-13
Added 48-NEC1 support to the Arduino-IRremote-mod

mintos5:
Hi, just wanted to share my workaround code for sending IR codes in 48-NEC1 protocol:
http://www.hifi-remote.com/johnsfine/DecodeIR.html#48-NEC1

The  Arduino-IRremote-mod doesn't support it natively, so I patched it in. The library supports sending RAW or pronto codes, but I think OpenBeken has a limited payload, so it will not fit.
The result is this example command:
**IRSend NEC-1-0303-0-30**
you need to specify **bit length!** (in hex: 0x30=48 bits)
and the 16bit aCommand is has 2 fields:
- command/function in the highByte (mask: 0xFF00)
- E value in lowByte (mask: 0x00FF)

I tested it with my FAN controllable with IR remote, it works nicely


openshwprojects:
Nice addition, thanks

openshwprojects (approved):

PR #1467, opened 2024-12-15
enable DHT driver on TR6260

divadiow:


NonPIayerCharacter:
It won't work yet, delay is implemented only in ds driver. Wait until rtl pull is merged, then it shall work

divadiow:
ah ok ok. however..

image

Issue #1227, opened 2024-05-12
W800 OTA not implemented/does not work


divadiow:
tried UART flashed W800 to 1.16.7, 1.17.358 and 1.17.571 to see if OTA has ever been implemented. 

log out just says this. web app and gui menu OTA option tried.

```
Debug:API:POST to api/ota
Debug:API:No OTA
```

DeDaMrAzR:
Just got one of these devices and just wanted to confirm it is not working.

divadiow:
GA release. Closing

PR #1471, opened 2024-12-19
OTA for W800 + "Simple OTA" w/o httpd

MaxineMuster:
Removed unnecessary (and wasting resources) fetching of "index?status=1" on pages not using "state" div 
Added simple "internal OTA" (just POST-ing a file) 

Basic tests done - try avoiding other TCP connections to save memory 
Sometimes there are memory related messages in log like "Thread create HTTP Client - errCOULD_NOT_ALLOCATE_REQUIRED_MEMORY".

W800s OTA working in most cases, sometimes device crashes during update. 
Device will immediately reboot after OTA completion, even printfs after writing last bytes to flash was not visible in console.


divadiow:
This PR fixes #1227

src/httpserver/http_fns.c
```diff
@@ -3016,7 +3016,42 @@ int http_fn_ota(http_request_t* request) {
...
+#if PLATFORM_BL602
+"accept='.bin.xz.ota'"
+#elif PLATFORM_LN882H
+"accept='.bin'"
+#elif PLATFORM_ESPIDF || PLATFORM_W600 || PLATFORM_W800
+"accept='.img'"
+#elif PLATFORM_BEKEN
+"accept='.rbl'"
+#else
+""
```
NonPIayerCharacter, line 3038: Maybe use .img as default for everything else, and add/define something like NO_PLATFORM_OTA in obk_config.h for those that doesn't support it.

MaxineMuster: Thanks for your suggestions! I especially like the "NO_PLATFORM_OTA" idea.

Regarding the default: Since I changed the code again, I think it's clearer/better readable if I put the entries per platform together, even if e.g. ".img" is repeated for some platforms.
Since I took your idea with the new define, there should be only "known" platforms, so a default value should cover no devices.

openshwprojects (approved):

PR #1474, opened 2024-12-21
ADS1115

openshwprojects:


DeDaMrAzR:
All test are successful, ADS resolution is 0.0001875V and active inputs should be pulled low with a 2k2 resistor to avoid noise, non active pins can be floating. 

test autoexec:

`// set SoftSDA and SoftSCL pins
startDriver I2C

//if you are adding another device pull addr pin of it to GND and it will get 0x49 address
//usage is **addI2CDevice_ADS1115 Soft <device address> <CH for A0> <CH for A1> <CH for A2> <CH for A3>**
//if you want to NOT read a channel append 0xff for that input as shown bellow where
//only first inputs are read on both devices
addI2CDevice_ADS1115 Soft 0x48 1 0xff 0xff 0xff
addI2CDevice_ADS1115 Soft 0x49 5 0xff 0xff 0xff

//publishing a channel as with voltage conversion
//addEventHandler OnChannelChange 1 publishFloat Voltage0 $CH1*0.000188472`

PR #1481, opened 2024-12-28
feat(httpserver): improved mimetype handling for js, css

klevytskyi:
I've slightly edited some lfs http handlers so that it will correctly respond with text/css mimetype for .css files.


klevytskyi:
Well had some caching issues initially, apparantly everything works fine. Here are the screenshots that depict correct mime types for content-type header of CSS files.

image

openshwprojects:
Ok I think this is acceptable. Thanks

PR #1483, opened 2024-12-29
Split ntp and clock code

MaxineMuster:
Move clock related commands from drv_ntp to drv_deviceclock if they are not bound to NTP

Fixed some misplaced code inside cmd_if.c (ntp and BL0937 code was in wrong "ifdefs") 
Code from timed_events.c : Switch from "strcasestr" to "wal_stricmp" to compile e.g. on W800

Enabled NTP and DST for all platforms to allow testing


MaxineMuster:
No longer needed

PR #1322, opened 2024-08-19
Try to trace where watchdog reboots on bl602

giedriuslt:
Traces all the main loop to serial console to try to catch why bl602 reboots

PR #1487, opened 2025-01-01
Update BL602 code to work with latest sdk.

giedriuslt:
This updates code so that BL602 works with latest sdk.
This has following advantages

- It now can connect to WP2/WPA3 and mixed networks
- Powersave now works (even with bl0937)

However,

- It's bigger, I worked hard to remove unnecessary stuff, but still ends up 4 kb bigger than before
- ~seems new sdk does not like github and does not build in there.~

Tested:
- OTA
- Basic functions
- BL0937
- Mqtt

Not tested:
- Everything else


giedriuslt:
Now it builds and looks working

giedriuslt:
Backported igmp and mqtt password fix, changed hostname to be OpenBL602_xxxxxxxx. Did some more cleanup.
Testing, hopefully it's good to merge

ataa:
Thanks for this.
So the hostname is not hardcoded in this version and can be changed?

giedriuslt:
No, it's still hardcoded unfortunately, I just changed it to match the general pattern

giedriuslt:
Seems to be rebooting every 1-2 days for me. Need to investigate more...

r0bb10:
i can confirm it connects to my wpa2/3 wifi and works now

MaxineMuster:
Since you mentioned the size is even bigger than the "old" sdk, I recently found that disabling TLS will reduce the size about 100k :

-rwxr-xr-x 1 max max 793244 Jan 14 20:03 OpenBL602_dev_20250114_190155.bin
-rwxr-xr-x 1 max max 689300 Jan 14 20:06 OpenBL602_dev_20250114_190617.bin


The diff is simple:
```
diff --git a/components/network/lwip/lwip-port/config/lwipopts.h b/components/network/lwip/lwip-port/config/lwipopts.h
index 13b7546c..aa783a8e 100644
--- a/components/network/lwip/lwip-port/config/lwipopts.h
+++ b/components/network/lwip/lwip-port/config/lwipopts.h
@@ -26,8 +26,8 @@
 
 #define LWIP_NUM_NETIF_CLIENT_DATA      1
 
-#define LWIP_ALTCP                      1
-#define LWIP_ALTCP_TLS                  1
+//#define LWIP_ALTCP                      1
+//#define LWIP_ALTCP_TLS                  1
 /**
  * NO_SYS==1: Provides VERY minimal functionality. Otherwise,
  * use lwIP facilities.
```

giedriuslt:
Yes disabling tls would be useful, we could enable more stuff for bl602. SendGet does not support tls even on bk7231. Unless we need that for mqtt? but probably there is no support for that either.

MaxineMuster:
I don't recall any usage of TLS in OpenBeken, so it should be safe to disable it. We might disable some more features/code to save some more space, but that will be harder in changing makefiles

giedriuslt:
I fixed the version string, and changed temp reading to use a new way from new sdk, which does not use wait loops. Now it's stable for 3 days already. So I think it's good yo merge now

danielschwab:
Thanks for your effort. Just want to ask about the state of this PR? Will this be merged soon?

giedriuslt:
I do not have rights to merge, so I don't know if this would be merged.

danielschwab:
@openshwprojects What do we need to merge this PR?

SaschaKP:
Latest update was excellent, my current device (BL602, dry contact switch by mhcozy) that is really lazy in connecting to wifi (about 30 minutes before connecting), now connects in less than 3 minutes.
I have also merged changes from https://github.com/openshwprojects/OpenBK7231T_App/pull/1322 to the SDK, since temp reading seems a bit lazy also.

openshwprojects:
@danielschwab @SaschaKP @giedriuslt **I can merge it even today**, but can you please double-check and give me final confirmation that it's stable and won't break stuff in release? I'm today short on time and my weekend went mostly into testing XR872 camera so I'm not sure if I manage to test myself.  Also, after I merge, can you keep an eye on your BL602s with new SDK and report any potential problems?

danielschwab:
I just flash the version of this PR yesterday and will test it for stability. I also have 4 more devices which I am going to flash and test. As for now it's seems to work stable for the last 27 hours. I don't have any connection issues (it takes like 10 sec. to connect to the ap) and mixed WPA2/WPA3 is working just fine.

We should also decide if this PR should disable TLS (see above: https://github.com/openshwprojects/OpenBK7231T_App/pull/1487#issuecomment-2590941164) to save some memory I guess. But I quite new here so I don't know if this is a good idea or not. I think @MaxineMuster and @giedriuslt can maybe help here.

@giedriuslt can certainly give us a testimonial about this PR and stability!?

If help is needed I am open to support where I can!

giedriuslt:
I tested stuff as I wrote, but I don't have any lights for example so can't say if they work. I have seen people reporting lights are working though.  Ota is working.

For stability I get a watchdog reboot every 3 to 10 days, and it's hard to trace, because module connected to the computer does not reboot. But the old one was not stable for some people either, as evident here #1322. I have never observed this myself, but with the new sdk I use different way to read temp.

openshwprojects:
Thank you for input! So, what do we do @danielschwab  , let's say, wait 2 days more and if no serious issues on your side, then I will merge?

Regarding TLS and 1MB chips support ( here https://github.com/openshwprojects/OpenBK7231T_App/pull/1534 )
@giedriuslt can you help us integrate this into separate build permutations, so we can include all 3 types of binaries in the release? 
Please take a look here: https://github.com/openshwprojects/OpenBK7231T_App/blob/main/.github/workflows/workflow.yaml Here's what Is did for Beken:
image
I think the same could be done for BL602, depending on what's needed.

Regarding disabling features - well, it all comes to the OTA file size. I remember we already had issue where OTA file was too big and OTA was failing every time. So we're trying to avoid that.

SaschaKP:
@openshwprojects I will test the device just as it is now, with all the functionality necessary for it to work as normal and will post a confirmation (or not) for it's stability in a pair of days in the morning (GMT+1), just to keep an eye on the serial debug. If there isn't any problem it's okay for me. I've just seen improvements in this PR vs my BL602 device, plus the fact that I can use the WPA3 also and I don't have to wait an infinite time for it to connect. Mileage for others may vary, but just to keep sure that it won't broke things for others, being cautious is the best way.

danielschwab:
> Thank you for input! So, what do we do @danielschwab , let's say, wait 2 days more and if no serious issues on your side, then I will merge?

Sounds good to me. I will provide my feedback in a couple of days.

EagleonePrimo:
I just install this PR today I will test it with my relay (SM-028_V1.3) so for now it connect and work fine on WPA2-PSK/WPA3-SAE Mixed Mode with 802.11w Management Frame Protection -> optional

I'm here for just confirm it work

SaschaKP:
I can confirm it work on my device, 3 days uptime, and works with WPA3/WPA2 networks (WPA2 connection only).
It's currently stable on my device.

openshwprojects:
Ok, @danielschwab  ? I think we can emrge it today?

danielschwab:
Sorry I wanted to test as long as possible and all my devices were running without any issues. One for 6 days. So I think we can merge!

openshwprojects (approved):

divadiow:
@giedriuslt 

> Powersave now works (even with bl0937)

please confirm this is still the case. ref: https://www.elektroda.com/rtvforum/viewtopic.php?p=21537185#21537185

giedriuslt:
Sorry, I never tested this scenario, because I never use powersave. I measured current with it and it works, but never tried if it's fine after restart.
Here is work around #1628

SaschaKP:
Just as a side note for anyone noticing weird connections on wifi, I've noticed that if the AP is too nearby the device won't connect (less than 50cm). (image as follows, wifi chip is like this, bl602) 
image

Elektroda post:
https://www.elektroda.com/rtvforum/viewtopic.php?p=21539067#21539067

I'm leaving this just as a side note for anyone

PR #1488, opened 2025-01-02
Various fixes for callback list management.

notzed:
I noticed some messy and broken logic in new_mttq.c by chance and offer some fixes and minor improvements.  As is removeCallback can't function since ID is never set in addCallback, and a malloc failure for the subscriptionTopic will cause a use-after-free since topic is only freed and not NULLified.  To simplify the array traversal this change also ensures that all numCallback entries are valid by compacting the array in removeCallback.

As per the commit log:

 - use-after-free in the case of a failed malloc
 - registercallback never sets ID, so removecallback can't work
 - clearcallbacks doesn't reset numCallbacks
 - redundant NULL checks for os_free
 - change to avoid empty slots in list by copying the last entry to the newly empty slot in removecallback

### DIFF
--- a/src/cmnds/cmd_main.c
+++ b/src/cmnds/cmd_main.c
@@ -21,6 +21,7 @@ int cmd_uartInitIndex = 0;
 #elif PLATFORM_LN882H
 #include <wifi.h>
 #include <power_mgmt/ln_pm.h>
+#include <system_parameter.h>
 #elif PLATFORM_ESPIDF
 #include "esp_wifi.h"
 #include "esp_pm.h"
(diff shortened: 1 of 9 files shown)


openshwprojects:
Hey, are you sure about this part?
image
postrstr is not a va function, it is not doing sprintf, so no formatting there. Are you sure that it needs escaping?

notzed:
Yeah that's correct.  It's a quoted string, and \s is not a valid escape sequence in C quoted strings - "undefined behaviour" - in this case gcc still includes \ in the output but issues a warning.  These quoted values are processed by the compiler not by printf etc.

PR #1496, opened 2025-01-06
http_client: fix sending HTTP POST without content

blocktrron:
Invoking HTTP POST without a request-body results in a request which incorrectly has content-type and content-length set.

This results from a missing check if a body is provided as an argument.

Add this missing check in order to allow POST requests without a message body to work.


openshwprojects:
Seems ok I guess, thx

openshwprojects (approved):

PR #1497, opened 2025-01-06
driver: tuyaMCU: introduce battery-powered device ack-delay

blocktrron:
This introduces a delay for sending the ACKs on battery powered TuyaMCU sensors.

Rationale is to allow users to delay the return to sleep state. This is useful when e.g. changing settings on the device.

Currently the device is sent back to sleep after the status has been read and commited to MQTT. This can be less than 5 seconds of uptime.

Set this value to a sane default, which allows the device to retain online for enough time for a user to change configuration.

### DIFF
--- a/src/driver/drv_tuyaMCU.c
+++ b/src/driver/drv_tuyaMCU.c
@@ -211,6 +211,7 @@ static int wifi_state_timer = 0;
 static bool self_processing_mode = true;
 static bool state_updated = false;
 static int g_sendQueryStatePackets = 0;
+static int g_tuyaMCUBatteryAckDelay = 20;
 
 // wifistate to send when not online
 // See: https://imgur.com/a/mEfhfiA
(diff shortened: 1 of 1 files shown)


openshwprojects:
While the generic idea is good, I am not sure if it should be 20 by default. People will not notice it and it will drain battery fast.

Maybe change default value to 0 and also add code to change this value to 20 when a pairing button is pressed, or something like that? I am not sure if I remember correctly, but we had some kind of pair message from tuya battery device received...

blocktrron:
You seem to be right. I'll look into this and ping you once i have something ready.

blocktrron:
@openshwprojects This is ready for a second review.

One thing to add - This is also necessary when submitting states using the HTTP client. Otherwise the state machine sends the device to sleep before the HTTP request is performed.

openshwprojects (approved):

openshwprojects:
It seems acceptable. Sorry for the delay

io2345:
Can you tell me how to use the delay and how to set a desired value?

PR #1501, opened 2025-01-07
RTL8710B/RTL8711AM

NonPIayerCharacter:


DeDaMrAzR:
It will be 2-3 weeks until I receive more RTL modules but so far T102 module is working just fine. I try to insert it into a working device and do some real life tests with it. 

image

What is left for this to be merged?

DeDaMrAzR:
Static IP works

image

But OTA is weird, it will work from FW1 to FW2 but not the other way around.

PR #1513, opened 2025-01-16
Fix calibration

### LINKED ISSUE #1502: Calibrating power metering with non-integers results in NaN power meter reading
Not looked at this extensively but I have been able to replicate an issue a user reported here https://www.elektroda.com/rtvforum/topic4096322.html

image

apply a non-integer value to power and readings on main GUI will be littered with NaN eg

image
setting back to integer does not fix

giedriuslt:
You can set 0 or NAN as calibration constant currently, which should not be allowed, and in my case bricks the device.
Should be validated correctly. 
Should fix #1502

### DIFF
--- a/src/driver/drv_pwrCal.c
+++ b/src/driver/drv_pwrCal.c
@@ -1,5 +1,6 @@
 #include "drv_pwrCal.h"
 
+#include <math.h>
 #include "../cmnds/cmd_public.h"
 #include "../logging/logging.h"
 #include "../new_cfg.h"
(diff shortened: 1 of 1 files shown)


divadiow:
I deliberately messed up my BL0937 device again on GA then flashed to 1513_merge_7263f06183e7 - the NaNs remained through counter clears. Also, applying a power value, integer or non, gave the below. voltage and current are OK. Please sanity check me - can you replicate?
image

divadiow:
if your modifications were only designed to stop it happening at all, and not fixing if already happened, then no worries.

openshwprojects:
How does it brick a device? I was aware about the issue but I haven't seen any way it can brick stuff

giedriuslt:
> if your modifications were only designed to stop it happening at all, and not fixing if already happened, then no worries.

This fix does nothing to correct wrong data, if you have NaN in there might be hard to get out, but recalibrating and reseting energy should I believe work

giedriuslt:
> How does it brick a device? I was aware about the issue but I haven't seen any way it can brick stuff

I tested with bl602, it might have different NaN/division by zero handling. The issue was after I broke calibration, web page no longer loaded. 
I tried to put it in safe mode, I could connect to device wifi, but but it still doesn't load for some reason. Although in this case it should I believe, maybe I did something wrong. 
Reflashing chip which full erase (which obviously removes any config) fixed the problem

openshwprojects (approved):

PR #1521, opened 2025-01-23
RTL87X0C SSDP enable

divadiow:


divadiow:
image
image


interesting. no duplicate in network places as seen with BK

PR #1525, opened 2025-01-30
Update hal_pins_tr6260.c

morganflint:
To check if GPIO 13 works as the rest. Elejktroda user insmod suggested this change when I explained to him that the original FW in my device used this GPIO, but insmod had excluded it in the code because he had taken it from the SDK.


morganflint:
Closing as divadiow did the same one to my request, as I couldn't get the green tick needed to launch the online building system

Issue #1523, opened 2025-01-26
BL602: latest release 1.18.24 not starting


Ga-Ol-St:
Build 1.18.24 not starting, Build 1.18.23 working fine.

**Firmware:**
- Version: 1.18.24
- Device: Scimagic-RC
- Chip/model: BL602
- Device config:
`{
  "vendor": "Tuya",
  "bDetailed": "0",
  "name": "Full Device Name Here",
  "model": "enter short model name here",
  "chip": "BL602",
  "board": "TODO",
  "flags": "1074266116",
  "keywords": [
    "TODO",
    "TODO",
    "TODO"
  ],
  "pins": {
    "12": "DS1820_IO;0",
    "14": "Rel;49",
    "17": "WifiLED_n;46",
    "21": "Btn_Tgl_All;49"
  },
  "command": "backlog startDriver DS1820;",
  "image": "https://obrazki.elektroda.pl/YOUR_IMAGE.jpg",
  "wiki": "https://www.elektroda.com/rtvforum/topic_YOUR_TOPIC.html"
}`

**To Reproduce**
1. update firmware (OTA or by UART) to 1.18.24
4. See error during boot

[error.log.txt](https://github.com/user-attachments/files/18550319/error.log.txt)

openshwprojects:
So this commit must be at fault:
https://github.com/openshwprojects/OpenBK7231T_App/commit/77efc7ebcaa2f9c42c6f20e3a332d711f161d499
Can you help us investigate - flash older version, remove startup command line and then retry newer version? Does new version still crash wih empty command line?
@giedriuslt do you have BL602 and can check?

openshwprojects:
Maybe stack size is the issue.

giedriuslt:
I see the same issue, removing command line in safe mode makes it work back.

openshwprojects:
Can you check is it a stack size issue?

openshwprojects:
Temporary fix (revert to single line execution):
https://github.com/openshwprojects/OpenBK7231T_App/commit/bc35f97f1a5420bd4bcaa8959abee235495afa4c

divadiow:
with 1.18.24 on BL602 mine stops here

```
Main_Init_Delay done
Current task sp data:
RA:23093b08, mstatus:80007880
A0:00000128 A1:00000000 A2:00000000 A3:00000000 A4:4201d820 A5:00001000 A6:00001000 A7:000000d6
[...]

PR #1529, opened 2025-02-02
Fix multi-line startup command for BL602

giedriuslt:
This fixes multi-line command for BL602 issue #1523 . Problem is that realloc is not working on BL602. So by moving malloc for g_scrbuffer to the function where it's used realloc is not required and so everything works. (except it would fail if single script line is more than 256 chars long, but that problem exists already and should not be stopping this fix)

I think realloc is broken on BL602 because malloc and free is redefined in aos_freertos.c file, while realloc is not, so the one libc is used and obviously you can't mix those two. I need to check maybe new sdk fixed realloc.

### DIFF
--- a/src/cmnds/cmd_script.c
+++ b/src/cmnds/cmd_script.c
@@ -246,7 +246,7 @@ typedef struct scriptInstance_s {
 } scriptInstance_t;
 
 int g_scrBufferSize = 0;
-char *g_scrBuffer = 0;
+char *g_scrBuffer = NULL;
 int svm_deltaMS;
 scriptFile_t *g_scriptFiles = 0;
 scriptInstance_t *g_scriptThreads = 0;
@@ -365,6 +365,11 @@ void SVM_RunThread(scriptInstance_t *t, int maxLoops) {
 	int loop = 0;
 	const char *start, *end;
 	int len, p;
+	
+	if(g_scrBuffer == NULL) {
+		g_scrBufferSize = 256;
+		g_scrBuffer = malloc(g_scrBufferSize + 1);
+	}
 
 	while(1) {
 		loop++;
@@ -432,10 +437,6 @@ void SVM_RunThreads(int deltaMS) {
 	c_run = 0;
 	svm_deltaMS = deltaMS;
 
-	if(g_scrBuffer == 0) {
-		g_scrBufferSize = 256;
-		g_scrBuffer = malloc(g_scrBufferSize + 1);
-	}
 
 	g_activeThread = g_scriptThreads;
 	while(g_activeThread) {
--- a/src/user_main.c
+++ b/src/user_main.c
@@ -1120,11 +1120,7 @@ void Main_Init_AfterDelay_Unsafe(bool bStartAutoRunScripts) {
 
 		// NOTE: this will try to read autoexec.bat,
 		// so ALL commands expected in autoexec.bat should have been registered by now...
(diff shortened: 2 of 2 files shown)


openshwprojects:
I can merge it but it would be nice to have realloc fixed instead of work around

giedriuslt:
realloc is working with updated sdk version here #1487

MaxineMuster:
> (except it would fail if single script line is more than 256 chars long, but that problem exists already and should not be stopping this fix)

I just struggled with this, and at least in my case it seems not true:

I had a quite long startup command, and after update the device was not booting anymore.
I blamed my code (firmware was 795k, so I thought I overwrote something), removed TLS again to reduce size, but it kept crashing ...

```
Main_Init_Before_Delay done

Main_Init_Delay

Main_Init_Delay done
Current task sp data:
RA:23077cfc, mstatus:80007880
A0:4200e440 A1:42026fd8 A2:4200e9d4 A3:4200e9d4 A4:00000210 A5:00000000 A6:00000000 A7:00000000
T0:23077ba0 T1:00000000 T2:ffffffff T3:00000001 T4:0000000a T5:42467830 T6:20353378
S0:00000210 S1:80000110 S2:42026fd0 S3:0000020b S4:c20270e0 S5:4200e440 S6:42026fd8 S7:00000017
S8:0000003a S9:4200f000 S10:42026e98 S11:a5a5a5a5
Exception Entry--->>>
mcause 38000005, mepc 23077d38, mtval c20270e4
Exception code: 5
  msg: Load access fault

```

... until I finally recalled this thread.

Device was doing fine again after I switched back to single line startup command:
```
diff --git a/src/user_main.c b/src/user_main.c
index 4ab4d0d3..7d76ebfa 100644
--- a/src/user_main.c
+++ b/src/user_main.c
@@ -1119,7 +1119,10 @@ void Main_Init_AfterDelay_Unsafe(bool bStartAutoRunScripts) {
 
                // NOTE: this will try to read autoexec.bat,
                // so ALL commands expected in autoexec.bat should have been registered by now...
-#if ENABLE_OBK_SCRIPTING
+#if PLATFORM_BL602
+               // temporary fix
+               CMD_ExecuteCommand(CFG_GetShortStartupCommand(), COMMAND_FLAG_SOURCE_SCRIPT);
+#elif ENABLE_OBK_SCRIPTING
                SVM_RunStartupCommandAsScript();
 #else
                CMD_ExecuteCommand(CFG_GetShortStartupCommand(), COMMAND_FLAG_SOURCE_SCRIPT);

```


Just for reference, the failing command is:

`backlog startDriver DS18B20 ; DS18B20_setsensor "0x28 0x61 0x64 0x35 0xFB 0x2C 0x22 0x72" "master bedroom" "3" ; DS18B20_setsensor "0x28 0x61 0x64 0x35 0xFB 0x5C 0x56 0x5D" "kitchen" "2" ; DS18B20_setsensor "0x28 0x61 0x64 0x35 0xFB 0x57 0x59 0x3F" "living room" "4" `

which is 267 chars.


Maybe I can try new SDK soon, but I will call it a day now.

giedriuslt:
Yes, I did not think about such scenario. The bl602 will crash if file has line longer than 256 chars because realloc is not implemented. However, old way to run startup command is not re/allocating anything, so it worked.

But here you should just make it multiline instead of using backlog, given that's now supported.

PR #1533, opened 2025-02-14
Add support for PowerSave 2 command to enable RF only sleep on Beken N/T

Zanadar:


divadiow:
maybe submit a post about this too for greater visibility?

https://www.elektroda.com/rtvforum/forum390.html

NonPIayerCharacter:
Remember making this suggestion on elektroda, but i forgot about it.
Btw, comment should be adjusted here: https://github.com/openshwprojects/OpenBK7231T_App/blob/82ec9c8ed63300627e9ceb967814bf2820899a68/src/cmnds/cmd_main.c#L842
And not in .md file, as it is regenerated every several months.

openshwprojects:
Very nice, but does it mean that we now have PowerSave 1 and PowerSave 2 with different meanings on different platforms?

Maybe it would be better to change the naming, for example, PowerSaveNoRF or PowerSave NoRF (PowerSave 1 = full powersave, PowerSave 0 = disable, PowerSave NoRF = only mcu sleep)

Zanadar:
@NonPIayerCharacter thank you for the guidance regarding the comment... I have updated in cmd_main.c as suggested!

@openshwprojects I agree it is not ideal, obviously I did the best I could to add the new function as a minimalistic change whilst remaining non-breaking for the sake of upgrades. I am not sure it is an issue that the PowerSave setting behaves differently on different platforms as I can't imagine configurations would be shared between platforms often. What do you think?

Zanadar:
Is this now approvable?

divadiow:
Seems like a nice thing to have

openshwprojects:
Configurations may not be shared between platforms, but tutorials will be. And users are beginners, that's why I vote for something like "PowerSave RF" and then try to implement it on all platforms. Otherwise we'll just confuse users. If "PowerSave RF" is not present on current platform, do nothing.

openshwprojects:
We already have "On LN882H PowerSave 1 = light sleep and Powersave >1 (eg PowerSave 2) = deeper sleep." so I can't accept giving "PowerSave 2" different meaning on Beken, especially if we are going to have the same "rf only powersave" on LN882H in the future. So I suggest "PowerSave RF"

Keep all legacy syntax supported, just add check for "PowerSave RF" and if it's supported, do RF powersave for BL0937, if not, ignore

divadiow:
xref https://www.elektroda.com/rtvforum/viewtopic.php?p=21577881#21577881

openshwprojects:
So it seems to be automatic now, so I think we'll close this pr?

Issue #809, opened 2023-04-26
.rbl build for OpenBK7231T is broken for Ubuntu 22.04, works on Ubuntu 20.04


vfonov:
**Describe the bug**
Building firmware on Ubuntu 22.04 for OTA update (.rbl file) doesn't work on Ubuntu 22.04, either in Docker or on bare metal.
Works on Ubuntu 20.04

**Firmware:**
- Version: 1.17.29 possibly earlier too
- Device NA 
- Chip/model: BK7231T
- Device config: NA

**To Reproduce**
Using Docker build: 

1. create docker image `docker build -t openbk_build --build-arg USERNAME=$USER . `
2. request build for OpenBK7231T: `docker run --env TARGET_SDKS="OpenBK7231T" --rm -it -v "$(pwd)/..":/OpenBK7231T_App  openbk_build`
3. See error
```
../toolchain/gcc-arm-none-eabi-4_9-2015q1/bin/arm-none-eabi-nm ../../../apps/OpenBK7231T/output/dev_20230426_122610/OpenBK7231T_dev_20230426_122610.axf | sort > ../../../apps/OpenBK7231T/output/dev_20230426_122610/OpenBK7231T_dev_20230426_122610.map
../toolchain/gcc-arm-none-eabi-4_9-2015q1/bin/arm-none-eabi-objdump -d ../../../apps/OpenBK7231T/output/dev_20230426_122610/OpenBK7231T_dev_20230426_122610.axf > ../../../apps/OpenBK7231T/output/dev_20230426_122610/OpenBK7231T_dev_20230426_122610.asm
../toolchain/gcc-arm-none-eabi-4_9-2015q1/bin/arm-none-eabi-objcopy -O binary ../../../apps/OpenBK7231T/output/dev_20230426_122610/OpenBK7231T_dev_20230426_122610.axf ../../../apps/OpenBK7231T/output/dev_20230426_122610/OpenBK7231T_dev_20230426_122610.bin
make[3]: Leaving directory '/OpenBK7231T_App/sdk/OpenBK7231T/platforms/bk7231t/bk7231t_os'
make[2]: Leaving directory '/OpenBK7231T_App/sdk/OpenBK7231T/platforms/bk7231t/bk7231t_os'
Start Combined
make[1]: *** [Makefile:102: build-BK7231] Error 130
make[1]: Leaving directory '/OpenBK7231T_App'
make: *** [Makefile:74: OpenBK7231T] Error 2
[...]

Issue #600, opened 2023-01-10
Compiling Windows OpenBK simulator on Linux


Xinayder:
I've been seeing progress being made on the Windows port of OpenBK which will allow speedy testing and simulating of the firmware.

I checked the code and it's currently based off some Windows libraries, which would make it a bit troublesome to work on Linux or macOS natively. I'm not an expert in C++ or socket programming but I think the Windows-only calls can be substituted when dealing with Linux or macOS, which would make the simulator run on those OSes.

For that, I'd like to know how is it possible to compile the simulator right now; most of the libraries required are already present on Linux or are easy enough to install.

It seems like the only way to compile the simulator right now is by using Visual Studio. Is there a way to compile it via Make? As I mentioned, I'm not that savvy with C++ and that includes Make, so if there are any tips or suggestions that can guide me to create a build target for Windows, it would be appreciated; so I can mess around with it and try to get a version of the simulator running on Linux.

openshwprojects:
Hello, thank you for your interest in Simulator (CC: @iprak  might be interested in this).
Currently, the only Windows-only part of Simulator is this menu bar:
![image](https://user-images.githubusercontent.com/85486843/211682057-2482c4a3-4562-40db-b8d4-aaa784f18f75.png)
I was unable to do it without WINAPI, at least I was unable to do it quickly.
Everything else should be portable. Sockets are portable. The time function can be also easily substituted on other platforms.

The simulator is almost fully functional, the only thing left to resolve is a MQTT connect lag that happens because of a blocking TCP connect call. 

I have compiled Simulator only in MSVC2017 so far. I do not know much about Make, it's not within my skills set yet, but it obviously should be very possible.

I will happily assist you with running/testing/building simulator and I have time for that, but you'll have to start figuring out Make stuff from scratch (same as me).

I am planning to upload simulator dependencies to separate repo soon.

Xinayder:
What I want to know firstly is, is the simulator (the screenshot you showed above) a separate part from the OpenBK port for Windows? That is, is it possible for me to just simulate the MCU running OpenBK without having to run the simulator?

If it's possible, it would be easier to start by porting the socket stack then porting the simulator. I did a quick search and turns out mimicking this menu bar outside Windows is tricky because it might depend on whether you want to use GTK or Qt or something else for your GUI. I might be wrong though, I'm not that familiar with low level programming using libs like SDL2 or OpenGL.

openshwprojects:
[...]

PR #1544, opened 2025-02-23
[POC] Run berry-lang

niteria:
This is just a proof-of-concept, is only meant to start a discussion and not in any way ready to be merged.

# Motivation

OpenBK tries to be in many ways compatible with Tasmota. One example of this compatibility is the HTTP API. 
Tasmota uses Berry: https://tasmota.github.io/docs/Berry/ for its scripting needs.

# What this is

I wanted to check how difficult it would be to add Berry in some capacity and have it expose one of the basic functionalities: `ChannelSet`.

This introduces a new command, `berry [string]`, which takes the input string and evaluates it as Berry code.

For example `berry 1+2` produces this in the log:
```
Info:CMD:[berry start]
Info:CMD:top = 1
Info:CMD:stack[0] = isnumber = 1
Info:CMD:stack[0] = isnil = 0
Info:CMD:stack[0] = isbool = 0
Info:CMD:stack[0] = isint = 1
Info:CMD:stack[0] = isreal = 0
Info:CMD:stack[0] = 3
Info:CMD:[berry end]
Info:CMD:[WebApp Cmd 'berry 1+2' Result] OK
```

It's not pretty, but we can see that it works as a calculator: `Info:CMD:stack[0] = 3` contains the result.

---

I have also bound the OpenBK function `ChannelSet` to the Berry function `channelSet(channel, value)`.

`berry channelSet(1, 0)` produces the following log output, which in my case turns off my smart plug:

```
Info:CMD:[berry start]
Info:CMD:be_ChannelSet top = 2
Info:CMD:be_ChannelSet ch = 1, v = 0
Info:GEN:CHANNEL_Set channel 1 has changed to 0 (flags 0)

Info:MQTT:Channel has changed! Publishing 0 to channel 1 
Info:CMD:top = 1
Info:CMD:stack[0] = isnumber = 0
Info:CMD:stack[0] = isnil = 1
Info:CMD:stack[0] = isbool = 0
Info:CMD:stack[0] = isint = 0
Info:CMD:stack[0] = isreal = 0
Info:CMD:[berry end]
```

# But...

I took some shortcuts:
* I create a new VM for every evaluation
* disabled a lot of features of Berry to get it to compile in the most expedient way possible
  * https://github.com/niteria/berry/commit/d2214af3a78eedad4957700163e49e0eab027a77 is what I had to do
* The build integration is currently messy; we link every Berry object file, even for features we've disabled.

There are some questions I haven't answered:
* how much does it increase the binary size?
* will it fit in RAM/stack on every supported platform?
* OpenBK already has a working language

# A dream

With this taken to its logical conclusion, we could have Berry API compatibility with Tasmota, `autoexec.be` and reusable modules stored on LFS and shared between users.

# My goal with this PR

I would love some feedback on whether the project would like to move in this direction.
It’s possible that this idea has been considered before and rejected for specific reasons.

# Some relevant resources with API, docs...

* Berry
  * Docs: https://berry.readthedocs.io/en/latest/
  * Porting Guide: https://berry.readthedocs.io/en/latest/source/en/Appendix-C.html
  * FFI: https://berry.readthedocs.io/en/latest/source/en/FFI-Example.html
* Tasmota+Berry:
  * Easier FFI: https://github.com/arendst/Tasmota/blob/development/lib/libesp32/berry_mapping/README.md


openshwprojects:
Very nice and promising, altough I see it currently has incorrect include path for Github build?

Please fix, I really like the idea. I've been thinking about doing similar but I considered lua, didn't check out berry yet but it seems good.

It would be useful to have it working also on OBK simulator (windows build).

don't worry about flash/ram usage, we have obk_config.h , we can have berry build and classic build

openshwprojects:
I've tried to get it to compile online but it's still missing a prebuild of berry step:
https://github.com/openshwprojects/OpenBK7231T_App/pull/1545/files

niteria:
Interesting, I wondered if the CI checkouts the repo with `--recurse-submodules` and expected it might be a problem.
From the log it still looks like the `src/berry` submodule is not checked out.
I will fix, or just in the interest of demo-ing copy over the directory instead of having a submodule.

niteria:
FWIW I've only built it for BK7231T on my machine using docker (because of #809)

niteria:
I've tried to build locally from your PR and I've noticed one mistake, but I don't think the build got that far.

niteria:
I was missing autogenerated files: https://github.com/berry-lang/berry/commit/bd42588c100694335ee81001552a4f05df9c678b

openshwprojects:
N builds well, rel is 486 KB, this PR is 531 KB, not bad

niteria:
Ok, I clearly need to incorporate it into the build system better, only OpenBK7231N and OpenBK7231N automagically compile every `.c` file and link it into final binary.

openshwprojects:
I can try to improve it a bit in few moments

openshwprojects:
Can berry support "waitframe" or "delay_s" command? I don't know how it works under the hood, but it would be good to have a linked list of active berry scripts that is able to "wait" given script thread for some time.

openshwprojects:
Something like that:
image

niteria:
> Can berry support "waitframe" or "delay_s" command? I don't know how it works under the hood, but it would be good to have a linked list of active berry scripts that is able to "wait" given script thread for some time.

The way Tasmota solved waiting is timers: https://tasmota.github.io/docs/Berry/#timers, very similar mechanism to what JavaScript does.

> On a second tought, maybe we could instead put berry into the cmd_script.c? Try to reuse those structures:

I'm still getting my bearings on OpenBK code, but having it be part of the event loop sounds very reasonable.
Otherwise I wouldn't know know how to implement `delay_s` or `waitFor`.

openshwprojects:
I would really like to have delay_s inside Berry script. Do you know Berry SDK enough to know how it can be done?

The way it would be implemented is fairly simple as long as Berry have a way to "pause execution after given C function call". The rest (waiting N seconds etc) would be done by OBK existing code.

I think we could just iterate all OBK commands and make them available in berry.

openshwprojects:
This is how current delay_s is saved:

```
static commandResult_t CMD_Delay_s(const void *context, const char *cmd, const char *args, int cmdFlags){
	float del;
	int delMS;

	Tokenizer_TokenizeString(args,0);
	// following check must be done after 'Tokenizer_TokenizeString',
	// so we know arguments count in Tokenizer. 'cmd' argument is
	// only for warning display
	if (Tokenizer_CheckArgsCountAndPrintWarning(cmd, 1)) {
		return CMD_RES_NOT_ENOUGH_ARGUMENTS;
	}
	if(g_activeThread == 0) {
		ADDLOG_INFO(LOG_FEATURE_CMD, "CMD_Delay_s: this can be only used from a script");
		return CMD_RES_ERROR;
	}

	del = Tokenizer_GetArgFloat(0);
	delMS = del * 1000;
	ADDLOG_EXTRADEBUG(LOG_FEATURE_CMD, "CMD_Delay_s: thread will delay %i extra ms\n",delMS);
	g_activeThread->currentDelayMS += delMS;


	return CMD_RES_OK;
}
```
But I am not sure about that be_return(vm);  , it requires futher testing.  I am worried that be_return just returns a value to berry call, and the berry execution is always resumed? I don't know yet

niteria:
> But I am not sure about that be_return(vm); , it requires futher testing. I am worried that be_return just returns a value to berry call, and the berry execution is always resumed? I don't know yet

I'm afraid you may be right.
Berry doesn't support coroutines: https://github.com/berry-lang/berry/issues/72

I'm sure it's possible to add them, but apparently Lua already has them: https://www.lua.org/pil/9.1.html

It's not clear whether https://github.com/szieke/embLua implements them and the project seems pretty inactive.

openshwprojects:
So what would you suggest?

I used lua before but I used it without "wait" as well, I had just callbacks and timers. 

So maybe we can try making berry integration for obk but inside cmd_berry.c, without supporting delay_s and waitfor?

For console commands, we could try to just iterate all of commands and feed arguments via obk tokenizer... so everything works out of the box..
Or maybe... add a command "obkCommand(name,arg1,arg2,arg3) as a obk commandline wrapper, this sounds better and cleaner.

niteria:
I don't have good enough understanding of the goals of the project to make a suggestion yet.

There are some advantages to using Berry:
* Tasmota uses it in a similar context, it works for them, and we can borrow ideas from them
* Berry seems to be an actively maintained project
* I expect it to have a smaller footprint than Lua (embLua has some ballpark numbers, eg. "TI CC2642R1: 2k stack, 20k Heap, 160k ROM (-O0), 100k ROM (-Oz)", I can test this)

There are some advantages to using Lua:
* It's a popular language
* It supports coroutines (needs to be tested on embLua)

I think programming with callbacks is a perfectly acceptable model, it's a bit heavy syntactically, `delay_s` would have to take a closure (lambda function / anonymous function / callback). Code using it would look something like this:

```
def turnOnAndOffAfter5sTwice()
  channelSet(1, 1)
  delay_s(5, / ->
    channelSet(1, 0)
    delay_s(5, / ->
        channelSet(1, 1)
           delay_s(5, / ->
              channelSet(1, 0)
           )
     )
  )
end
```

Relevant docs: https://berry.readthedocs.io/en/latest/source/en/Chapter-5.html#lambda-expression

Like in similar languages that use callbacks (eg Javascript) we could make it more convenient with Promises:

```
def turnOnAndOffAfter5sTwice()
  channelSet(1, 1)
  delay_s(5)
    .then(\ -> channelSet(1, 0))
    .then(delay_s(5)
    .then(\ -> channelSet(1, 1))
    .then(delay_s(5)
    .then(\ -> channelSet(1, 0))
end
```

niteria:
I think I like Berry more because we can benefit from the community around it, while with `embLua` I expect we would effectively need to become maintainers of `embLua`.

But I don't want to make it sound like `embLua` is the only `lua` that can run on microcontrollers, I just haven't made any research yet.

niteria:
I think to gather more data on whether Berry is viable I will take one of the obk scripts I use ([Advanced turn off after time with timer on UI and timer setting on UI and kept in flash](https://www.elektroda.com/rtvforum/viewtopic.php?t=4032982&highlight=) from https://github.com/openshwprojects/OpenBK7231T_App/blob/main/docs/autoexecExamples.md) and try to implement it in Berry as `autoexec.be`.

openshwprojects:
Can we compile berry for Windows simulator as well? It would help a lot

openshwprojects:
Are you aware about OBK self tests? They are like unit tests. They run on Windows build.

niteria:
I can take a look and try to add it to the Windows simulator, but I don't have a build environment set up on Windows, so I expect it might take some time.

Berry is quite easy to compile straight from the checkout (you just need `make` and `gcc`), so maybe in the mean time that would be useful for you: https://github.com/berry-lang/berry?tab=readme-ov-file#build-and-run

They also have an online version where you can just run Berry code: https://berry-lang.github.io/try-online/

openshwprojects:
Adding berry on Windows requires two steps.
1. enable it in obk_config.h (I added define for berry)
2. add requird file paths here:
https://github.com/openshwprojects/OpenBK7231T_App/blob/main/openBeken_win32_mvsc2017.vcxproj

niteria:
I was able to compile it in obkSimulator, but it crashes on any `berry` command.
I've locally removed the FFI binding, and it crashes even on simple expressions like: `berry 1+1` in the parser.
Initial debugging points to some kind of memory overrun, the data in `vm->strtab.table` is mangled "from afar" by some `memcpy`.

I will need to check whether a clean checkout of Berry runs under MSVC2017 correctly.
I suspect some disagreement between MSVC2017 and GCC on what some part of the code means, or just straight-up undefined behavior.

niteria:
After the last update obkSimulator no longer crashes on `berry` commands on my machine, but somehow the binary downloaded from CI crashes.

openshwprojects:
That's a very strange error, maybe you can also report it to berry's repo... 

Let me know what's the state of things now, is any help needed?

On my side, I added another self test system, this time the one that can run on any hardware device, details here:
https://www.elektroda.com/rtvforum/viewtopic.php?p=21457592#21457592
Maybe it can be used to help testing berry at later stages.

niteria:
> That's a very strange error, maybe you can also report it to berry's repo...

It's already fixed upstream: https://github.com/berry-lang/berry/issues/460

> Let me know what's the state of things now, is any help needed?

I need to figure out why the CI built obkSimulator crashes on `berry 1+1` while it works locally for me.
I hadn't had time to dig into that, but perhaps if it's not too much trouble it'd be helpful to know if a locally built obkSimulator also works for you on `berry 1+1`.

niteria:
> On my side, I added another self test system, this time the one that can run on any hardware device, details here:
https://www.elektroda.com/rtvforum/viewtopic.php?p=21457592#21457592
> Maybe it can be used to help testing berry at later stages.

Thanks, I will keep that in mind.

niteria:
> msbuild openBeken_win32_mvsc2017.vcxproj /p:Configuration=Release /p:PlatformToolset=v143

So the CI is actually using Visual Studio 2022 to compile... Gotta set up a new vm with newer Windows...

niteria:
obkSimulator from CI no longer crashes for me. It was because of the mess I made getting it to build initially.

Things to do: 

- [ ] clean up the build process, target: official `berry` as submodule
- [ ] autoexec.be
- [ ] enough support to implement a timer script I use on my plug

openshwprojects:
I've checked the new be_port etc files and noticed file operations function wrappers - does it mean that we could use Berry among LFS to write files? That would be very useful.

niteria:
I believe so, here's Tasmota's glue code: https://github.com/arendst/Tasmota/blob/master/lib/libesp32/berry_tasmota/src/be_port.cpp#L21

niteria:
After a bit of a detour with #1563 I have a most basic version of `delayMs` working.

I'm exploring adding async/await syntax to `berry`, the maintainers seemed sympathetic to it in the past https://github.com/berry-lang/berry/issues/66 and `berry` is quite a small language and already supports closures, so no extension would be necessary to the virtual machine itself.

niteria:
I can now add a file `autoexec.be` (can have a different name):

```
autoexec = module("autoexec")

autoexec.init = def (self)
	print("Hello from autoexec.be")
	return self
end

return autoexec
```

and `berry import autoexec` will import it and run `init`.

The idea is that I can just add `berry import autoexec` to `autoexec.bat` and have berry running after restart if I want.

niteria:
With this `autoexec.be`:

```
autoexec = module("autoexec")

autoexec.make_timer = def(relayCh, countdownCh, turnOffTimeCh)
  setStartValue(turnOffTimeCh, -1)
  if !channelGet(turnOffTimeCh)
	 channelSet(turnOffTimeCh, 15)
  end
  setChannelType(turnOffTimeCh, 'TextField')
  setChannelLabel(turnOffTimeCh, 'Turn off time')

  setChannelType(countdownCh, 'TimerSeconds')

  def tick()
	v = channelGet(countdownCh)
	v -= 1
	channelSet(countdownCh, v)
	if v > 0
	  delayMs(1000, tick)
	else
	  channelSet(relayCh, 0)
	end
  end  

  channelEv = 'Channel' + str(relayCh)
  def onTurnedOn()
	channelSet(countdownCh, channelGet(turnOffTimeCh))
	delayMs(1000, tick)
	addChangeHandler(channelEv, '==', 1, onTurnedOn)
  end

  def onTurnedOff()
	channelSet(countdownCh, 0)
	addChangeHandler(channelEv, '==', 0, onTurnedOff)
  end

  addChangeHandler(channelEv, '==', 1, onTurnedOn)
  addChangeHandler(channelEv, '==', 0, onTurnedOff)
end

autoexec.init = def (self)
	print("autoexec.be")
	return self
end

return autoexec
```

I can now: `berry import autoexec; autoexec.make_timer(1, 2, 3)` to get a countdown timer.

niteria:
@openshwprojects
Looking at the timer implementation above I realise that it would not benefit much from an `async/await` syntax.
It subscribes to a few events and the only place where it's waiting in a semi-sequential manner is in the calls to `delayMs(1000, tick)`

Other scripts might benefit more from `async/await`, especially if extended with HTTP request capabilities, which would highlight its advantages. However, I believe this pull request has achieved its proof-of-concept goals and reached its threshold of usefulness.

I’m inclined to merge this as-is and continue improving it incrementally as needs arise. There are some small issues I need to clean up, but I’d like your opinion:

1. Is the current functionality worth merging?
2. Do you see any blockers?

Outstanding Issues:
  
* [ ] The Windows build clones all submodules unnecessarily.
* [ ] Unnecessary debugging statements remain in the code.
* [ ] Silent failures occur on incorrect argument counts/functions in bound functions like channelSet. (I’m tempted to defer fixing this unless it causes problems.)
* [ ] No way to restart the `berry` VM without restarting the device.
* [ ] Killing a `berry` SVM thread does not free the associated closure.
* [ ] The `cmd_berry.c` file is now redundant.
* [ ] Handling of uniqueIDs is hacky but seems tolerable since SVM doesn’t heavily rely on them.
* [ ] Some copy-paste code remains in be_AddChangeHandler.
* [ ] Local testing files (`*.user`) accidentally snuck into commits.

This pull request has a long commit history with several non-building commits. Would you prefer I clean it up or start a fresh pull request?

openshwprojects:
I think it's certainly worth merging as-is, just make sure it's disabled by default and that there is a clear instruction how to enable it via obk_config.h . Maybe add separate .md file and link to it from doc readme?

@NonPIayerCharacter changes are also good, I would like to have them as well, but my only question would be how much flash footprint extra function binding have? If they are taking much space, we will probably need to split obk_config options into OBK_ENABLE_BERRY and BERRY_ENABLE_I2C, etc, etc

Don't worry about commits history, I can do "squash and merge" here and it will be merged as a single commit. Maybe just let me know if you're able to merge @NonPIayerCharacter changes as well before I merge this PR?

niteria:
That's very cool. 

You probably want a separate `bvm* g_vm` for your `rtos` thread, I don't believe `berry` is thread-safe.

> rtosDelayMs feels like it's delaying twice as much.

Maybe it's just slow... You could try dumping the byte code of the functions in question. `debug.codedump` will print the bytecode of a function: https://github.com/berry-lang/berry/blob/master/src/be_debuglib.c#L269

Your branch for my future convenience: https://github.com/NonPIayerCharacter/OpenBK7231T_App/commits/_berry/

niteria:
> I think it's certainly worth merging as-is, just make sure it's disabled by default and that there is a clear instruction how to enable it via obk_config.h . Maybe add separate .md file and link to it from doc readme?

Will do. I will also let you know when I'm done doing all the clean up and consider it ready.

> Maybe just let me know if you're able to merge @NonPIayerCharacter changes as well before I merge this PR?

I don't have a setup to test I2C and I'm unfamiliar with rtos threads, so I think it's better if @NonPIayerCharacter sends a PR after this is merged.

NonPIayerCharacter:
@niteria i opened a pull request to your repo
Cleaned my code && tidied it up.
Changed how files are handled, we don't need to check if lfs is present in every function, only in file open.
Instead check if file is not null.
Moved most of the code from cmd_script to cmd_berry
Aforementioned i2c

Full driver code for AHT2X, written in berry. Unfortunately, it cannot be imported in script/autoexec. I get stack overflow in timer thread on BK7238.
```
aht2x = module("aht2x")
import string
aht2x.init = def (self)
	class AHT2X
		var i2c
		var ChTemp, ChHum
		var CalTemp, CalHum
		def init()
			self.ChTemp = -1
			self.ChHum = -1
			self.CalTemp = 0
			self.CalHum = 0
		end
		def initI2c(clk, data)
			self.i2c = initI2c(clk, data)
			print("init...")
		end
		def setChannels(chtemp, chhum)
			self.ChTemp = chtemp
			self.ChHum = chhum
			setChannelType(chhum, "Humidity_div10")
			setChannelType(chtemp, "Temperature_div10")
			setChannelLabel(chhum, "AHT2X Humidity")
			setChannelLabel(chtemp, "AHT2X Temperature")
			print(string.format("Channels set, temp: %i, hum: %i", chtemp, chhum))
		end
		
		def calibrate(temp, hum)
			self.CalTemp = temp
			self.CalHum = hum
			print(string.format("AHT2X_Calibrate: Calibration done temp %f and humidity %f", self.CalTemp, self.CalHum))
		end
		
		def reset()
			startI2c(self.i2c, (0x38 << 1))
			writeByteI2c(self.i2c, 0xBA)
			stopI2c(self.i2c)
			rtosDelayMs(20)
		end
		def sensorInit()
			startI2c(self.i2c, (0x38 << 1))
			writeByteI2c(self.i2c, 0xBE)
			writeByteI2c(self.i2c, 0x08)
			writeByteI2c(self.i2c, 0)
			stopI2c(self.i2c)
			rtosDelayMs(20)
			var data = 128
			for i:0..15
				if (data & 0x80) != 0
					print("Initializing...")
					rtosDelayMs(20)
					startI2c(self.i2c, (0x38 << 1) | 1)
					data = readByteI2c(self.i2c, true)
					stopI2c(self.i2c)
				else
					break
				end
			end
			if ((data & 0x68) != 0x08)
				print("Initialization Failed!")
			else
				print("Init success")
			end
		end
		def measure()
			startI2c(self.i2c, (0x38 << 1))
			writeByteI2c(self.i2c, 0xAC)
			writeByteI2c(self.i2c, 0x33)
			writeByteI2c(self.i2c, 0)
			stopI2c(self.i2c)
			rtosDelayMs(80)
			var data = [0x80]
			bool ready = false
			for i:0..10
				print(data)
				if (data[0] & 0x80) != 0x80
					print("ready")
					ready = true
					break
				else
					print("Sensor is busy, waiting...")
					rtosDelayMs(20)
					startI2c(self.i2c, (0x38 << 1) | 1)
					data = readBytesI2c(self.i2c, 6)
					stopI2c(self.i2c)
				end
			end
			if ready != true
				print("Timed out")
				return
			end
			if(data[1] == 0x0 && data[2] == 0x0 && (data[3] >> 4) == 0x0)
				print("Humidity unrealistic")
				return
			end
			int raw_temperature = ((data[3] & 0x0F) << 16) | (data[4] << 8) | data[5]
			int raw_humidity = ((data[1] << 16) | (data[2] << 8) | data[3]) >> 4
			
			real g_humid = (raw_humidity * 100.0 / 1048576.0) + self.CalHum
			real g_temp = (((200.0 * raw_temperature) / 1048576.0) - 50.0) + self.CalTemp
			if(self.ChTemp != -1)
				channelSet(self.ChTemp, int(g_temp * 10))
			else
				print("ch temp is not set")
			end
			if(self.ChHum != -1)
				channelSet(self.ChHum, int(g_humid * 10))
			else
				print("ch hum is not set")
			end
			print(string.format("AHT2X_Measure: Temperature:%fC Humidity:%f%%", g_temp, g_humid))
		end
	end
	return AHT2X()
end

return aht2x
```
```
//berry import aht2x
berry aht2x.initI2c(22,15)
berry aht2x.setChannels(1,2)
berry aht2x.reset()
berry aht2x.sensorInit()
berry aht2x.measure()
meas:
berry aht2x.measure()
delay_s 5
goto meas
```

NonPIayerCharacter:
Also should mention, that size increased catastrophically. Current release BK7238 ota image is 483kb. This branch is 596kb. Currently it overwrites lfs on ota.

niteria:
There are some options for the code size, I want to explore them. I purposefully limited the initial set of FFI functions for it to not be blocker.

I didn't want to expand the scope of this pull request beyond proving the concept, there's already quite a bit going on: build system with autogenerated files, changes to SVM scheduler, FFI functions...

openshwprojects:
Don't worry about flash size, the feature will be disabled by default, and we'll add a proper warnings to readme. 

I will also test it more a bit later.

In the future, it would be nice to have option to build multiple obk_config.h combinations per single online build, but that's a matter for totally separate PR.

niteria:
New functions like `be_initI2c` added by @NonPIayerCharacter highlight a problem of mixing a garbage collected language and a language with manual allocation and deallocation.

If I:
```
i2c = initI2c(clk, data)
```
and then try to destroy the VM, I will leak the memory (and other resources) that `malloc(sizeof(softI2C_t));` in `be_initI2c` allocated.

This is a contrived example, but there are other examples that will leak:

```
i2c = initI2c(clk, data)
i2c = initI2c(clk, data)
```
or
```
i2c = initI2c(clk, data)
a() // throws
be_deinitI2c(i2c)
```
or

```
i2c = initI2c(clk, data)
delayMs(10000, / -> be_deinitI2c(i2c))
// now restart SVM
```

`berry` provides a mechanism that deals with this, but it's not well documented.

`BERRY_API void be_newcomobj(bvm *vm, void *data, bntvfunc destroy);` associates a `destroy` function with a pointer
than later the GC will use when the object is garbage collected during normal operation or on shutdown.

@NonPIayerCharacter: could you send a PR to my branch using `be_newcomobj`?

openshwprojects:
be_newcomobj seems like a good solution at the first glance, but I wonder how it works internally?

Will it work with:
```
i2c = initI2c(clk, data)
i2c = initI2c(clk, data)
```
and if so, will it also handle:
```
i2c = initI2c(clk, data)
i2c2 = i2c;
i2c3 = i2c;
```
Is it internally reference-counted?

niteria:
(I run my original answer through perplexity to improve readability)

### Berry's Garbage Collection: Mark-Sweep Explained

Berry doesn't use reference counting for memory management. Instead, it employs a Mark-Sweep Garbage Collection (GC) algorithm, a type of *tracing garbage collection*. A good overview of this can be found on Wikipedia: [https://en.wikipedia.org/wiki/Tracing_garbage_collection\#Tri-color_marking](https://en.wikipedia.org/wiki/Tracing_garbage_collection#Tri-color_marking).

In essence, the GC identifies live objects by tracing through the object graph, starting from a set of "root" objects. Think of these roots as objects that are always considered live – global variables, values held in VM registers, or data on the VM stack. Anything the GC *can't* reach from these roots is considered dead and is then freed. This process is often implemented using a technique called *tri-color marking*, which divides objects into sets of white, gray, and black to efficiently track reachability during the collection cycle.

### Examples

#### Example 1: Object Replacement

```plaintext
i2c = initI2c(clk, data)
i2c = initI2c(clk, data)
```

Here, the two calls to `initI2c` each create a new, distinct object. Let's call them `A` and `B`. After the first line, the variable `i2c` points to object `A`. This keeps `A` alive as it is reachable from a root. However, after the second line, `i2c` is reassigned to point to object `B`. Now, *nothing* points to `A` anymore. During the next GC cycle, `A` will be deemed unreachable and its memory will be freed.

#### Example 2: Multiple References

```plaintext
i2c = initI2c(clk, data)
i2c2 = i2c;
i2c3 = i2c;
```

In this case, *three* variables (`i2c`, `i2c2`, and `i2c3`) all point to the *same* object. The object will remain alive as long as at least one of these variables points to it. If you want the object to be freed, you need to assign all three variables to something else, such as `nil`.

### The Advantage of Garbage Collection: Handling Circular References

One major advantage of GC over reference counting is its ability to handle circular references. Consider this scenario:

```plaintext
a = []
b = []
a.push(b)
b.push(a)
```

Here, `a` contains a reference to `b`, and `b` contains a reference to `a`, creating a loop. Reference counting would fail to collect these objects because each object always has a reference to the other, even if they are no longer used by the program. GC, however, can correctly identify that `a` and `b` are unreachable from the root set and free them.

### `be_newcomobj`: Custom Destruction

The `be_newcomobj` function provides a crucial capability: it allows us to tell the GC that when a C pointer becomes unreachable, a specific C function needs to be called to destroy the underlying resource.

This is particularly important when dealing with:

* **Memory allocated with `malloc`:** The associated function would be `free`.
* **C++ objects:** You'd need to call `delete` to properly invoke the object's destructor.
* **External Resources:** If the pointer represents a resource like a file handle or an open network connection, the associated function would close the handle or connection.

In essence, `be_newcomobj` ensures that resources held by C/C++ code are correctly cleaned up when the Berry GC determines they are no longer needed, providing a bridge between the managed and unmanaged worlds.

niteria:
In one of my future PRs I want to integrate https://github.com/berry-lang/berry_mapping/tree/main to make defining C bindings easy and possibly reduce the binary size.

It should also enable a crazy feature, an ability to call a C function without it being bound previously. I suspect the final binary has symbols stripped, but the build process generates a `.map` file with addresses of all symbols.

It also has a different approach to what in my PR are `_suspended_closures`. It may or may not be better. Something to be explored.

Is there some place that describes what's our binary size budget? I've also run into OTA overwriting my lfs with this PR.

NonPIayerCharacter:
I was aware of a potential memory leak with i2c, which is why i made be_deinitI2c to free said memory, and left it up to script writers to utilize it. And i think if vm is destroyed, then something is already seriously wrong, that some small memory leak would be insignificant.
That said, i was unaware of be_newcomobj, which seems to be a good thing. It would also allow to reduce amount of code by not writing deinitializers/destructors, like i used for i2c, and leave it up to berry to free memory.

Berry mappings also seems very good, but from what i see in quick example, c functions still need to be bound.
I think i will eventually rewrite i2c to be a module, instead of current c-like functions implementation. Then add gpio, irqs, etc.

Btw, how to compile .be to .bec currently?

niteria:
> I was aware of a potential memory leak with i2c, which is why i made be_deinitI2c to free said memory, and left it up to script writers to utilize it. And i think if vm is destroyed (can it crash?), then something is already seriously wrong, that some small memory leak would be insignificant. (unless a user deliberately closes the vm, which is still questionably why would he?) That said, i was unaware of be_newcomobj, which seems to be a good thing. It would also allow to reduce amount of code by not writing deinitializers/destructors, like i used for i2c, and leave it up to berry to free memory.

As explained in https://github.com/openshwprojects/OpenBK7231T_App/pull/1544#issuecomment-2722595165 it is **not only during VM shutdown**.

niteria:
@NonPIayerCharacter: Are you planning to fix this?

NonPIayerCharacter:
@niteria https://github.com/niteria/OpenBK7231T_App/pull/2

openshwprojects:
Very good job guys, it looks very promising. I'm planning to find some more free time to look into it deeper, I'm also considering making a stripped-down OBK version with more low level API for berry, like you know, WiFi connection, etc, direct pins access. Those C binding would also come handy then.

openshwprojects:
Any news? What's left before merge?

Or shall I merge and try to help with futher integration?

openshwprojects:
Is Berry able to process binary data? That could help with TuyaMCU stuff.

How much is missing for #600 ? Do you need #600 with SDL window support on Linux, or just OBK itself?

niteria:
> Is Berry able to process binary data? That could help with TuyaMCU stuff.

There's the `bytes` datatype for binary data https://berry.readthedocs.io/en/latest/source/en/Chapter-2.html#bytes
I think it should have no problem processing binary data.
This is not exactly TuyaMCU, but I think you'll find this interesting https://tasmota.github.io/docs/ULP/#blink-an-led-fsm

> How much is missing for #600 ? Do you need #600 with SDL window support on Linux, or just OBK itself?

I only want a convenient way to run the tests on Linux, so I don't need SDL.

openshwprojects:
I will look into it now, maybe I can help, but I can check now just with WSL 
image

openshwprojects:
my progress so far:
https://github.com/openshwprojects/OpenBK7231T_App/pull/1577

openshwprojects:
I'm still on it, it's going slowly but no serious issues so far. The only downside is that I'm following the approach from #600 which kinda seems to use WINDOWS define as in "SIMULATOR" define, so there will be some little cleaning to be done once I get it working.

niteria:
@openshwprojects: 
I've reached a state of this PR that I'm happy with. 
Please take a look and see if anything needs fixing before merging. 
I've left it enabled on Windows by default and disabled for all the other platforms.

After this gets merged I will take a break from working on Berry related stuff, but feel free to tag me on related PR's and issues.

openshwprojects:
I really like how it turned out, @niteria  . I'm playing around witth it and I'm trying to get this button working natively:
image
I've allowed startScript to check for be ext and then rerun script via berry. It works at first try, however, it seems that pressing button again stops the script and it does not rerun... which brings the question, is there some berry state that is kept between **resetSVM** calls?

niteria:
> I really like how it turned out, @niteria . I'm playing around witth it and I'm trying to get this button working natively: image I've allowed startScript to check for be ext and then rerun script via berry. It works at first try, however, it seems that pressing button again stops the script and it does not rerun... which brings the question, is there some berry state that is kept between **resetSVM** calls?

I `resetSVM` needs to do the same thing as `CMD_StopBerry`, there's global state held in `g_vm` like imported modules and globals. I forgot to do that.

openshwprojects:
ok i will add it

niteria:
The behavior of import is pretty well explained here: https://berry.readthedocs.io/en/latest/source/en/Chapter-8.html#behavior-of-import

openshwprojects:
@niteria what would be the best way to call Berry from C? I'm am considering handlers for stuff like MQTT incoming data (topic, payload) and TuyaMCU data points (dpID, type, value) or maybe even uart (hex string).
In docs I can see:

```
    be_loadstring(vm, "print('Hello Berry')"); // Compile test code
    be_pcall(vm, 0); // Call function
```
so the simplest idea is to let user specify module name and function and then just do:

```
  char buffer[512];
sprintf(buffer,"import %s; %s(%s)",module, funcName, args);
    be_loadstring(vm, buffer); // Compile test code
    be_pcall(vm, 0); // Call function
```
but maybe there is more efficient way? Manually load stuff on stack... any suggestions?

niteria:
Manually loading stuff on the stack like in `be_ScriptDelayMs` when we call `suspend_closure` avoids the need to parse and compile the code in the string, so it has to be more efficient.

It also lets Berry return some result to C in a structured way.

For handlers I would go with a similar system to the one you already have for events. Have a `handleEvent` function on a Berry side that you call from C, and then in Berry that function would just look through some data structure (probably a `map`) to see if someone registered a callback for that event.

Then you could have a C function that wraps calling of `handleEvent` with different number of arguments.

I can sketch something up if my suggestion if too vague, but I'm not sure when I'll have the time to do that.

openshwprojects:
I'm analyzing the code and this approach makes me wonder:
```
const char berryPrelude[] =
	"_suspended_closures = {}\n"
	"_suspended_closures_idx = 0\n"
	"\n"
	"def suspend_closure(closure)\n"
	"  _suspended_closures_idx += 1\n"
	"  _suspended_closures[_suspended_closures_idx] = closure\n"
	"  return _suspended_closures_idx\n"
	"end\n"
	"\n"
	"def resume_closure(idx)\n"
	"  closure = _suspended_closures[idx]\n"
	"  _suspended_closures.remove(idx)\n"
	"  closure()\n"
	"end\n";
```
why is it even needed? I remember using LUA C library in some of my projects and I always was able to just get a reference to function there, store it in integer, and then push it to stack with a variable number of arguments and call it.... Is  berry different? Why is it done in Berry itself and not in C?

niteria:
There used to be a way to do it in Berry directly via "registry" (https://github.com/berry-lang/berry/issues/65), but it's since has been removed: https://github.com/berry-lang/berry/pull/276

You can't simply get a reference to a function because it might be a closure, which retains references to objects on the heap that it has captured. As a result, explicit handling is necessary to inform the Garbage Collector whether the function is still in use or can be discarded.

Putting the function/closure in the `_suspended_closures` map achieves 2 things: 
* gives us an index to be able to refer to it later
* makes the function/closure one of the GC roots, keeping it alive

https://github.com/berry-lang/berry_mapping/blob/main/src/be_cb_module.c is an alternative approach for this, but it has its own limitations.

openshwprojects:
So berryPrelude is basically an universal mechanism of mapping function (with context) to index?

I see, my current idea (probably for separate PR, maybe very soon) is to change current berryPrelude handling into an universal "closure to index" system with index register and then use it for both be_ScriptDelayMs  and for events

niteria:
> So berryPrelude is basically an universal mechanism of mapping function (with context) to index?

Yes, and it it lets the garbage collector know that something outside of Berry wants to keep the function alive.

> I see, my current idea (probably for separate PR, maybe very soon) is to change current berryPrelude handling into an universal "closure to index" system with index register and then use it for both be_ScriptDelayMs and for events

Sounds interesting, I will take a look once it's done.

I want to get this PR merged to have a stable base for experimenting with different ideas in parallel.

`_suspend_closures` is purposefully as simple as possible, so that there's no question about correctness.

Having a simple, stable base also lets you measure whether complex optimizations are actually worth it.

openshwprojects:
I'll merge it and I'll ping you in related upcoming PRs. Thanks for valuable contribution!

openshwprojects (approved):

openshwprojects (approved):

openshwprojects:
there is final thing that I am not sure about - why Berry commands are "channelSet" and "channelGet" while OBK commands are setChannel, addChannel, etc?

niteria:
No real reason, I think I forgot to change them to match OBK. 

I also expected to write a a more OOP interface:
```
ch = Channel(10)
ch.set(1)
ch.add(2)
```

niteria:
> I'll merge it and I'll ping you in related upcoming PRs. Thanks for valuable contribution!

Thank you for the review and the final clean up!

PR #1548, opened 2025-02-26
Cmd Flags 64bit repair + webui

xjikka:
fix cmd flags to support 64bit + display on webui in Flags section


xjikka:
Checked on simulator and BK7238.
@divadiow - if you have some time, could you please check it?

openshwprojects (approved):

PR #1553, opened 2025-03-02
ECR6600

NonPIayerCharacter:


openshwprojects (approved):
i don't have this chip yet, but it seems good enough... thanks!

Issue #298, opened 2022-10-17
[solved & testing needed] realloc seems to be broken on T platform, but works on N


openshwprojects:
I have never used realloc on OBK before, but it seems like it's broken on T platform.
realloc is used cJSON_Print, and it hangs device where realloc is called.
As a workaround, I disabled all reallocs calls in cJSON on T platform:
https://github.com/openshwprojects/OpenBK7231T_App/commit/a88140c583a4193b3659da26c1d041b3fc9c1dcb
And here I added some test console commands:
https://github.com/openshwprojects/OpenBK7231T_App/blob/main/src/cmnds/cmd_test.c
**If you're reading this, can you check on your devices the following commands:**
- testMallocFree- should work on T and N
- testRealloc- should hang on T
- testJSON - should work on T and N
**and post here results?**

Extra points for person who can figure out why realloc is broken on T, especially considering that heap4.c from RTOS seems very similar on both T and N.

iprak:
I think I might have into something like this which is why I switched to `cJSON_PrintPreallocated`.

iprak:
Do you know where FIX_REALLOC_ISSUE might be getting set? https://github.com/openshwprojects/OpenBK7231T/blob/eaf4e1a606571b9f13d9c933da61670c712a6c8d/platforms/bk7231t/bk7231t_os/beken378/os/mem_arch.c#L30

openshwprojects:
@iprak , I don't know, but this is a good find. I didn't see it before. Maybe FIX_REALLOC_ISSUE is defined on N but not on T?

iprak:
That function for N seems to be similar but there are other difference in the file itself. But still it might be worth a try.

openshwprojects:
@iprak  os_realloc works on T, but realloc doesn't

openshwprojects:
https://github.com/openshwprojects/OpenBK7231T_App/commit/e77d66292c0cac21226f9591b9ceb8118e973b7e

solved? still, if you can, please check

iprak:
Confirmed: all 3 test commands pass on T.

PR #1563, opened 2025-03-10
Expand testRealloc

niteria:
I was debugging an issue in another pull request and I noticed that `realloc` was behaving weird on my OpenBK7231T device.

I've found #298, so it looked like a known problem.

The old test doesn't hang my device, so extended it and it looks like sometimes reallocated memory gets overwritten.

The test writes some known values into the initially allocated memory and then checks if after every `realloc` the contents of that memory stay the same.

Here's a log from my device:
```
Info:CMD:Realloc difference: rep 90, i 2 j 0 initialra1 230 ra1 242 msg[j] 8 (j % 100) 0
Info:CMD:Realloc difference: rep 90, i 2 j 1 initialra1 230 ra1 242 msg[j] 121 (j % 100) 1
Info:CMD:Realloc difference: rep 90, i 2 j 2 initialra1 230 ra1 242 msg[j] 66 (j % 100) 2
Info:CMD:Realloc difference: rep 90, i 2 j 3 initialra1 230 ra1 242 msg[j] 0 (j % 100) 3
Info:CMD:Realloc difference: rep 90, i 2 j 4 initialra1 230 ra1 242 msg[j] 88 (j % 100) 4
Info:CMD:Realloc difference: rep 90, i 2 j 5 initialra1 230 ra1 242 msg[j] 1 (j % 100) 5
Info:CMD:Realloc difference: rep 90, i 2 j 6 initialra1 230 ra1 242 msg[j] 0 (j % 100) 6
Info:CMD:Realloc difference: rep 90, i 2 j 7 initialra1 230 ra1 242 msg[j] 0 (j % 100) 7
Info:CMD:Realloc has been tested! Total calls 90, reps now 100, reallocBroken 1
```

It's pretty non-deterministic and sometimes takes a while to fail. 
During my debugging I've noticed it's related to timers and one way we use timers is logging.

Issue #298 mentions https://github.com/openshwprojects/OpenBK7231T/blob/eaf4e1a606571b9f13d9c933da61670c712a6c8d/platforms/bk7231t/bk7231t_os/beken378/os/mem_arch.c#L30 which references `platform_is_in_interrupt_context` which I assume is related to timers as well.


niteria:
Looks like `os_realloc` is not broken in the same way, after "1000 calls" it still hasn't detected any memory overwrites.

openshwprojects:
Hmm that's certainly an interesting observation, it won't hurt to include it in the main tree. I'll merge it, thanks

PR #1588, opened 2025-03-30
berry tests

openshwprojects:


openshwprojects:
@niteria I see one issue, which I'm most likely going to fix tomorrow. Your berry's addChangeHandler fires always only once, and it's obvious that's how it will work from code - is uses script thread as a method to get event....
It could be acceptable, but in OBK, we also have addChangeHandler which ironically is ulimited in uses and fires forever until removed...
Is your behaviour of addChangeHandler in berry intended?
I think that I will slightly alter current addChangeHandler handling in berry to allow fifth argument (number of uses), and then alter thread code to count it down and restore waiting if needed.

niteria:
Ah, I should have documented this.

I'm aware that's how it works, I implemented it before `cancel` worked, so it was convenient for it to not repeat. It was also easy to resubscribe from Berry.

It's my personal preference, but it's usually more natural for me to keep resubscribing instead of subscribing and then canceling.

I should have at least picked a different name...

openshwprojects:
No problem, I think I'll attempt to adjust it tomorrow, most likely add one more argument (number of repeats - default infinite?) and try to implement it along with self tests. I just hope that this:

```
void berryResumeClosure(bvm *vm, int closureId) {
	if (!be_getglobal(vm, "resume_closure")) {
		return;
	}
	be_pushint(vm, closureId);
	// call resume_closure(closureId)
	be_call(vm, 1);
}
```
can be called multiple times, in a current form

niteria:
If the current version was renamed to `addChangeHandlerOnce` we could just add this to the standard library:
```
def addChangeHandler(ch, rel, v, cb):
  def cb2(): cb(); addChangeHandlerOnce(ch, rel, v, cb2); end
  addChangeHandlerOnce(ch, rel, v, cb2);
end
```

I'm having a hard time imagining a use-case where number of repeats is different than one or infinity.

niteria:
I believe your solution also works.

openshwprojects:
I've changed old delay ms to javascript-style setTimeout and setInterval, now I am going to look into events, but I am starting to wonder, why do even use old SVM for berry? Maybe it would be cleaner to repliace the "thread list" mechanism in cmd_berry.c and allow us to disable old SVM entirely..

niteria:
> Maybe it would be cleaner to repliace the "thread list" mechanism in cmd_berry.c and allow us to disable old SVM entirely..

Yes, I think it's cleaner. There's no data that's only valid for SVM or for Berry on the thread list, eliminating a class of bugs, one of which I caught in my last PR.

openshwprojects:
Interesting stuff, I didn't really think about using vargs there. Still, it's kinda strange that AI hardcoded type values like 0, 1, 2, instead of using enum. I'll consider porting to that soon

niteria:
Love the HTTP stuff, I wanted to do that at some point.

With the greater degree of programmability that Berry provides, I think I'll be able to use OBK to replace a custom ESP-8266 device I made for Kindle remote page turning using one of the stock Tuya scene switches.

The idea is as follows: 
* OBK stays in AP mode, Kindle (jailbroken, with KOReader) connects to the AP. 
* When the button is pressed on the scene switch a KOReader HTTP endpoint is called to go to next page. 
* If Kindle goes to sleep it notifies the scene switch via HTTP endpoint to go to deep sleep (This is where the new capability is useful!). 
* The scene switch goes to deep sleep by itself if the button is not pressed for 10 minutes. 
* The button wakes it up from deep sleep if it's in deep sleep state.

openshwprojects:
That's nice idea, consider describing it/showing on Elektroda someday.

I have another pending idea - using Berry as PHP. Put HTML file in LFS with <?berry ?> tags and run berry while fetching it. However, I am not sure if it's feasible with Berry, gonna test more

openshwprojects:
I think it could work, but we would need separate berry VM per file run...

niteria:
> I think it could work, but we would need separate berry VM per file run...

Why? If it's about the globals, I believe Berry code can be structured in the way that doesn't use globals

openshwprojects:
I've pushed a working sample:
image
but it seems harder than expected. Now I realize that I can't just parse separately each berry block if I want to have for loops spanning multiple berry blocks hm

openshwprojects:
I think that for the full PHP like experience, the tokenizer would have to parse whole file, and convert HTML blocks outside berry code blocks to something like "static berry prints"/"Echos", only then multi-block ifs and fors would work.

openshwprojects:
I could introduce internal function to berry, like FE (FileEcho) and it would take index of first and number of characters, and replace all string constants with FE(x,y) calls in berry, and convert full file to Berry

openshwprojects:
image

openshwprojects:
it seems to work but without newlines in html cuz they break berry strngs, maybe i just need to escape them

openshwprojects:
Ok it works, now like in PHP:
image
but memory requirements...

openshwprojects:
passes 
image

PR #1596, opened 2025-04-05
tuyaMCU: replace network state magic values

blocktrron:
Replace magic values present in the code with the more descriptive macros.

As these macros are redundant with the comments, drop the repsective comments.


openshwprojects:
seems ok, thank you https://developer.tuya.com/en/docs/iot/tuyacloudlowpoweruniversalserialaccessprotocol?id=K95afs9h4tjjh

openshwprojects (approved):

PR #1597, opened 2025-04-05
tuyaMCU: avoid state machine interfering with reset AP mode

blocktrron:
Don't continue with the state machine if the device is in Open WiFi mode based on a user button press.

This avoids the device shutting down itself when in recovery mode. The MCU will reset the device after 90 seconds automatically.

Communicate to the MCU the device is in configuration mode. This will change the status LED blink pattern to reflect the state.


openshwprojects:
I'm not really able tto test it right now but it seems that it may really improve configuration mode. I'll accept it

openshwprojects (approved):

PR #1608, opened 2025-04-17
add IRRemoteSP

openshwprojects:


openshwprojects:
Are you referring to pulling the "original" IRRemoteESP or would you like to have our own custom IRRemoteESP repo?

NonPIayerCharacter:
Original

openshwprojects:
Hmm I think it may not be viable at the present state of things, our IRRemoteESP has some serious code changes and hacking present already. The main issue is that we can't seem to have an interrupt on "any pin change", only on rising or falling edge. And that's why we have altered IRRemoteESP to rely on high speed timer interrupt, and not on the change... this is not perfect and we often run into timing precision issues with IR but at least it works in some cases

PR #1619, opened 2025-04-27
xr809 toolchain update

divadiow:


divadiow:
9kb file size reduction 
image

divadiow:
7kb reduction with gcc-arm-none-eabi-9-2019-q4-major
image

PR #1626, opened 2025-05-02
xr872 console en

divadiow:


divadiow:
repointed to OpenXR872 repo + new commit

PR #1627, opened 2025-05-02
Add multiple sensors and mutliple GPIOs to driver

MaxineMuster:
Use "HAL usleep" as proposed by @rpv-tomsk with his implementation for Beken

Fix broken HTML (DS1820 driver was forgotten when extending API with "PreState" so all entries where present twice  )

Extended dirver is now enabled for all devices using DS1820 to allow testing.

Can later be turned on with "#define DS1820full 1"


openshwprojects:
that's a long requrested feature, thank you. Is it working, final? No breaking changes?

MaxineMuster:
I will do a review and test on all platforms I own.
The main change (beside extending the code of DS18B20 driver) is using "HAL_Delay_us" with an implementation for BK7231(X) from @rpv-tomsk. There is some test code I can get rid of and we should agree on the "#define" for the new feature:
Actually it's "#define DS1820full" which will only have an effect, if "ENABLE_DRIVER_DS1820" is set.

We could leave it this way with enabling the driver and allow to use the extension (maybe rename it to "DRIVER_DS1820_FULLFEATURE").
Or pretend we have two drivers and use two separate defines like "ENABLE_DRIVER_DS1820_SIMPLE" and "ENABLE_DRIVER_DS1820_FULLFEATURED".

What do you think is better?


I mainly opened this as a PR now because in my local repo at the moment I get strange errors (mainly for Beken Smart-Meter builds, sometimes also IR). The same code is build here without a problem ?!?
It's building fine, but generating OTA fails:```


partition    flash_name       phy addr   size       logic addr size       file
------------ ---------------- ---------- ---------- ---------- ---------- ----------------
bootloader   beken_onchip_crc 0x00000000 68K        0x00000000 64K        bootloader_enc.bin
app          beken_onchip_crc 0x00011000 956K       0x00010000 921359     OpenBK7231T_DS1820_extended_81ee07b75aee_irRemoteESP_enc.bin

image size: 1024 KB
image name: all_1.00.bin

export app partition image: OpenBK7231T_DS1820_extended_81ee07b75aee_irRemoteESP_enc_uart_1.00.bin

time: 0.99s
Good bye!
End Combined
generate ota file
malloc(): corrupted top size
Aborted (core dumped)
make[1]: *** [Makefile:332: build-BK7231] Error 134
make: *** [Makefile:264: OpenBK7231T] Error 2
```

Long story short: I'll report back when I'm "done" ...

NonPIayerCharacter:
I think i figured out why Beken ota generation crashed with
`malloc(): corrupted top size`
Most likely image name is too long. Limit is probably 64 chars, and it crashes when it is longer than that.

MaxineMuster:
> I think i figured out why Beken ota generation crashed with `malloc(): corrupted top size` Most likely image name is too long. Limit is probably 64 chars, and it crashes when it is longer than that.

Ah, thanks a lot, that makes sense!

openshwprojects:
Ah that's why longer pr names seem to crash

MaxineMuster:
No longer needed

PR #1629, opened 2025-05-03
w800 + berry

NonPIayerCharacter:
changes are what previously discussed on the forum
plus:
Unified makefiles for all platforms except esp, ln882, bl602 and default bekens.
OBK_VARIANT is now defined for all platforms bar esp, ln882 and bl602.
Wire library instead of i2c functions for berry.
Flash can be read on W600/W800.
Berry prebuild generation fixed (done in makefile, + fixed rebuilding)
XR806 and XR872 don't suffer from calibration file problems, so removed build retrying.
XR809 and XR872 now use BK7231T gcc, instead of downloading it from external server.


openshwprojects:
Thank you, it seems very good at the first glance. Just to be sure - no breaking changes? I see EasyFlash, but that's still not used for OBK config, correct?

Unified makefile is also great, I would like to have it unified futher.

One concert - why remove retry for XRadio? As far as I know, the partition config calibration retry may be required at random point of time when one of the partition grows too big to fit current layout. So, if in the future we add much more code to, let's say, app partition of XR872, it may require config rearrangement. That's why retry cfg is here. 

Apart from that, if no breaking changes, then I can merge it even today.

NonPIayerCharacter:
EF currently only for flash vars on W800. I thought about adding it to W600 - but it is already constipated in flash.
Unified makefile breaks W600 for some reason, gcc links generic hals (mostly, flash config links correctly, but there are no mentions of hal_wifi_w800 in map file). So i mostly reverted it on W600. I don't know why it happens.

And regarding XRadio, it can be fixed more neatly, by replacing this line:
https://github.com/openshwprojects/OpenXR809/blob/cbe84cb187fbf33a17a581e68709dbba33b41d94/project/project.mk#L266
with this:
```
	($(IMAGE_TOOL) $(IMAGE_TOOL_OPT) -c $(IMAGE_CFG) -o $(IMAGE_NAME).img || true) && \
	$(IMAGE_TOOL) $(IMAGE_TOOL_OPT) -c $(IMAGE_CFG) -o $(IMAGE_NAME).img
```
I don't know how it is done in 806 and 872, but it can probably be solved the same way.

Also added lfs to w600

openshwprojects:
Hm ok, so, if you are disabling retry then add this line in PR. Apart from that, I think I can merge this PR as is today, if there are no breaking changes/broken ota/worse WiFi quality/etc.

MaxineMuster:
@NonPIayerCharacter is OTA working for you with this version?
It won't for me. Using 1629_merge_365dcc027b6f trying to do OTA to last release starts, long until W800 is online again, but the version is unchanged. Need to flash by UART for now.

Starting OTA shows no change in LOG, it simply continues, and after about one minute it's showing many watchdog entries and later reboots:

```
Info:MAIN:Time 290, idle 0/s, free 62616, MQTT 0(18), bWifi 1, secondsWithNoPing 219, socks 3/8 
Info:SENSOR:DS1820[37] - Starting conversion
Info:MAIN:Time 291, idle 0/s, free 62616, MQTT 0(18), bWifi 1, secondsWithNoPing 220, socks 3/8 
Info:SENSOR:DS1820[37] - Temp=22.75
Info:MAIN:Time 292, idle 0/s, free 62616, MQTT 0(18), bWifi 1, secondsWithNoPing 221, socks 3/8 
Info:MAIN:Time 293, idle 0/s, free 62616, MQTT 0(18), bWifi 1, secondsWithNoPing 222, socks 3/8 
Info:MAIN:Time 294, idle 0/s, free 62616, MQTT 0(18), bWifi 1, secondsWithNoPing 223, socks 3/8 
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
WDG IRQ
```

NonPIayerCharacter:
@MaxineMuster 
It works fine for me, successfully downgraded from 1629_merge_365dcc027b6f to 1629_merge_f584e77bd4d9.
```
Debug:OTA:Downloaded 399354 / 411292
Debug:OTA:Downloaded 401400 / 411292
Debug:OTA:Downloaded 403446 / 411292
Debug:OTA:Downloaded 405492 / 411292
Debug:OTA:Downloaded 407538 / 411292
Debug:OTA:Downloaded 409584 / 411292
Debug:OTA:Downloaded 409600 / 411292
Entering initLog()...
Commands registered!
initLog() done!
EasyFlash V4.1.0 is initialize success.
You can get the latest version on https://github.com/armink/EasyFlash .
Main_Init_Before_Delay done
Main_Init_Delay
Main_Init_Delay done
Info:MAIN:Main_Init_Before_Delay
Info:CFG:####### Boot Count 13 #######
Info:CFG:####### HAL_FlashVars_GetBootFailures= 1
Warn:CFG:CFG_InitAndLoad: Correct config has been loaded with 26 changes count.
Error:CMD:no file early.bat err -2
Info:GEN:PIN_SetupPins pins have been set up.
Info:MAIN:Main_Init_Before_Delay done
```

openshwprojects:
Can running drivers affect OTA? I am not sure if we disable drivers tick during  OTA already

NonPIayerCharacter:
Do not merge yet, for some reason on BK7238 situation is like on W600 - generic wifi hals are linked

NonPIayerCharacter:
Solved, the problem was that OBK objs were archived into static library and then linked without --whole-archive.
Pull requests are open in all XR's, W600 and W800

divadiow:
> Strange, I just retried:
> 
> Flashing latest release 1.18.98 OTA to 1629_merge_365dcc027b6f: - OTA messages shown, OTA successfull OTA back to release: - no OTA messages, reboot after approx a minute, no change in version.
> 
> Could I do something wrong?
> 
> Tried with HLK-W801-KIT -V1.1

UART 1.18.98 -> 1629_merge_365dcc027b6f also fine. 1629_merge_365dcc027b6f -> 1.18.98 OTA seems to work but the console output is different, no flash progress before reboot:

`Info:MAIN:Time 118, idle 0/s, free 75232, MQTT 0(7), bWifi 1, secondsWithNoPing 53, socks 2/8 
Info:MAIN:Time 119, idle 0/s, free 75232, MQTT 0(7), bWifi 1, secondsWithNoPing 54, socks 2/8 

towrite 340352 writelen=0
magic_no=2701131679, img_type=1, zip_type=1, signature=0
Network info:
	sta:rssi=63, SSID=test, BSSID=3a:7a:0e:e3:6b:7e, channel=1, encr=WPA2_PSK_CCMP
	IP=192.168.137.82, GW=192.168.137.1, MASK=255.255.255.0, MAC=28:6d:cd:ce:90:78, DNS=192.168.137.1
Info:MQTT:mqtt_host empty, not starting mqtt
Info:MAIN:Time 120, idle 0/s, free 68576, MQTT 0(8), bWifi 1, secondsWithNoPing 55, socks 3/8 
Info:MAIN:Time 121, idle 0/s, free 68472, MQTT 0(8), bWifi 1, secondsWithNoPing 56, socks 3/8 
Info:MAIN:Time 122, idle 0/s, free 68576, MQTT 0(8), bWifi 1, secondsWithNoPing 57, socks 3/8 
Info:MAIN:Time 123, idle 0/s, free 68576, MQTT 0(8), bWifi 1, secondsWithNoPing 58, socks 3/8 

 user task 
Entering initLog()...
Commands registered!
initLog() done!`

with 1.18.98 to OTA other:

```
Downloaded 374692 / 411304
Downloaded 378876 / 411304
Downloaded 383986 / 411304
Downloaded 389096 / 411304
Downloaded 393216 / 411304
Downloaded 398326 / 411304
Downloaded 403436 / 411304
Downloaded 408546 / 411304
etc
```

this is with HLK-W801 ~60cm away from NIC with ext aerial/Windows hotspot (Wifi RSSI: Fair (-63dBm))

NonPIayerCharacter:
Check the log i posted earlier, progress is now reported via ADDLOG_DEBUG, not just in console.
To see progress you must set log level to debug.

NonPIayerCharacter:
Wait a little, i'll soon add berry support to LN and ESP, ass well as generalize cmake source list.

NonPIayerCharacter:
Berry can now be enabled on all platforms.
Added build for BL602 berry
ESP variant builds, 4mb enables berry
Removed configs, now using preprocessor defines
Moved OBK_OTA_EXTENSION to new_common.h
Fixed potential issues with BL602 and ESP regarding weak functions
Get XR806 gcc from github
Fixed some warnings

divadiow:
fantastic

NonPIayerCharacter:
Better partitions for BL602
[partition_cfg_2M.toml.txt](https://github.com/user-attachments/files/20190149/partition_cfg_2M.toml.txt)
Maximum flash utilization. I removed unneeded partitions save one (used for dts, shrinked to 16K), and expanded firmware and ota space to 1M and 584K respectively.
LFS size and address is unchanged, cfg/fv address is unchanged, but its size is expanded to 76K.
Base partition file is from 1.8.0 dev cube.

divadiow:
to 4mb dev board
image
[initialbootafterflash.txt](https://github.com/user-attachments/files/20192806/initialbootafterflash.txt)
image
image


no LFS for me. didn't @giedriuslt's LFS implementation only work for 1.4.8 deprecated type partition layout or something or did you change that too?
image

NonPIayerCharacter:
I'll fix it later, using functions from [here](https://github.com/openshwprojects/OpenBL602/blob/master/components/sys/blmtd/include/bl_mtd.h) to use media partition, and that way it should work on all variants.

NonPIayerCharacter:
Check if it works now, with both old and new partition layout.

NonPIayerCharacter:
Fixed, also if media doesn't exist it now tries to use OTA partition.

divadiow:
**1629_merge_dbe7e1e9752d_berry** insmod_pr1629_partition_cfg_2M.toml - autoexec creates and stays after reboot ✅ then OTA to **1629_merge_dbe7e1e9752d** - autoexec remains in place ✅

**1629_merge_dbe7e1e9752d_berry** old toml - BLDC1.4.8 - autoexec creates and stays after reboot ✅ then OTA to **1629_merge_dbe7e1e9752d** - doesn't like that - BL602 doesn't revive - needs reflash ❌
[otalog.txt](https://github.com/user-attachments/files/20195457/otalog.txt)

**1629_merge_dbe7e1e9752d** old toml - BLDC1.4.8 - autoexec creates and stays after reboot ✅ then OTA to **1629_merge_dbe7e1e9752d_berry** - doesn't like that - BL602 doesn't revive - needs reflash ❌

**1629_merge_dbe7e1e9752d_berry** 1.9.0 2m standard toml - doesn't flash - Exception: Error: fw bin size is overflow with partition table, don't create ota bin (didn't expect this to work - just wanted to see it) - ❌

**1629_merge_dbe7e1e9752d** 1.9.0 2m standard toml - autoexec creates and stays after reboot ✅ then OTA 
 to **1629_merge_dbe7e1e9752d_berry** - autoexec remains in place ✅

**1.18.1 (old SDK)** 1.9.0 2m standard toml - no LFS as expected then OTA to **1629_merge_dbe7e1e9752d_berry** ✅ then create persistent autoexec ✅

NonPIayerCharacter:
OTA to berry with 1.4.8 layout failure is expected, since its fw partition is too small.
But i don't know how you managed to flash berry to 1.4.8 and it booted, since fw bin size is bigger than partition size.
And that is probably why it crashed during ota, because it tried to overwrite fw.
And what about non-berry 1.4.8 to non-berry 1.4.8?

divadiow:
> But i don't know how you managed to flash berry to 1.4.8 and it booted, since fw bin size is bigger than partition size.

Hmm. I'll redo to be sure.

> And what about non-berry 1.4.8 to non-berry 1.4.8?
> And if LFS created in release, does it persist in this pull request?

Will check

divadiow:
> Hmm. I'll redo to be sure.

yes, it does flash OK. 
image

> And what about non-berry 1.4.8 to non-berry 1.4.8?

**1.18.1 (old SDK)** old 2m toml - BLDC1.4.8 - autoexec creates and stays after reboot then OTA to **1629_merge_dbe7e1e9752d** but autoexec is gone after OTA ❌ subsequent autoexec creation works and persists ✅

**1.18.99 (new SDK)** old 2m toml - BLDC1.4.8 - autoexec creates and stays after reboot then OTA to **1629_merge_dbe7e1e9752d** but autoexec is gone after OTA ❌ subsequent autoexec creation works and persists ✅

NonPIayerCharacter:
This should be merge ready then, the only breaking change is that BL602 LFS will be erased on update (1.4.8 only), but will then work on all layout configurations.
To use BL602 with berry, uart flashing will be needed with posted earlier partition layout. And i would recommend using it anyway, because even with berry, there is still about 100k free space left.

openshwprojects:
LFS loss is acceptable, we already had this due to OTA growing on Beken.

So... we need updated BL602 flashing guide? I didn't read whole topic yet, I need to take time to check it

divadiow:
just thinking about users of mhflasher who OTA exploit to OpenBL602, with old SDK and fixed at version **Apr 18 2024 07:21:40 version 1.17.551**, and the MagicHome partition layout noted in https://www.elektroda.com/rtvforum/viewtopic.php?p=21324606#21324606

I flashed 4mb BL602 to **Zengge_RGB_LED_AK001-ZJ2145_33_33_20201024_ZG-BL-IR.bin** then mhflasher OTA to 1.17.551 (no LFS) then OTA to **1629_merge_dbe7e1e9752d**  - OTA worked fine to new SDK and also users get a working LFS after, so seems like a decent update path for mhflasher users to latest BL602 ✅

divadiow:
same success story with **Zengge_GRB_LED_AK001-ZJ21410_33_74_20220225_ZG-BL.bin** and **Zengge_RGB_LED_AK001-ZJ2145_33_33_20201024_ZG-BL-IR.bin**

NonPIayerCharacter:
You are trying an older version.
Try the last one.
The problem was that linker picked generic functions (in .a archive) and was solved with whole-archive flag.
Same problem was in W600.

MaxineMuster:
> You are using an older version.
> Try the last one.

Oh, that's so embarrassing, sorry. I read your post about this linker issue but didn't remember before posting 😳 
Connecting to WiFi and "safe mode" is working as expected!

divadiow:
nice
image
image
1mb custom read to file
[custom_0-1048576.zip](https://github.com/user-attachments/files/20646712/custom_0-1048576.zip)

MaxineMuster:
@NonPIayerCharacter @openshwprojects 

I think I finally found he cause of OTA failing for my W800.

You will remember (or I'll remember you ;-)) that log never showed any sign of the "started" OTA, which is allway displayed when ota function of HTTP starts, so I looked into TCP server code, for it never seemed to start HTTP processing.


My first guess:

The infinite while(1) loop to copy received data to the buffer.

If buffer is to small, it's increased.

But there is no limit for this, hence until recv will indicate that complete buffer was copied, we will increase the buffer.
We will never exit the loop, since we wait for recv to be "empty".

So if the device receives much data very fast (as this seems to be the case with Firefox), we might never be able to empty the recv buffer until we run out of memory.

I made a fix for this "would-be" cause.

```
diff --git a/src/httpserver/http_tcp_server.c b/src/httpserver/http_tcp_server.c
index ac3b21f1..ad2d2802 100644
--- a/src/httpserver/http_tcp_server.c
+++ b/src/httpserver/http_tcp_server.c
@@ -57,6 +57,12 @@ static void tcp_client_thread(beken_thread_arg_t arg)
 	char* buf = NULL;
 	char* reply = NULL;
 	int replyBufferSize = REPLY_BUFFER_SIZE;
+	const int max_grow_count = 2;
+	int grow_count = 0;
+#ifdef FIONREAD
+    int recv_waiting = 0;
+#endif
+
 	//int res;
 	//char reply[8192];
 
@@ -93,8 +99,27 @@ static void tcp_client_thread(beken_thread_arg_t arg)
 		if (received < remaining) {
 			break;
 		}
-		// grow by 1024
+		// grow by 1024 ... but only up to "max_grow_count" times to avoid "limitless" increasing buffer
+
+		if (grow_count >= max_grow_count) {
+			ADDLOG_ERROR(LOG_FEATURE_HTTP, "Max buffer growth reached (%d). Buffer size is %d. Stopping receive.", grow_count, request.receivedLenmax);
+			break;
+		}
+		int oldsize = request.receivedLenmax;
 		request.receivedLenmax += 1024;
+		grow_count++;
+
+		// Query how much data is waiting in the TCP RX buffer, if supported
+#ifdef FIONREAD
+		recv_waiting = 0;
+		if (ioctl(fd, FIONREAD, &recv_waiting) == 0) {
+			ADDLOG_DEBUG(LOG_FEATURE_HTTP, "Growing buffer from %d to %d bytes (data in socket: %d bytes)", oldsize, request.receivedLenmax, recv_waiting);
+		} else {
+			ADDLOG_DEBUG(LOG_FEATURE_HTTP, "Growing buffer from %d to %d bytes (FIONREAD not supported)", oldsize, request.receivedLenmax);
+		}
+#else
+		ADDLOG_DEBUG(LOG_FEATURE_HTTP, "Growing buffer from %d to %d bytes (FIONREAD not available)", oldsize, request.receivedLenmax);
+#endif
 		request.received = (char*)realloc(request.received, request.receivedLenmax+2);
 		if (request.received == NULL) {
 			// no memory
diff --git a/src/httpserver/new_tcp_server.c b/src/httpserver/new_tcp_server.c
index cb586f1b..aa4af0dc 100644
--- a/src/httpserver/new_tcp_server.c
+++ b/src/httpserver/new_tcp_server.c
@@ -41,6 +41,11 @@ static void tcp_client_thread(tcp_thread_t* arg)
 	char* buf = NULL;
 	char* reply = NULL;
 	int replyBufferSize = REPLY_BUFFER_SIZE;
+	const int max_grow_count = 2;
+	int grow_count = 0;
+#ifdef FIONREAD
+	int recv_waiting = 0;
+#endif
 
 	GLOBAL_INT_DECLARATION();
 	GLOBAL_INT_DISABLE();
@@ -74,8 +79,26 @@ static void tcp_client_thread(tcp_thread_t* arg)
 		{
 			break;
 		}
-		// grow by INCOMING_BUFFER_SIZE
+		// grow by INCOMING_BUFFER_SIZE ... but only up to "max_grow_count" times to avoid "limitless" increasing buffer
+		if (grow_count >= max_grow_count) {
+		    ADDLOG_ERROR(LOG_FEATURE_HTTP, "Max buffer growth reached (%d). Buffer size is %d. Stopping receive.", grow_count, request.receivedLenmax);
+		    break;
+		}
+		int oldsize = request.receivedLenmax;
 		request.receivedLenmax += INCOMING_BUFFER_SIZE;
+		grow_count++;
+
+	        // Query how much data is waiting in the TCP RX buffer, if supported
+#ifdef FIONREAD
+	        recv_waiting = 0;
+		if (ioctl(fd, FIONREAD, &recv_waiting) == 0) {
+		    ADDLOG_DEBUG(LOG_FEATURE_HTTP, "Growing buffer from %d to %d bytes (data in socket: %d bytes)", oldsize, request.receivedLenmax, recv_waiting);
+		} else {
+		    ADDLOG_DEBUG(LOG_FEATURE_HTTP, "Growing buffer from %d to %d bytes (FIONREAD not supported)", oldsize, request.receivedLenmax);
+		}
+#else
+	        ADDLOG_DEBUG(LOG_FEATURE_HTTP, "Growing buffer from %d to %d bytes (FIONREAD not available)", oldsize, request.receivedLenmax);
+#endif
 		GLOBAL_INT_DISABLE();
 		request.received = (char*)realloc(request.received, request.receivedLenmax + 2);
 		GLOBAL_INT_RESTORE();

```
Though its working with the "old" W800 code (and this might be an idea to add some fail-safe buffer growing) it didn't help for the new code. It showed "Growing buffer from 2046 to 4094 bytes (FIONREAD not supported)" - and then crashed.

Next idea was: Maybe realloc doesn't work with disabled interrupts?

So I tried:

```
		GLOBAL_INT_DISABLE();
	        ADDLOG_DEBUG(LOG_FEATURE_HTTP, "Before growing buffer ... ");
		bk_printf("Before growing buffer ... ");
		request.received = (char*)realloc(request.received, request.receivedLenmax + 2);
	        ADDLOG_DEBUG(LOG_FEATURE_HTTP, "After growing buffer ... ");
		bk_printf("After growing buffer ... ");
		GLOBAL_INT_RESTORE();
		if(request.received == NULL)

```

Ah, I'm getting nearer, only "Before growing buffer ..." is displayed.

Now I tried:
```

//		GLOBAL_INT_DISABLE();
	        ADDLOG_DEBUG(LOG_FEATURE_HTTP, "Before growing buffer ... ");
		bk_printf("Before growing buffer ... ");
		request.received = (char*)realloc(request.received, request.receivedLenmax + 2);
	        ADDLOG_DEBUG(LOG_FEATURE_HTTP, "After growing buffer ... ");
		bk_printf("After growing buffer ... ");
//		GLOBAL_INT_RESTORE();
		if(request.received == NULL)

```

But still no success - it also will not print the second part.

So, long story short, realloc seems to crash the device, and since it's usually not called, it's not common.

But you can try it yourself, simply make the first buffer (too) small, e.g. 256.
In theory that should be no problem, for it's size should be resized if more data is present.
But it crashes for me on first try for resizing / using realloc.


I found two ways to make OTA possible with all browsers:

Increase the buffer from 2048 to 4096, so no "realloc" is needed

`#define INCOMING_BUFFER_SIZE		4096
`

Or simply don't increase buffer, but start HTTP processing to reduce buffer (the first 2046 bytes will be enough to have the full header and start OTA in this case):

```
		request.receivedLen += received;
//		if(received < remaining)
		if(received <= remaining)	// never increase buffer, but break even if buffer is full and other data waiting
		{

```

But it would be better to find out, why realloc crashes, but that's beyond my possibilities

giedriuslt:
It might be same story as it was on BL602 where malloc was redefined as pvPortMalloc, while realloc was not redefined to anything, so used regular  libc code, which then crashed

NonPIayerCharacter:
It was obvious why realloc didn't work, since it uses libc implementation on os heap. And it probably didn't work before?
There is not much difference between default pseudo-heap_2 and new heap_5.

Added pvPortRealloc for W800, check if it works now.

Added mqtt for XR872, now also uses lwip 2.1.2 and freertos 10.
Readded OTA header for XR872 in a way that should work on 1m flash.
All of it needs checking.

MaxineMuster:
Thanks to @giedriuslt for your hint and especially to @NonPIayerCharacter for all the work and the fix! 
Just to be sure I added some debug output and can confirm, it's working now:

```
Info:MAIN:Time 53, idle 0/s, free 77720, MQTT 0(3), bWifi 1, secondsWithNoPing -1, socks 2/8 
Info:MAIN:Time 54, idle 0/s, free 77720, MQTT 0(3), bWifi 1, secondsWithNoPing -1, socks 2/8 
Info:MAIN:Time 55, idle 0/s, free 77720, MQTT 0(3), bWifi 1, secondsWithNoPing -1, socks 2/8 
Info:MAIN:Time 56, idle 0/s, free 77528, MQTT 0(3), bWifi 1, secondsWithNoPing -1, socks 2/8 
Debug:HTTP:Try growing buffer from 2046 to 4094 bytes ....
Debug:HTTP:... successfully grown buffer from 2046 to 4094 bytes ....
Debug:API:POST to api/ota
Debug:OTA:OTA post len 415044
Debug:OTA:Downloaded 2685 / 415044
Debug:OTA:Downloaded 5757 / 415044
Debug:OTA:Downloaded 7805 / 415044
```


While looking at both implementation of TCP server code, I'm really not sure if I found another possible issue?!?

When looking at the "old" TCP server (src/httpserver/http_tcp_server.c), the code is:

```
		// grow by 1024
		request.receivedLenmax += 1024;
		request.received = (char*)realloc(request.received, request.receivedLenmax+2);
		if (request.received == NULL) {
			// no memory
			return;
		}

```

Simply "returning" after a failed memory (re)allocation seemed like a possible memory leak.

But even the new TCP server might have a potential memory leak, if I'm not mistaken. 
I mean this code:

```
		GLOBAL_INT_DISABLE();
		request.received = (char*)realloc(request.received, request.receivedLenmax + 2);
		GLOBAL_INT_RESTORE();
		if(request.received == NULL)
		{
			// no memory
			goto exit;
		}

```

Though we take care on freeing buffers in "exit":

When doing

`request.received = (char*)realloc(request.received, request.receivedLenmax + 2);`

don't we loose any chance to use/free the memory if it fails? Even more, if we did successfully enlarge the buffer before, "buf" will still address the old/first memory, so what happens, if we "free" this (which was done before by a successful realloc)?

As I said, maybe I simply don't understand, but I would think we should do something like:

```
// if we didn't resize buffer, this is not needed, but if we did, buf is not the same as request.received any more
buf=request.received;
request.received = (char*)realloc(request.received, request.receivedLenmax + 2);
...
```

then, if realloc fails, we still have the pointer to the "old" memory, which should be unchanged in case of realloc fails, and since we "overwrote" the original pointer with NULL, we else would have lost the chance to access/free it?

After that, we could not only proceed to "exit" but (I think) even decide not to bail out and free everything but just do

```
request.received = buf;
```
and "break" so the buffer "up to now" can be processed??

NonPIayerCharacter:
According to C99 if realloc fails, memory is not freed.
But it is better to fix it in a separate pull request, this one is big enough as it is.

MaxineMuster:
Thanks.
If my thought isn't totally wrong that if the failed realloc won't free memory (I also checked in the "pvPortRealloc" code) and we then lost the pointer to the previous memory, I can try a PR for that.
I just wanted to make sure, I'm not totally wrong here ...

divadiow:
> Readded OTA header for XR872 in a way that should work on 1m flash.

1mb XF16 is OK and still accepts console interrupt for UART upgrade too
image
1mb doesn't save config so no MQTT test. 
need to desolder 2mb for SPI flash first before XF16 2mb test

divadiow:
2mb XF16 boots too. everything else the same as 1mb

DeDaMrAzR:
This looks to be a completed PR, will be merged very soon.

PR #1640, opened 2025-05-18
Fix tests on linux

niteria:
The tests were segfaulting on linux, I don't know how they work on Windows, perhaps casting `void*` to `int` happens to work there. Generally it's not portable.

Berry has a dedicated `comptr` type, so let's use that and be portable.

Also makes `run_closure` handle arbitrary number of arguments, see https://berry.readthedocs.io/en/latest/source/en/Chapter-5.html#function-with-variable-number-of-arguments-vararg for docs.

Test Plan:
* CI
* make -f custom.mk, ./build/win_main -runUnitTests 2

### DIFF
--- a/src/berry/be_run.c
+++ b/src/berry/be_run.c
@@ -12,9 +12,9 @@ const char berryPrelude[] =
 	"  return _suspended_closures_idx\n"
 	"end\n"
 	"\n"
-	"def run_closure(idx,x,y)\n"
+	"def run_closure(idx, *args)\n"
 	"  closure = _suspended_closures[idx]\n"
-	"  closure(x,y)\n"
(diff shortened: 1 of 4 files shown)


openshwprojects:
Interesting, is it tested and ready to merge? Did it fail on Linux because int is there 32 bit and pointer was 64 bit?

niteria:
> Did it fail on Linux because int is there 32 bit and pointer was 64 bit?
Yes, `sizeof(int) == 4` and `sizeof(void*) == 8` on my machine.

I haven't tested it on a device, but the linux test doesn't segfault anymore.
If it worked before on a device, it should keep working, this just doesn't make assumptions about `sizeof(void*)` and `sizeof(int)`.

PR #1642, opened 2025-05-19
Enable cse7766 driver on esp

divadiow:


divadiow:
For use with ESP32-C3/CSE7759B combo module

NonPIayerCharacter:
UART is not implemented properly on ESP.
I tried to fix it several times, but there were always missing bytes and hard faults if packets were too big.

divadiow:
Ah bum. Ok

NonPIayerCharacter:
Well, it might work, but there will be a lot of missing packets

divadiow:
will soon find out. will make new forum post with all details

PR #1644, opened 2025-05-20
ln882h CMakeLists.txt - activate sm16703 driver

odi117:
Generate a custom build for ln882h that includes the sm16703 led driver.


NonPIayerCharacter:
It won't work that way, especially without enabling it in obk_config.h. And it's also not implemented in code for LN882H.
But you can check out WS2811 port for LN882H https://github.com/openshwprojects/OpenBK7231T_App/pull/1414

odi117:
> It won't work that way, especially without enabling it in obk_config.h. And it's also not implemented in code for LN882H. But you can check out WS2811 port for LN882H #1414

Ok I appreciate pointing me in that direction

PR #1648, opened 2025-05-23
Domoticz

matth75:
Hello,

I am a french student and for a 2 week school project I added a basic domoticz linking feature to the OpenBeken project. The idea was to implement domoticz support for switches, power strip and power/voltage sensor (if there is available BL09XX, for instance...)

 I added two files for basic jSON formatting as intended by Domoticz standard. in the new src/domoticz folder.

I also modified the main_cfg struct to be able to store the domoticz index associated to the openbeken device (this part assumes there could be as much as 64 different indexes, if there is 64 different channels, but I think it is too much flash storage used for that and it could be improved...). 

I modified the mqtt files to be able to subscribe to the Domoticz/out/<idx> topic in order to receive instructions from domoticz. There are also some changes to the function ChannelOnChanged in order to send new switch state to Domoticz.

I also added a button for configuring Domoticz in the web UI, and some numeric fields to get the domoticz index.

I tested it on some basic switches and on a power strip, which works fine for me, on BK7231N devices. 

I am aware that the code is not optimized since it is my first GitHub project collaboration, but I tried my best ! Let me know if it is interesting for your project ! 

Matthieu

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,5 +15,7 @@
 	"C_Cpp.clang_format_fallbackStyle": "{ BasedOnStyle: LLVM, UseTab: Always, IndentWidth: 4, TabWidth: 4, BreakBeforeBraces: Attach, AllowShortIfStatementsOnASingleLine: false, IndentCaseLabels: false, ColumnLimit: 0, AccessModifierOffset: -4, NamespaceIndentation: All, FixNamespaceComments: false }",
 	"prettier.tabWidth": 4,
 	"prettier.useTabs": true,
-	"prettier.printWidth": 120
+	"prettier.printWidth": 120,
+	"cmake.sourceDirectory": "/home/mrouet/src/localBeken/OpenBK7231T_App/platforms/ESP-IDF",
+	"C_Cpp.dimInactiveRegions": false
 }
--- a/src/domoticz/domoticz.c
+++ b/src/domoticz/domoticz.c
@@ -1,0 +1,217 @@
+#include "../new_common.h"
+#include "../httpserver/http_fns.h"
+#include "../new_pins.h"
+#include "../new_cfg.h"
+#include "../ota/ota.h"
+// Commands register, execution API and cmd tokenizer
+#include "../cmnds/cmd_public.h"
+#include "../driver/drv_tuyaMCU.h"
+#include "../driver/drv_public.h"
+#include "../hal/hal_wifi.h"
+#include "../hal/hal_pins.h"
+#include "../hal/hal_flashVars.h"
+#include "../hal/hal_flashConfig.h"
+#include "../logging/logging.h"
+#include "../devicegroups/deviceGroups_public.h"
+#include "../mqtt/new_mqtt.h"
+#include "../httpserver/hass.h"
+#include "../cJSON/cJSON.h"
+#include <time.h>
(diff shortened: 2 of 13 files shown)


openshwprojects:
Hey, thank you for contribution. That's something that was already requested by users several times, so it will be very helpful for us if we manage to merge it. I will try to help you to get it fully working, so for a start, I'll review PR step by step.

The first impression is good, and the Domoticz handling seems ok and correctly integrated with firmware, but there some miscellanous issues that we need first to resolve.

From a brief glance:
1. You don't need to include .vscode/settings.json in the commit
2. This is not correct because you added new fields before fastConnect data, so fastConnect offset changed, so fastConnect settings will be lost:
image
To add fields correctly to our simple config struct, make sure to respect struct layout
3. Is this a test code? Or is 78 a magic number for RSSI in Domoticz?
	send_domoticz_message(ch, (int) CHANNEL_GetFinalValue(ch), "hey", 78, abs((int) HAL_GetWifiStrength()));
4. Are domoticz indexes 16 bit on purpose? Sorry, I don't know domoticz well....
5. I think it indeed might be better to settle on , let's say, maximum 8 domoticz channels
6. You also got some compile errors, probably need to add missing #ifdef ENABLE_DOMOTICZ

I can help you resolve those issues but I don't know much about domoticz. 

There are two extra things that can be also done (as a features):
- add Domoticz tests to selftests of OBK
- post a guide/report/demonstration of Domoticz on Elektroda in our Smart Home Guides section https://www.elektroda.com/rtvforum/forum517.html

matth75:
Hi, thanks for the first feedback ! I will try to fix the compilation issues, but in my testing process I only used BK7231N, which explains why it failed almost all build tests... 

I think 8 domoticz channels is enough too, since there is one channel per switch.

3. This is indeed a test code, 78 doesn't mean anything. Domoticz is using JSON format and standard messages that I was testing. A typical Domoticz message is JSON and the important fields are idx, nvalue(for int) and svalue (for strings), so the rest is pretty much useless
4. They are 16bit because I thought it would take less space in flash, but if we only keep 8 domoticz channels then they can be int as well

I will try to get it to compile by adding the #ifdef...

If you want to look into Domoticz, its pretty simple : 
1. Create switch in Domoticz UI, which assigns an idx
2. Use the Idx to subscribe the OB device to Domoticz/out/<idx> topic
3. Send JSON message like {"idx":<OBdevice_idx>, "nvalue":"0"} to switch it off ("nvalue":1 to switch it on) to domoticz/in topic
4. Receive JSON message and parse it to gather the idx and the nvalue in order to toggle OB device

This website is ok to understand messages https://piandmore.wordpress.com/2019/02/04/mqtt-out-for-domoticz/ 
I used a Raspberry Pi for Domoticz

matth75:
Hi, I tried adding some #ifdef ENABLE_DOMOTICZ where I could but it does not seem to work...

Is it possible that in some build configurations my new domoticz folder is not seen by the compiler ? Is there some specific configuration files that I have to change in order for this compile to work on all different chips ?

JanKiel2000:
Any progress? I'm also interested in integration with Domoticz.

matth75:
Hi, sorry I stopped working on this since it was only a 2 weeks school project... Good luck with that !

PR #1661, opened 2025-05-31
Update http_fns.c

openshwprojects:


openshwprojects:
@niteria  so this will show a memory leak now? Let's see...

niteria:
This particular leak is silenced with `Test_HassDiscovery`, so you could try removing that from `platforms/linux/asan.suppressions`, but perhaps on this ubuntu image ASAN also needs to be enabled by an ENV var: `export ASAN_OPTIONS=verbosity=1:halt_on_error=1`

openshwprojects:
It seems it has triggered some kinf of strange error on beken..

niteria:
I've removed the suppression and it fails as espected: https://github.com/niteria/OpenBK7231T_App/actions/runs/15364237395/job/43235091182

niteria:
BTW your last commit is also covered by `Test_HassDiscovery` from the suppression file, `Test_HassDiscovery` is on the call stack, so it ignores it.

Weird that it's failing the other tests while the `main` built.

openshwprojects:
it works, vry nice, i gonna try to fix other leaks and i will also post a tutorial for that on forum so other contributors can benefit from leaks detection

PR #1662, opened 2025-05-31
Add raw IR dump log

Razzile:


divadiow:
did this work? what was the use case?

please close if this was not intended for merge consideration

PR #1665, opened 2025-05-31
Try to fix mem leaks in Win32

openshwprojects:


openshwprojects:
@niteria  what has happened here? No stack trace suddenly. Too long execution/too long log?
image

openshwprojects:
Intresting, I was working under assumption that this works like CRT_DEBUG_FLAG of MSVC (whatever it was called), so it dumps leaks at the main() exit.... ok one more leak solved

PR #1664, opened 2025-05-31
Run linux tests with UBSAN

niteria:
Undefined Behavior Sanitizer (UBSAN) [1] is a similar tool to ASAN, useful for catching bugs related to undefined behavior.

All the problems it detects seem related to the bitshifts related to flags. Here's a sample:

```
src/new_pins.c:2437:8: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
	#0 0x55a1b51011c1 in PIN_get_Relay_PWM_Count src/new_pins.c:2437
	#1 0x55a1b500389b in NewLED_InitCommands src/cmnds/cmd_newLEDDriver.c:1627
	#2 0x55a1b5150208 in Main_Init_BeforeDelay_Unsafe src/user_main.c:1151
	#3 0x55a1b51506e7 in Main_Init_Before_Delay src/user_main.c:1290
	#4 0x55a1b5150a05 in Main_Init src/user_main.c:1428
	#5 0x55a1b515902c in SIM_ClearOBK src/win_main.c:166
	#6 0x55a1b515a0fd in main src/win_main.c:534
	#7 0x7f1acfdabd8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
	#8 0x7f1acfdabe3f in __libc_start_main_impl ../csu/libc-start.c:392
	#9 0x55a1b4fd8eb4 in _start (/home/niteria/tmp/broken-realloc/OpenBK7231T_App/build/win_main+0x10ceb4)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/new_pins.c:2437:8 in
```

```
src/new_cfg.c:591:9: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
	#0 0x5629450eb803 in CFG_HasFlag src/new_cfg.c:591
	#1 0x562944ffdb53 in CMD_Init_Early src/cmnds/cmd_main.c:1007
	#2 0x562945153230 in Main_Init_BeforeDelay_Unsafe src/user_main.c:1162
	#3 0x5629451536e7 in Main_Init_Before_Delay src/user_main.c:1290
	#4 0x562945153a05 in Main_Init src/user_main.c:1428
	#5 0x56294515c02c in SIM_ClearOBK src/win_main.c:166
	#6 0x56294515d0fd in main src/win_main.c:534
	#7 0x7f9f5e12bd8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
[...]

PR #1651, opened 2025-05-25
Ability to run linux tests with AddressSanitizer

niteria:
This makes it possible to compile and run the tests with Address Sanitizer (ASAN) [1].

The command is:
```
make -f custom.mk clean; ASAN=1 make -f custom.mk && ./build/win_main -runUnitTests 2
```

Building without ASAN works without change.

Address Sanitizer saves a lot of time debugging memory issues, for example for a bug where we index in an array with negative index it gives this nice error message:
```
=================================================================
==2532658==ERROR: AddressSanitizer: global-buffer-overflow on address 0x5642389c5e37 at pc 0x564237e67492 bp 0x7fff17726e50 sp 0x7fff17726e40
READ of size 1 at 0x5642389c5e37 thread T0
    #0 0x564237e67491 in CHANNEL_GetType src/new_cfg.c:427
    #1 0x564237dfc933 in TuyaMCU_ApplyMapping src/driver/drv_tuyaMCU.c:1126
    #2 0x564237dff92f in TuyaMCU_ParseStateMessage src/driver/drv_tuyaMCU.c:1632
    #3 0x564237e019fa in TuyaMCU_ProcessIncoming src/driver/drv_tuyaMCU.c:1977
    #4 0x564237e027ee in TuyaMCU_RunReceive src/driver/drv_tuyaMCU.c:2178
    #5 0x564237e02dc2 in TuyaMCU_RunFrame src/driver/drv_tuyaMCU.c:2335
    #6 0x564237dd2ce6 in DRV_RunQuickTick src/driver/drv_main.c:536
    #7 0x564237ec4037 in QuickTick src/user_main.c:946
    #8 0x564237eca8a5 in Sim_RunFrame src/win_main.c:106
    #9 0x564237eca993 in Sim_RunFrames src/win_main.c:134
    #10 0x564237ebce4a in Test_TuyaMCU_Basic src/selftest/selftest_tuyaMCU.c:369
    #11 0x564237ecab1f in Win_DoUnitTests src/win_main.c:194
    #12 0x564237ecba57 in main src/win_main.c:535
    #13 0x7fb55312ed8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #14 0x7fb55312ee3f in __libc_start_main_impl ../csu/libc-start.c:392
    #15 0x564237cb16b4 in _start (/home/niteria/tmp/broken-realloc/OpenBK7231T_App/build/win_main+0xcf6b4)

[...]

PR #1672, opened 2025-06-10
Try fixing possible memory leaks in TCP servers

MaxineMuster:
As a result from some tests in #1629:

I think there are possible memory leaks in TCP server code:
    in case realloc fails, regular TCP server will simply "return", leaving all allocated buffers untouched

For both servers: If realloc is used, "buf" pointer is not changed.
    if later "free(buf)" is called, that will probably be invalid (the memory was already freed during a successful "realloc")
    if realloc fails, we will lose pointer to the old (unchanged, in this case) buffer, which is overwritten by NULL 

Some more theoretic problem: 
    both servers will in theory increase buffers memory unlimited, until all data from recv fits.
    That could possibly consume all memory. 

So here's my try to address these possible issues.

Tested only with W800 yet

As a test it uses low sizes for buffers to "force" resizing buffers (512 bytes as start, maximum is 20248)

There are some minor changes to allow new_tcp_server on W800.

In "new_tcp_server" I also needed to lower stacksize if memory is low - W800 else crashed from time to time
   directly after creating a  TCP Client thread
This will not happen on "regular" TCP server
   
This first commit will build "regular" TCP server on W800, next will be with new_tcp_server.

EDIT: And maybe someone knows, why the same code, which is successfully compiled here (all checks o.k.) fails for some "Beken" variants (powerMetering, IRremote..) im my repo with something like

```
 export app partition image: OpenBK7231T_TCP_server_memoryleak_3e1852758489_tuyaMCU_enc_uart_1.00.bin

time: 0.94s
Good bye!
End Combined
generate ota file
malloc(): corrupted top size
Aborted (core dumped)
make[1]: *** [Makefile:332: build-BK7231] Error 134
```

### DIFF
--- a/src/obk_config.h
+++ b/src/obk_config.h
@@ -72,7 +72,7 @@
 #define OBK_OTA_EXTENSION 		".img"
 #define ENABLE_I2C					1
 #define ENABLE_OBK_SCRIPTING			1
-//#define NEW_TCP_SERVER				1
+#define NEW_TCP_SERVER				1
 
 
 #elif WINDOWS


divadiow:
> I think there are possible memory leaks in TCP server code:

probably a dumb question, but does/would @niteria leak checks pick this up? 

https://github.com/openshwprojects/OpenBK7231T_App/pull/1664
https://github.com/openshwprojects/OpenBK7231T_App/pull/1651

niteria:
> probably a dumb question, but does/would @niteria leak checks pick this up?

If it's not triggered by any of the simulator tests, then ASAN wouldn't catch it. If possible, it would be nice to add a test that triggers it.

MaxineMuster:
Fixed in #1990

PR #1673, opened 2025-06-12
pins for XR872

divadiow:


NonPIayerCharacter:
#elif PLATFORM_XR872, instead of doing it in XR809
Also separate XR809 and XR872 pin definitions by ifdefs
And maybe sort XR809 pins?

divadiow:
probably easy if you know what you're doing ;)

divadiow:
i see all this stuff needs splitting out for xr872/xr806 :/

https://github.com/openshwprojects/OpenBK7231T_App/tree/main/src/hal

PR #1687, opened 2025-06-23
autoexecExamples.md: Use C++ higlighting

C0rn3j:
Also unifies code block not having extra lines, it was all over the place.

This makes the docs infinitely more readable.


openshwprojects:
It;s a nice idea, but it should be done in getcommands.js - the md file is autogenerated

C0rn3j:
That should do it.

Hopefully the linter step isn't too annoying, I see that most of the codebase uses dynamic indents already (spaces were converted to tabs) and it unified the code style.

I ran the script and pushed just the autoexec doc from it, but there's quite a few doc changes not relevant to this PR hanging in other files, so I just left those uncommited.

openshwprojects:
Nice, btw, I think there is a bug in current (main repository) getcommands.js . It tends to remove some entries from code, from drivers, etc

PR #1707, opened 2025-07-01
Update ESP-IDF to v5.5-beta1 and add minimal ESP8266 support

NonPIayerCharacter:
Added BL0937 support
Added ESP32-C5 and ESP32-C61 support

ESP8266 both 1MB and 2MB builds
ESP8266 is very problematic. Only about 25-35k heap available.
OTA while successful, is not working.
After setting up wifi, it has to be reset to connect, since somehow it does not start connecting and WIFI_STA_DISCONNECTED is not called.
PWM had to be disabled on ESP8266, it crashes.
Largely untested I only tested pins, and OTA, that's it.


divadiow:
Wow. 👍

DeDaMrAzR:
New beast in town 😁 Module in test is default ESP8266 module with 4Mb 

Let's test and see what is working and what is going on. First impressions:

 - Flashing goes without any problems
 - AP is available right after and accessible
 - upon connecting to network my DHCP did show tasmota_<last 3 MAC digits> and it got an IP but the device did not reset nor connect, log still showed 0.0.0.0 for IP, after manual reboot it worked.
 - LED on GPIO2 works
 - NTP works

Really impressive work!! Device feels super responsive and can't wait more tests and development on this!

Bravo!

openshwprojects (approved):

PR #1716, opened 2025-07-08
IR driver for RTL87X0C

divadiow:
i know about the timer port requirement. just seeing what this does anyway


NonPIayerCharacter:
First you would need to add `SRC_CPP += $(OBKM_SRC_CXX)`  in OpenBeken.mk
Then rewrite driver code to use HALs, instead of beken specific gpio and other functions.

PR #1723, opened 2025-07-17
hide channel for BL0937_SEL_n

divadiow:


divadiow:
image

divadiow:
image

divadiow:
oh. there are more?

MaxineMuster:
Maybe we can look again at the whole defnition of pins and roles?
I once proposed in #1378 to change the definition so all is in one place.

This way we "know" about the number of channels used for a role ...

No, PR is not thoroughly tested up to now, but I wanted to get some feedback on the general idea (and later forgott about ist ...)

divadiow:
yeh cool. I noticed the other DAT/CLK ones shortly after.. and, yeh..

MaxineMuster:
Since PR was quite old, I had to add the "newer" roles PWM_ScriptOnly(_n) but now it compiles ;-).

It should fix the issue with the channels?!? (Simulator will show no channel for BL0937_SEL_n)
Or are there more? Any "DAT/CLK ones"?

divadiow:
nice.

well, potentially these, but I'm unsure on some.  eg. why do some pairs only have 1 hidden?

image

MaxineMuster:
IIRC I only copied the mimic of "int PIN_IOR_NofChan(int test){" from new_pins.c
And, if this role has no "special treatment", the default is returned: 1 channel.
So it's quite possible, roles are missing the correct number of channels.

(Note to myself: Check if still needed and if change that function, too, now we defined the channels per role.
Its IORoles[i].NumberOfChannels for now.)

I know e.g. for SHT3X_SDA the channels are used for temp and hum.

So we will probably need to check the drivers, if they use the channels to be sure ...

E.G. drv_gn6932.c (aka drv_tm_gn_display_shared.c) seems NOT to use any channel.
We should check with p.kaczmarek2/openshwprojects to be sure.

divadiow:
SHT - ah yes. memory fails.

divadiow:
> We should check with p.kaczmarek2/openshwprojects to be sure.

@openshwprojects

MaxineMuster:
If we do a rework, we might add the usage of a channel.

The data is sent in JS as an array of pairs ["role-name", <# of channels>],
We could easyly extend this by the "usage", to give some examples:

```
["DHT11",2,"Temperature","Humidity"],
["DHT12",2,"Temperature","Humidity"],
["DHT21",2,"Temperature","Humidity"],
["DHT22",2,"Temperature","Humidity"],
["DS1820_IO",1,"Temperature of single DS1820"],

```

Then we could 
- make sure, every channel not hidden in fact has a valid meaning
- give some hint (e.g. display this as "title"):

 

https://github.com/user-attachments/assets/962f896d-79e6-4278-8c63-2d5d78d2bb3f

EDIT:
This is just a quick change to the saved source of the config page:

Extended the array as above
extended the function " hide_show()" to display a title, if set
(the two lines with er.title / ee.title are the only change to the JS source)

```
function hide_show() {
	n=this.name;
	er=getElement('r'+n);
	ee=getElement('e'+n);
	ch=r[this.value][1];
	er.disabled = (ch<1);
	er.style.display= ch<1 ? 'none' : 'inline';
	er.title=(t1=r[this.value][2])? t1:"";
	ee.disabled = (ch<2);
	ee.title=(t2=r[this.value][3])? t2:'';
	ee.style.display= ch<2 ? 'none' : 'inline';
}
```


EDIT2:

shorter JS code:
```
cf=(e,v,t)=>{e.disabled=v;e.style.display=v?'none':'inline';e.title=t;};
function hide_show() {
	n=this.name;
	rv=r[this.value];
	c=rv[1];
	cf(getElement('r'+n),(c<1), rv[2]??'');
	cf(getElement('e'+n),(c<2), rv[3]??'');
}

```

divadiow:
Oh cool!

PR #1725, opened 2025-07-19
CHANNEL_IsInUse() report only used channels

MaxineMuster:
Actually, CHANNEL_IsInUse() will report all channel values. Change to only report used channels up to PIN_IOR_NofChan() for this role

### DIFF
--- a/src/new_pins.c
+++ b/src/new_pins.c
@@ -1727,10 +1727,11 @@ bool CHANNEL_IsInUse(int ch) {
 
 	for (i = 0; i < PLATFORM_GPIO_MAX; i++) {
 		if (g_cfg.pins.roles[i] != IOR_None) {
-			if (g_cfg.pins.channels[i] == ch) {
+			int NofC=PIN_IOR_NofChan(g_cfg.pins.roles[i]);
+			if (NofC>=1 && g_cfg.pins.channels[i] == ch) {
 				return true;
 			}
-			if (g_cfg.pins.channels2[i] == ch) {
+			if (NofC>=2 && g_cfg.pins.channels2[i] == ch) {
 				return true;
 			}
 		}


MaxineMuster:
If accepted, I would propose an additional command "CHANNEL_PinUsingCh(int ch)" to report the pin a channel is used (or the "type" if not assigned to a channel by returning "type" enum + PLATFORM_GPIO_MAX to distinguish.
To be discussed...

Could be useful if we plan some kind of check for double assignment of channels, so wen can report 
"Channel x is allready used on Pin n"

```
// will return -1 if no use is found
// will return (first) pin using the channel or, 
// in case channel is used by defining it's type, "Type+PLATFORM_GPIO_MAX" 
// 
int CHANNEL_PinUsingCh(int ch) {
	int i;
	for (i = 0; i < PLATFORM_GPIO_MAX; i++) {
		if (g_cfg.pins.roles[i] != IOR_None) {
			int NofC=PIN_IOR_NofChan(g_cfg.pins.roles[i]);
			if (NofC>=1 && g_cfg.pins.channels[i] == ch) {
				return i;
			}
			if (NofC==2 && g_cfg.pins.channels2[i] == ch) {
				return i;
			}
		}
	}
#if (ENABLE_DRIVER_DS1820_FULL)
#include "driver/drv_ds1820_full.h"
	int x=ds18b20_channel_used_on_pin(ch);
	if (x >=0) return x;
#endif
	if (g_cfg.pins.channelTypes[ch] != ChType_Default) {
		return (int)g_cfg.pins.channelTypes[ch]+(int)PLATFORM_GPIO_MAX;
	}

	return -1;
}

bool CHANNEL_IsInUse(int ch) {
	return (CHANNEL_PinUsingCh(ch)>-1);
}

```

openshwprojects (approved):

PR #1729, opened 2025-07-19
Split NTP and clock functions

MaxineMuster:
To allow other sources than NTP for setting the clock, the general approach needs to be changing.

Here's my actual proposal.


MaxineMuster:
No longer needed

PR #1736, opened 2025-07-22
SM16703P: learn to handle RGBCW, e.g. more than 3 channels per pixel.

slo617:
As the title says, this adds support for SM16703P compatible ICs with more than 3 channel.
For example
 - 4 channel: RGBW
 - 5 channel: RGBCW
 - Also more or less channels are possible, if such hardware exists, e.g. RGB with 2 warm-white channel (of same temperature)
 - Also crazy combination would be possible, like a white and a red channel, but no other channels (if such hardware exists)

I did it as generic as possible.
Modification should be fully backward compatible.

`SM16703P_Init` command has been modified to support also cold-white and warm-white channel.
Default is still "RGB", e.g. `SM16703P_Init 20` is same as `SM16703P_Init 20 RGB` and enables 20 RGB-ICs with "RGB" channel order - as before.
To use OpenBeken with 5 channel RGB-ICs call something like: `SM16703P_Init 20 RGBCW`

I did these modification to solve [https://www.elektroda.com/rtvforum/topic4062886.html](https://www.elektroda.com/rtvforum/topic4062886.html). I got myself the same string lights which are working fine now.

Flag 4 is still needed to enable RGBCW controller for RGBICs. (No changes on that)

This is my first contribution on OpenBeken. Please kindly let me know if something should be done in a different way.

### DIFF
--- a/src/cmnds/cmd_newLEDDriver.c
+++ b/src/cmnds/cmd_newLEDDriver.c
@@ -660,7 +660,7 @@ void apply_smart_light() {
 #endif
 #if	ENABLE_DRIVER_SM16703P
 	if (pixel_count > 0 && (g_lightMode != Light_Anim || g_lightEnableAll == 0)) {
-		SM16703P_setAllPixels(finalColors[0], finalColors[1], finalColors[2]);
+		SM16703P_setAllPixels(finalColors[0], finalColors[1], finalColors[2], finalColors[3], finalColors[4]);
 		SM16703P_Show();
 	}
 #endif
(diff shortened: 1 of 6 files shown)


openshwprojects:
Nice, but shouldn't const enum ColorChannel default_color_channel_order[3] have size 5?

slo617:
Hi @openshwprojects , thanks for your reply.
`const enum ColorChannel default_color_channel_order[3]` holds the default value only, it is not changed.
We will allocate memory on the heap with `malloc()` if a different channel order is passed over with the init command.
That is more flexible and also allows more (or less) then 5 channels, e.g. "RGBCWW" - with 2 warm white channels.
For sure, support for 6 or more channels is not needed for now, but it comes for free and (even more important) we do not need to care about any special cases that might come up.

PR #1738, opened 2025-07-25
Fix spacing in main page

giedriuslt:


openshwprojects:
thx, btw, if you can, cause if remember correctly, you know bl602 somewhat well, can you also help us with getting bl602 1mb builds working and official?

PR #1753, opened 2025-08-09
add "hass_init_textField_info()" to fix ESP builds

MaxineMuster:
ESP builds with "-Werror=implicit-function-declaration", so we get

 error: implicit declaration of function 'hass_init_textField_info';


openshwprojects:
huh thx, i was sure... nevermind

PR #1761, opened 2025-08-12
mega spellings fix

divadiow:
I appreciate this is mostly pointless because a lot are comments.

There are a couple of log out and faq type adjustments. I've mostly shied away from touching anything code so as not to break anything. I'll close if not bothered


divadiow:
closing to revise

PR #1787, opened 2025-09-03
enable bridge on BL602

divadiow:
https://www.elektroda.com/rtvforum/topic4136806.html


goblinmaks:
Why it is unmerged ?

divadiow:
I'll re-open but it'll be up to @openshwprojects if it should be merged.

Issue #1822, opened 2025-10-02
Wrong color from LEDs, broken gamma algorithm


jcam:
**Describe the bug**
When I use the Home Assistant built in color picker, very often the bulbs turn the wrong color

**Firmware:**
- Versions tried: 1.18.181, 1.18.109 and 1.17.736
- Device: Daybetter 800lm RGB+2700-6500K 
- Chip/model: wb2l / BK7231T
- Device config: bp5758d on pins 7 and 8, with BP5758D_Map 2 1 0 3 4

**To Reproduce**
1. Add a bulb to home assistant
2. Open the light entity for the device, choose a color on the color picker
3. Observe the color
4. Use MQTT explorer to see what color was selected, and compare that color to the light

**Screenshots**
Here I selected orange:

<img width="520" height="589" alt="Image" src="https://github.com/user-attachments/assets/879f9516-efc5-440d-9f9f-ca4d6ccee05d" />

Here's the color that was sent through MQTT:

<img width="316" height="256" alt="Image" src="https://github.com/user-attachments/assets/2c2ebc28-9be4-4884-9b40-69bc74be54c2" />

This is what color FF8402 should be:
https://www.color-hex.com/color/ff8402

<img width="455" height="310" alt="Image" src="https://github.com/user-attachments/assets/fd067539-2dce-4635-9a05-d74fea579498" />

This is what the light is doing. The light on the left is using ESPHome (it is dining-room-1), the light on the right is using OpenBK 1.18.181. Both are showing the same color, in theory.
If it isn't clear from the image, because photos look different on different screens, the left ESPHome bulb looks like the same orange as FF8402, while the OpenBK bulb looks greenish yellow.

<img width="1243" height="818" alt="Image" src="https://github.com/user-attachments/assets/10965131-6d66-4847-b692-4ae8fcef3984" />

**Additional context**
I thought maybe it was just the way these bulbs were until I tried ESPHome on this one bulb, and there the colors work perfectly. Unfortunately the startup time for ESPHome is terrible compared to OpenBK so I'd much rather stick to this project!

jcam:
After a WHOLE LOT of digging and trying to understand what was going on, I have finally figured out that the problem is not with the BP5758D implementation, instead it is with the gamma correction.
Gamma should be applied to the final color value, not just the brightness. It shouldn't scale all three colors identically, it should scale them individually.

I'll work on a PR and some examples, but you can see how ESP calculates gamma here:
[...]

PR #1825, opened 2025-10-05
Fix gamma correction

jcam:
The previous version of the gamma correction function didn't correct the gamma of the individual color channels, instead it was correcting the gamma only on the overall brightness.

This new algorithm, which borrows some logic from the ESPHome gamma implementation, resolves this and now colors on the bulbs match the selected colors.


jcam:
As noted in #1822 this currently does not utilize rgb_bright_min or cw_bright_min.

openshwprojects (approved):

PR #1853, opened 2025-11-03
Add support for Frequency variable in AddChangeHandler

kcstonacek:
Adding support for the Frequency variables in AddChangeHandler.  This can be useful for frequency based load control, like inverter based microgrids.

Because change handlers are compared as integers, the frequency is referenced as *100 (similar to the current variable).  e.g. for 52.34hz: 
AddChangeHandler Frequency > 5234 setChannel 1 0 

Also, added AddChangeHandler Current selftests and associated docs for Current variable referenced as *1000 while I was there.

Some points for discussion:
- is this worth adding float change handler comparators for this?  I don't think so.  While awkward, it's functional and documented enough that those using this will understand what must be done.
- should this be behind a flag?  maybe, but frequency is already stored and published, so might as well make it available as a variable
- should the order of VCPPublishThreshold be changed?  I believe it's not working properly for the optional parameters, so I see this as a non-breaking change.  I think the real question is: should we correct the arg index for energy and power (with default) so they can be properly changed with VCPPublishThreshold?

### DIFF
--- a/src/selftest/selftest_energyMeter.c
+++ b/src/selftest/selftest_energyMeter.c
@@ -21,17 +21,17 @@ void Test_EnergyMeter_Basic() {
 	SELFTEST_ASSERT_EXPRESSION("$current", 0);
 	SELFTEST_ASSERT_EXPRESSION("$power", 0);
 	SELFTEST_ASSERT_EXPRESSION("$frequency", 0);
-	CMD_ExecuteCommand("SetupTestPower 230 0.26 60 50.10 0", 0);
+	CMD_ExecuteCommand("SetupTestPower 230 0.26 60 50.00 0", 0);
 	Sim_RunSeconds(10, false);
 	SELFTEST_ASSERT_EXPRESSION("$voltage", 230);
 	SELFTEST_ASSERT_EXPRESSION("$current", 0.26f);
 	SELFTEST_ASSERT_EXPRESSION("$power", 60);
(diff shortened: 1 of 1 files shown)


openshwprojects:
I need a little change before merge. You are breaking current people configs by adding VCPPublishThreshold  Frequency arg as not the last argument. Can you make this Pr non-breaking?

kcstonacek:
@openshwprojects

Sure, not a problem.  However, I don't think this is actually breaking anything (unless I'm not understanding the use of Tokenizers) - take a look at the Tokenizer_GetArgFloat:

https://github.com/openshwprojects/OpenBK7231T_App/blob/36218fd38e0f513d5aae8ba452d4d8baa835b34f/src/driver/drv_bl_shared.c#L534-L538

to me this looks like the last two args use the same values as OBK_POWER .  Shall I correct the Tokenizer indexes, or update the docs to reflect this?

openshwprojects:
strange, they are reading same argument 3 times. Well, the most important priority is to keep commands as before pr, so maybe just keep using this arg 2 as generic power delta and use next argument (the first not used) for your feature, with a correect default option. 

Long story short - do not break behaviour of existing configs, so we don't confuse users

divadiow:
maybe an issue since this PR on BL602 with BL0937... https://www.elektroda.com/rtvforum/viewtopic.php?p=21794416#21794416

PR #1854, opened 2025-11-04
BL0937 accuracy/res improvement by slow down readout

listen-obk:
adopt measure intervals to output frequency of bl0937 for more resolution  (~1,7W at 1Hz)


divadiow:
This looks interesting. What's the effect of OTA updating to this version of you've already got BL0937 configured? Does it breaks stats or keep working as before? 

Have you tried it on multiple platforms?

listen-obk:
Currently I test only on new Tuya plugs with EC6600. OTA from .205 (and to .207) main version worked well but new calibration is required. Because I wanted to see frequency of BL0937 in log and due to restrictions of Scale (raw values in int) I added a factor of 100 for V and 1000 for I and P. There might be better solutions for that.

I did not check impact on stats yet - I would expect it will work well.

PR #1866, opened 2025-11-19
test

openshwprojects:


openshwprojects:
berry time and openweathermap access with self tests

PR #1893, opened 2025-12-12
Custom build for XR806 based TRV801W thermostat

jkrzyszt:
This device seems to require DCDC power configuration.


divadiow:
DCDC variant of OpenXR806 is now available in general release

PR #1901, opened 2025-12-14
Restrict Semantic Release to specific repository owner

xjikka:
To allow run workflow on forks


xjikka:
nobody needs it, I'm cancelling the PR

PR #1911, opened 2025-12-19
Driver for DCF77 radio controlled clock

MaxineMuster:
Quite common in Germany are Radio Controlled Clocks, using the DCF77 broadcasted time signal as time source.
Bildschirmfoto vom 2025-12-19 19-07-30
Since we now have a flexible clock, we can use this as clock source, too.

Starting with "startdriver DCF77"
optional arguments are
"fall"             per default, a rising edge will mark the "bits", if you need a falling edge (e.g. if using a opto-isolator), use this argument
"httpstat"     add some details from driver on main page
Bildschirmfoto vom 2025-12-19 19-25-45

Arguments to "tune" reception (usually "0" is 100ms, "1" is 200 ms - you can change the "breaking" point with the below values (actual code uses 180 as default for both values)
"end0=X"     set max duration of 0 bit to X ms
"start1=X"    set min duration of 1 bit to X ms

This is some debug output:

Bildschirmfoto vom 2025-12-19 18-09-33

It might also be used with an opto-isolator for devices with main on the IO pins:

This is my test with a BK7231N switch (powered by a PoE power supply with 48V) and the receiver input on the "switch" (since opto coupler inverts the signal, I used it with "fall" argument). I'm using a common PC-817 and a 220 ohm resistor for 3.3V for DCF77 device:

IMG_20251219_182545202~2

### DIFF
--- a/src/obk_config.h
+++ b/src/obk_config.h
@@ -88,7 +88,7 @@
 //#define ENABLE_DRIVER_SM16703P					1
 //#define ENABLE_DRIVER_PIXELANIM					1
 #undef ENABLE_HTTP_MAC
-#define ENABLE_DRIVER_DCF77					1
+//#define ENABLE_DRIVER_DCF77					1
 
 #elif PLATFORM_W800
 
@@ -121,7 +121,7 @@
 #define ENABLE_DRIVER_PIXELANIM					1
 #define ENABLE_DRIVER_TINYIR_NEC				1
 // allow testing, remove before merging
-#define ENABLE_DRIVER_DCF77					1
+//#define ENABLE_DRIVER_DCF77					1
 
 #elif WINDOWS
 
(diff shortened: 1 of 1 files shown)


MaxineMuster:
Updated code.

Changed from own "IORole" to giving pin as argument:

startDriver DCF77 \<pin\> ...

Further optional arguments are

- "fall" per default, a rising edge will mark the "bits", 
&emsp;&emsp;&emsp; if you need a falling edge (e.g. if using an opto-isolator), use this argument
- "httpstat"&emsp;&emsp;&emsp;&emsp;&emsp;add some details from driver on main page
- "end0=\<value\>"&emsp;&emsp;set max duration of 0 bit to \<value\> ms
- "start1=\<value\>"&emsp;&emsp;set min duration of 1 bit to \<value\> ms

grafik

PR #1914, opened 2025-12-22
TuyaMCU: Add version 3 data units support

protectivedad:
This is a work in progress. I have a ZY-D02 door sensor. It uses the TuyaMCU and the T1-3S Tuya wifi model.

It sends the dpId information as:
```
Received by WiFi module:
55 AA	03		34		00 0E	0B01000101010101016501000101	BE	
HEADER	VER=03	Unk		LEN		0B01000101010101016501000101	CHK	

Received by WiFi module:
55 AA	03		34		00 0E	0B01000101010101016604000102	C3	
HEADER	VER=03	Unk		LEN		0B01000101010101016604000102	CHK	
```

I'm updating the TuyaMCU driver to process this information.

### DIFF
--- a/src/driver/drv_tuyaMCU.c
+++ b/src/driver/drv_tuyaMCU.c
@@ -250,7 +250,7 @@ enum TuyaMCUV0State {
 static byte g_tuyaBatteryPoweredState = 0;
 static byte g_hello[] = { 0x55, 0xAA, 0x00, 0x01, 0x00, 0x00, 0x00 };
 //static byte g_request_state[] = { 0x55, 0xAA, 0x00, 0x02, 0x00, 0x01, 0x04, 0x06 };
-static bool g_tuyaMCU_powerSave = false;
+static bool g_tuyaMCU_batteryPoweredMode = false;
 
 typedef struct tuyaMCUPacket_s {
 	byte *data;
@@ -2270,7 +2270,7 @@ commandResult_t Cmd_TuyaMCU_EnableAutoSend(const void* context, const char* cmd,
 	return CMD_RES_OK;
 }
 
-commandResult_t Cmd_TuyaMCU_PowerSave(const void* context, const char* cmd, const char* args, int cmdFlags) {
+commandResult_t Cmd_TuyaMCU_BatteryPoweredMode(const void* context, const char* cmd, const char* args, int cmdFlags) {
 	int enable = 1;
 
 	Tokenizer_TokenizeString(args, 0);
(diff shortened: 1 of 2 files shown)


protectivedad:
It looks like I have no idea what I am doing. I was doing this through the PR because I couldn't find any instructions on building for the BK7238, but all the builds are failing. If anyone can point me towards how I can get the PR working. I just cloned the source made a minor change and uploaded it.

openshwprojects:
This is C compile error, you need to add  {    } block around image

like that:
image


or move byte reply[2] outside (before) switch

openshwprojects:
I think you also need to set reply[1] = 0; , otherwise this byte is not initialized and has random value

protectivedad:
> I think you also need to set reply[1] = 0; , otherwise this byte is not initialized and has random value

I shouldn't have submitted the PR yet. It's my first time using actions to compile a project. I didn't know it would compile on my github without submitting the PR. If I am spamming you with updates just let me know how I can unsubmit the PR until I'm ready.

Thanks, for all the work.

openshwprojects:
please keep working on this pr, that's the whole idea here, so you can build easily and i can help

protectivedad:
> please keep working on this pr, that's the whole idea here, so you can build easily and i can help

Okay.

When the BK7238 is repowered the channels are zeroed. and then when the TuyaMCU sends the status they are populated. The problem is that the 0 is the same as door closed. So it ignores the update and doesn't publish to MQTT. Is the TuyaMCU driver supposed to populate the channels with values from the previous power cycle?

My current startup command line is:
```
backlog loglevel 6; logfeature 6 0; startDriver tuyaMCU; linkTuyaMCUOutputToChannel 101 bool 1; setChannelType 1 ReadOnly; linkTuyaMCUOutputToChannel 102 val 3; setChannelType 3 ReadOnly;
```

I have no flags selected, though I tried flag 46 and flag 47 but they had no effect.

EDIT:
Just ignore. I forgot they default to 0 because of the "Configure Startup".

protectivedad:
Okay, this is pretty much done. The version 3 battery operated device I have flows nicely in the start of "TuyaMCU_RunStateMachine_V3" function when Power Saving is set. The device also works when Power Saving is NOT set.

The whole driver could be simplified if instead of using the tmSensor you set version and processing based on the TuyaMCU version sent in the product information response. I was going to look into it but I don't have any version 0 devices. Just in broad strokes I was thinking set a global version number. Assume non-zero. Send product info request, grab version number from response. Then process startup accordingly. Use my new command to allow non-zero versions to be battery efficient. No need for tmSensor because the driver will auto detect low power "communication".

openshwprojects:
Ok, can you just change the name to keep syntax similar to our PowerSave?
So change: tuyaMcu_enablePowerSaver to tuyaMcu_powerSave

also, for the sake of documentation, can you write a short description in tuyamcu topic:
https://www.elektroda.com/rtvforum/topic4038151.html
what this does exactly, why and how?

protectivedad:
Let me know if the syntax changes work for you and then I'll post in the TuyaMCU thread.

protectivedad:
I'd like it if this could be tested on a non-battery powered version 3 sensor. I don't have any. If they work then I'd go for `tuyaMcu_powerSave` if they don't then `tuyaMcu_batteryPoweredMode` would be best.

protectivedad:
Renamed command and tested:
```
tuyaMcu_batteryPoweredMode
tuyaMcu_batteryPoweredMode 0
tuyaMcu_batteryPoweredMode 1
```

All results are what I expect.

openshwprojects (approved):

PR #1918, opened 2025-12-26
drv_battery: Fix for incorrect log messages

protectivedad:
Removed unused reading of ADC which caused incorrect log messages saying device wasn't on battery.
Changed "DEBUG" message after correct ADC reading to "INFO" so the battery information is still in the log.


protectivedad:
The documentation says setting `Battery_Cycle 0` should stop polling battery sensor, but it actually makes it poll every second. I've fixed that also and will add it as a second commit on this PR. It will poll once on boot (as it did before) and if `Battery_Cycle 0` is run it will do the last poll and then stop polling.

protectivedad:
I experimented with increasing the timing of the first reading, because with my device the reading is only valid after the device finishes the boot. It didn't really make a difference unless I increased it above 5 cycles, so I left it the way it was 1 cycle. At one cycle the web page gets the first value (maybe invalid), but it never gets published. If anyone has a suggestion on how to ensure the first reading is good (in the driver) let me know.

I use `waitFor MQTT 1` and then set the `Battery_Cycle 0` to poll and publish one value.

protectivedad:
Made the battery driver more resilient. Invalid readings (anything less than half the expected voltage) are ignored. Effectively this means the command `Battery_Cycle 0` can be run to stop reading the ADC at any time during the boot process and at least one valid value will be read. Removing the reliance on `waitFor MQTT 1`.

divadiow:
should this be merged? seems sad not to make use of your work

protectivedad:
Has there been any other changes to the drv_battery anyone working on it? If not then I think it should be merged.

PR #1925, opened 2026-01-03
Fix bl602 crashing trying to publish NaN

giedriuslt:
Bl602 crashes after frequency publishing was introduced in 18.209. With bl0937 frequency is NaN and trying to publish that crashes. This fixes the problem.


divadiow:
nice one. 

xref: https://github.com/openshwprojects/OpenBK7231T_App/pull/1853#issuecomment-3705664755

openshwprojects:
I request changes - do this in the MQTT publish float handler, not in BL code.

giedriuslt:
Moved the fix to float publish handler

PR #1928, opened 2026-01-04
Make soft I2C clock frequency configurable

staalb:
Necessary to get a RGB LED bulb working (Action LSC SMART GLS A60 SMD 220-240V 8W 700lm RGB + 2700-6500K E27). It has a SM2185N controller, which seems to work with the SM2235 driver. However because the frequency of the CLK line is too high, the SM2185N controller often misinterprets frames. When turning on or off the LED, changing the brightness or color random thing happen.

Therefore, this change makes it possible to lower the clock frequency using a new command: `SoftI2C_SetClkPeriod`. The default period is 4, setting it to 40 seems to be fine.

This change also removes a duplicate definition of `SM2135_DELAY`.

I will make a forum post with more details about this device.


staalb:
See https://www.elektroda.com/rtvforum/viewtopic.php?p=21798914

openshwprojects (approved):

PR #1931, opened 2026-01-05
Send currents to controller immediately if changed

staalb:
As discussed in this post: https://www.elektroda.com/rtvforum/viewtopic.php?p=21799955#21799955

Startup commands are executed only after the LED driver state is restored after a reboot (when flag 12 "[LED] Remember LED driver state" is enabled). Without this change, When the SM2235_Current command is executed as a startup command, it would have no effect on the already enabled LED driver.


openshwprojects:
Currently (no pun intended) I've put a simpler solution, but I'm still open to new suggestions. For now, I'll close this PR.

Issue #963, opened 2023-10-30
Possible solution of config safe on W800


djcfg123:
I have the same problem as #913 . 
I accidentally chose 1.17.1 from the list and saving the configuration, parameters, etc. works on it, on the new ones it doesn’t work, immediately after the firmware the configuration file error and loading default. 
I have a suggestion: in the source code of the firmware offset 0x1F0303 is indicated, which is taken from the SDK example, that is, the recording area of ​​the configuration structure "mainConfig_t" which in earlier versions has the size MAGIC_CONFIG_SIZE_V3 2016 and works on the W800, and after adding SSID 2 has the size MAGIC_CONFIG_SIZE_V4 3584 and does not allow you to save the configuration on the W800. However, the user data area is at the starting address 0x8000000 + 1E0000, it might be worth shifting 1F0303 to 1E0000 (or so), unless of course there is a memory layout somewhere else that affects the beginning of the user block (maybe I didn't see this in the code).
Perhaps I would test this theory myself, but I’ve been trying to set up the SDK to compile the firmware for 2 days now and I’m getting build errors like “inttypes.h: No such file or directory”, “sys_rtos.h: No such file or directory” and similar
![image](https://github.com/openshwprojects/OpenBK7231T_App/assets/93264015/83275e43-49a0-4016-bfe4-254cbcd1f739)

openshwprojects:
Hey, Why not use online builds? I can check this for you

openshwprojects:
https://github.com/openshwprojects/OpenBK7231T_App/pull/964

djcfg123:
Tell me, how can I use the online building? I tried to do this through the W800 SDK according to the guide written in BUILDING.md, but nothing works

openshwprojects:
I have merged that fix... for w800 config, dsoes it work now?

you can fork our repo and it should build on your fork, it will also build if you open a pull request after I approve your pull if you're a new contributor

djcfg123:
Yes, its working 😀
I was also able to fork and compile the firmware for w800, and there I shifted the address from 1F0303 to 1E0000 (as in the first message of this issue) and enabled SSID2 support for W800, it works, saves the configuration correctly, but I did not find the code itself  which uses the second network if it was not possible to connect to the first.
But is it still possible to enable IR support for the W800?

openshwprojects:
What has changed in RTOS? When? Can you point to specific commit that may break the system? It sounds like an important issue to fix.

[...]

Issue #835, opened 2023-05-24
W600 cannot save settings


henkiejan1:
**Describe the bug**
I flashed before my W600 module and works fine. But with the new release i can put in wifi settings for x times but it looks like it saves nothing and booted just as AP... If i press on the scan wifi network button i got: W800 TODO. The downloaded latest version i got did not have the alpha extension...

**Firmware:**
- Version [OpenW600_1.17.127.fls]
- Device: Wavgat WIS600-01s
- Chip/model: WinnerMicro W600
- Device config?

**To Reproduce**
Flash a W600 device and try to connect to wifi

openshwprojects:
Hello, the WiFi network scan is TODO, as described, but still, device should save settings.
I can try to find my W600 board and check.
In a meantime, can you flash by UART?
Can you try version from before April 20, 2023 and check if there also config save is broken?

henkiejan1:
Update: flashed the version op 19 april and that works! But needed to set static ip settings but ok. :)

openshwprojects:
So you're saying that newer version is broken? Are you able to tell when exactly it broke?

dy008:
I also have the same problem. After flashing into the 1.17.129 version, the wifi settings cannot be saved, only the AP function is enabled, and no configuration can be done. Flashing into 1.17.06 is normal

openshwprojects:
Fixed in:
https://www.elektroda.com/rtvforum/viewtopic.php?p=20616695#20616695

PR #1936, opened 2026-01-07
Try fix flash restrictions for W600 and W800 

MaxineMuster:
...by moving start of FLASH_VARS (changing FLASH_VARS_STRUCTURE_ADDR)

When trying to add some data to flash config for a possible usage as WPA-AP I ran into the long-standing "limitations" for W600 and W800 regarding flash config.

E.g. here for W800 https://github.com/openshwprojects/OpenBK7231T_App/issues/963
and here for W600 https://github.com/openshwprojects/OpenBK7231T_App/issues/835

While the solution was a "workaround" to reduce config size for these platforms, I think I found the real "culprit" and this should be easy to fix:
Can it be, that this simply was caused by the code for flashVars on W600 (and later also used for W800) "overwriting" the config ?!?

```
src/hal/w800/hal_flashVars_w800.c:

…
FLASH_VARS_STRUCTURE flash_vars;
static int FLASH_VARS_STRUCTURE_SIZE = sizeof(FLASH_VARS_STRUCTURE);

//W800 - 0x1F0303 is based on sdk\OpenW600\demo\wm_flash_demo.c
//W600 - 0xF0000 is based on sdk\OpenW600\demo\wm_flash_demo.c
//2528 was picked based on current sizeof(mainConfig_t) which is 2016 with 512 buffer bytes.

#if defined(PLATFORM_W600) 
#define FLASH_VARS_STRUCTURE_ADDR (0xF0000 + 2528)
#else
#include "easyflash.h"
#endif
…
```

Maybe simply the "hardcoded" start in flash (FLASH_CONFIG_ADDR + 2528) prevented the enlarged config to be saved (or config "beyond" 2528 been overwritten)?
I tried with an increased value and it seems this really solves the issue, so the special case for W600 would be obsolete and the restrictions in new_pins.h (allowing e.g. no second SSID for W600 and W800) could be removed.

BTW: For W800 restriction should have been resolved some time ago when it was moved to easyflash ?!?

@divadiow kindly did some promising tests (since I don't have a W600) and I tried on W800 (reverting change to easyflash) in my testing branch https://github.com/MaxineMuster/OpenBK7231T_App/tree/_WPA%2BGUI.

But I think it would be better to "seperate" this fix from adding a new feature ...

### DIFF
--- a/src/new_cfg.c
+++ b/src/new_cfg.c
@@ -855,16 +855,6 @@ void CFG_InitAndLoad() {
 		}
 		WiFI_SetMacAddress((char*)g_cfg.mac);
 #endif
-#if defined(PLATFORM_W600)
-		if (g_cfg.version <= MAIN_CFG_VERSION_V3){
-			// we read a valid V3 config, convert to V5
-			// (flash_vars should have been moved before by HAL_FlashVars_IncreaseBootCount() ... )
-			g_cfg.version = MAIN_CFG_VERSION;
-			// zero additional memory - it included flashvars and old date might be interpreted as web password 
-			memset((char*)(&g_cfg) + MAGIC_CONFIG_SIZE_V3, 0, sizeof(mainConfig_t) - MAGIC_CONFIG_SIZE_V3);
-			g_cfg_pendingChanges ++;
-		}
-#endif
 		addLogAdv(LOG_WARN, LOG_FEATURE_CFG, "CFG_InitAndLoad: Correct config has been loaded with %i changes count.",g_cfg.changeCounter);
 	}
 
@@ -877,7 +867,16 @@ void CFG_InitAndLoad() {
 #if ALLOW_WEB_PASSWORD
 	// add web admin password configuration
 	if (g_cfg.version<5) {
+#if defined(PLATFORM_W600)
+	// W600 changed from V3 to V5, so at this point
+	// we read a valid V3 config but need a V5 config.
+	// Memory might contain old data (from flash_vars)
+	// so let's zero complete additional memory (including g_cfg.webPassword, making it empty).
+		memset((char*)(&g_cfg) + MAGIC_CONFIG_SIZE_V3, 0, sizeof(mainConfig_t) - MAGIC_CONFIG_SIZE_V3);
+		g_cfg_pendingChanges ++;
+#else
 		strcpy_safe(g_cfg.webPassword, "", sizeof(g_cfg.webPassword));
+#endif
 	}
 #endif
 	g_cfg.version = MAIN_CFG_VERSION;
--- a/src/new_pins.h
+++ b/src/new_pins.h
@@ -1553,13 +1553,6 @@ typedef struct mainConfig_s {
 	char ping_host[64];
 	// ofs 0x000005E0 (dec 1504)
 	//char initCommandLine[512];
-//20260106 - try for V4 of config for W600 and W800
-#if 0
-//#if PLATFORM_W600 || PLATFORM_W800
-#define ALLOW_SSID2 0
-#define ALLOW_WEB_PASSWORD 0
-	char initCommandLine[512];
-#else
 #define ALLOW_SSID2 1
 #define ALLOW_WEB_PASSWORD 1
 	char initCommandLine[1568];
@@ -1579,7 +1572,6 @@ typedef struct mainConfig_s {
 	byte disable_web_server;
 	// offset 0x00000CBC (3260 decimal)
 	char unused[324];
-#endif
 } mainConfig_t;
 
 // one sector is 4096 so it we still have some expand possibility


MaxineMuster:
Just realized it's not yet complete - missing changes in new_cfg (where W600 is still set to "V3" and hence using a wrong configSize (configSize = MAGIC_CONFIG_SIZE_V3).
Will try to fix this later ...

divadiow:
I have flashed 1936_merge_59dbd052ec1d

I can join test AP (W600 is client) - credentials persist.
startup command persists
random flags persist

LFS autoexec.bat persists

here's a 2mb dump taken from flash.vue web app for analysis
[W600_QIO_w600CD0A369D.zip](https://github.com/user-attachments/files/24478302/W600_QIO_w600CD0A369D.zip)

I was expecting to see content at or shortly after E8000 (https://github.com/openshwprojects/OpenBK7231T_App/blob/4f4af7347c3ca8a35c151c5e97e3bd85ae6623f4/src/littlefs/our_lfs.h#L125)

but it looks like there's another LFS header at E9000, and then my content - is this expected/related?

image

I'm just testing something MQTT because saving settings reboots whole W600 with no crash - checking general release.

divadiow:
also OTA 1936_merge_59dbd052ec1d -> 1936_merge_59dbd052ec1d seems successful and settings/lfs persisted. 

persistence with 1936_merge_59dbd052ec1d -> 1936_merge_f29c4b792eab

divadiow:
OK, I think MQTT is a different issue. caught a crash with 1.18.238 - nothing else configured, just when saving MQTT settings.

```
Info:MAIN:Time 47, idle 0/s, free 57961, MQTT 0(3), bWifi 1, secondsWithNoPing -1, socks 2/8 
Stack Range[20006be4,200073e4], used size:45
Priority:15
Taskname:tcpip_t
20007330:2001e2c8 0802f2cf 2001d3b4 20020b20 
20007340:200073d4 a5a5a5a5 2001e2bc 00000000 
20007350:8001a8c0 00000000 20029478 9501a8c0 
20007360:00005010 2001e2c4 00000000 200073d4 
20007370:20009570 20029800 00000008 200277d8 
20007380:00000000 08078d60 a5a5a5a5 a5a5a5a5 
20007390:a5a5a5a5 080282df a5a5a5a5 0803af3b 
200073a0:0000000d 00000000 00000000 00000000 
200073b0:2001e5b4 2001e5b4 2001e5c8 2001e5c8 
200073c0:08078d60 a5a5a5a5 a5a5a5a5 0803e16f 
200073d0:00000000 2000e0a0 a5a5a5a5 00000000 
200073e0:a5a5a5a5 a5a5a5a5 heap:54018
```

MaxineMuster:
Looks good, thanks.
I'm curious about the two LFS signatures, will try to find if I can find a reason for this. 
If I'm not mistaken it was introduced in
https://github.com/openshwprojects/OpenBK7231T_App/pull/1629
and with 0xE8000...

divadiow:
did I miss any test out with this PR? 

re LFS - I did do full flash erase before flash of PR using Upgrade Tools (W600 in EF doesn't appear to erase at the moment).

I'll do it again to be sure. I wonder what @NonPIayerCharacter thinks

divadiow:
OK, this is:

full erase and PR flash
image

join AP, dump - no LFS headers
[withautoexec_w600.zip](https://github.com/user-attachments/files/24483793/withautoexec_w600.zip)

create autoexec = two headers 0xE8000 and 0xE9000
[noautoexec_w600.zip](https://github.com/user-attachments/files/24483737/noautoexec_w600.zip)

[customreadsnips.zip](https://github.com/user-attachments/files/24483735/customreadsnips.zip)

divadiow:
wait. maybe it's nothing. it must be the read process? could jlink read I guess to be sure. 

image

lmk if you want these posts deleting to keep this pr focussed

-----------------

having said that, no errors reading those areas this time and still..
[W600_QIO_w600CD0A369D (1).zip](https://github.com/user-attachments/files/24483990/W600_QIO_w600CD0A369D.1.zip)

MaxineMuster:
> create autoexec = two headers 0xE8000 and 0xE9000

Oh, that's embarrassing, why didn't I see? 
That will probably be simply the second "superblock" (or whatever the structure to hold the contents in the filesystem is named here) after a 4k block of the first one...

divadiow:
oh. ok. so all as expected it sounds! 

sorry for the distraction away from PR!

MaxineMuster:
No need to be sorry for anything. Its all related to this PR. 
So, from your perspective, PR works as expected, allowing "full" config also for W600/W800 and no need for restricting to config version 3.

MaxineMuster:
Did some changes trying to make a smooth upgrade. Maybe @divadiow you could give it another try?
I hope now upgrading from V3 config to this PR will keep both config and flashvars (and saving is still working ;-))
Thanks!

divadiow:
I'm afraid even from fresh untouched general release to latest PR this is the result when reconnecting after OTA
image
[otathenboot.txt](https://github.com/user-attachments/files/24946183/otathenboot.txt)

MaxineMuster:
Oh, sorry.
For the moment I can only think of the data in the additional space for config was not empty, but contains flashvars. I didn't "erase" the flash.
Let me check, if that's interpreted as web password.
Will check tomorrow, thanks again!

divadiow:
bum. same login prompt after OTA to 1936_merge_38607c823f84

MaxineMuster:
Oh no ...

Would you mind to fetch another flash dump and share it here?

divadiow:
OK. 1.18.251 - > 1936_merge_8c0f2004a68a no password prompt

1.18.251 - set mqtt host, flags 5,6, startup command ntp , startup completion to 10s. OTA to 1936_merge_8c0f2004a68a = all still present :)

[W600_QIO_w600CD0A369D (2).zip](https://github.com/user-attachments/files/24972662/W600_QIO_w600CD0A369D.2.zip)

MaxineMuster:
So it's finally working?
Great, I really appreciate your constant and "resilient" support.

I will try to strip some of the actual debug output so we can see, how much overhead the special handling will cost us in terms of additional image size.

divadiow:
did it again. 1.18.251-> 1936_merge_3fd82c1f8da4
flags, startup, pins etc persisted

divadiow:
however, does https://github.com/openshwprojects/OpenBK7231T_App/pull/1969 make this flashvars work redundant?

MaxineMuster:
> however, does #1969 make this flashvars work redundant?

Indeed, using easyflash is the more elegant version to get rid of the flashvars at this memory area.
So let's wait until #1969 is merged then it's only removing special case for W800 to use the complete page for config.

We'll still need a special case for update to keep config - or leave it to only defining new size (rendering the actual config invalid because CRC won't match).

PR #1939, opened 2026-01-07
Allow MQTT connection and processing in one every second iteration

protectivedad:
Further to the discussion https://www.elektroda.com/rtvforum/viewtopic.php?p=21801342#21801342

The current MQTT every second loop allows only connect or process each turn. This PR allows both to occur in the same turn. It will only do this if the fast connect flag is selected. When the fast connect flag is selected after submitting the MQTT connect it will wait for up to 200 ms for a connection. If a connection occurs it will then continue and process MQTT messages.

In my testing this shaves a second off the post time.


protectivedad:
This is ready to be tested it is built on my main-timing PR. Without the timing PR there is no real way to test and make sure that the improvements really increase the speed of the system. Plus with the timing information you can see what is holding up the MQTT from working faster.

One thing I noticed was the original code, if the system triggered a reconnect it would force a disconnect in line 2235 but process MQTT messages as if the broker was still connected for the rest of the loop. I decided that didn't make any sense so the new behaviour is to force disconnect and exit.

That should be the only change in the original routine except for being able to process after a connect.

I do have one other enhancement I need for my non-TuyaMCU door sensor. Right now MQTT_ChannelPublish information is dropped if the broker is not connected. In this PR I have added code to check if the broker is ready and if not post to Queue.

I'm not sure about removing the /get from the publish.

protectivedad:
Tested on BK7238 TuyaMCU and BK7231N non-TuyaMCU door sensors.

PR #1940, opened 2026-01-08
7231n/7238 battery variants

NonPIayerCharacter:
https://www.elektroda.com/rtvforum/topic4160268.html

### DIFF
--- a/src/user_main.c
+++ b/src/user_main.c
@@ -190,6 +190,15 @@ static int get_tsen_adc(
 #endif
 
 #if PLATFORM_BEKEN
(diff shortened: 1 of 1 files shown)


protectivedad:
I was thinking a define in the obk_config would give maximum flexibility and document what is going on. That way looking at the obk_config will immediately tell someone the device is using 250, 10 or 0 ms delay. Then custom compiles are easy without adding #if's all over the place.

https://github.com/protectivedad/OpenBK7231T_App/commit/de885e6a3df1f6d059037977cdcb192350f1b17c

Renamed to battery of course.

openshwprojects (approved):

PR #1942, opened 2026-01-10
BMP280 enable

internetson:


openshwprojects:
Looks like makefile needs to be updated

PR #1943, opened 2026-01-10
Extra power save feature

NonPIayerCharacter:
Unknown if stable.
Reduces current by 0.001-0.002A, ~0.005A on RTL8720C.
Enabled by default on BL602 (same as in SDK).
XRs untested.

Can be enabled on LN882H with some SDK changes. Requires disabling tickless idle, which increases current by ~0.005A. With this enabled, current is lower that with tickless idle by 0.005-0.01A.


divadiow:
there's also a new powersave (well, "Add idle suspend mode") mode in newer XR872 SDK 
https://www.elektroda.com/rtvforum/viewtopic.php?p=21775390#21775390

PR #1944, opened 2026-01-12
logging: fix in tcp_log_ports declaration

protectivedad:
TCP log ports were initialized as {-1, 0} which would lead to the first one being assigned and the second one being used as client_fd = 0 at least once. Might have caused problems. Not sure but I changed it to be {-1, -1}.

### DIFF
--- a/src/logging/logging.c
+++ b/src/logging/logging.c
@@ -99,7 +99,7 @@ static int g_extraSocketToSendLOG = 0;
 static char g_loggingBuffer[LOGGING_BUFFER_SIZE];
 
 #define MAX_TCP_LOG_PORTS 2
-int tcp_log_ports[MAX_TCP_LOG_PORTS] = {-1};
+int tcp_log_ports[MAX_TCP_LOG_PORTS] = {-1, -1};
 
 
 void LOG_SetRawSocketCallback(int newFD)


protectivedad:
@NonPIayerCharacter 
Could this have been what caused the problem that #1930 was pushed for? It would be nice if the BK7238 didn't have to have a spinning tcp log.

This works on my BK7238 device and doesn't use the continuously running threads.

NonPIayerCharacter:
I'll test it on N+BL0937 plug and ALT firmware.
Before my commit, there were spikes in measurements (like from 250v to 60000)
This might take some time.

NonPIayerCharacter:
There are no giant spikes, but it's still worse that how it is now.
image

Or not. I flashed last release, and nothing changed. Perhaps i've had loglevel set to 0 previously?
But i think separate log thread is better, since there are more idle cycles.

protectivedad:
I'm not following. Just to clarify the PLATFORM_BEKEN version publishes on demand. The other version uses a separate thread in a while loop that polls for changes. The polling version is the one that you'd prefer?

I've resubmitted with only the one line fixed. It was only a problem after a network connection to port 9000.

NonPIayerCharacter:
Separate thread seems to be more efficient.

Anyway, in the early development of BK7238 and alternate BK7231N and BL0937 there were problems with big current/voltage spikes like i mentioned earlier, unless log was disabled. I still don't know why or how this fixes it.

PR #1955, opened 2026-01-24
ESP8266 changes + LN882H sdk update

NonPIayerCharacter:


openshwprojects:
I think I need it right away.

PR #1960, opened 2026-01-25
Change Python version from 2 to 3 in workflow

xjikka:
Updated Python version for installing requirements and esptool.

### DIFF
--- a/src/user_main.c
+++ b/src/user_main.c
@@ -247,12 +247,12 @@ void extended_app_waiting_for_launch2(void) {
 #endif
 
 
-#if PLATFORM_LN882H || PLATFORM_ESPIDF || PLATFORM_ESP8266 || PLATFORM_REALTEK_NEW
+#if PLATFORM_LN882H || PLATFORM_ESPIDF || PLATFORM_REALTEK_NEW
 
-__attribute__((weak)) int LWIP_GetMaxSockets() {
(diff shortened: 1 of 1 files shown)


xjikka:
@openshwprojects How do you build for the ESP8266? I had to make the changes in this PR to get it compiling.

xjikka:
@NonPIayerCharacter How do you build for the ESP8266? I had to make the changes in this PR to get it compiling.

NonPIayerCharacter:
Doing local builds should be easy with #1955

xjikka:
> Doing local builds should be easy with #1955

with PR #1955 changes, online workflow build is now possible - I tried in #1962

kofec:
I'm not an expert but what about this line:
pip2 install virtualenv ? pip2 -> pip3 ?

xjikka:
> I'm not an expert but what about this line: pip2 install virtualenv ? pip2 -> pip3 ?

I know, thank you, but:

I'm not looking for a solution, I already found it. I was just asking how others compile it when the workflow doesn't work. For those who can't, I've offered a simple solution with this PR to get the compilation working.

However, the day before, @NonPIayerCharacter already made #1955, which solves this, which I overlooked.

And regarding the pip change - I did the simplest way possible (minimal changes) in workflow.yaml so as not to break anything else. Hence the python3 call instead of pip change

PR #1975, opened 2026-02-04
HW Timer HAL

NonPIayerCharacter:


divadiow:
What does this do?

NonPIayerCharacter:
Timer abstraction for new IR driver. And if tested and working, for old IR and RC drivers.
I tested IR receive on BK7238, RTL8720C, RTL8720D and RTL8721DA. Last two weren't working before that, but work now.
Platforms that need testing (both send and recv) - BL602, LN882H and LN8825, and IR send on those i tested IR recv before.

IR is not enabled, but can be now on ECR, W600, W800, XRs.

Plus it will allow simultaneous IR and RC

PR #1983, opened 2026-02-08
Docker local build

DeDaMrAzR:
- Added build_tool.py: Interactive CLI build tool with dependency resolver
- Added GUI_build_tool.py: GUI wrapper for build tool
- Added entrypoint.sh: Docker entrypoint script with parallel build support
- Updated Dockerfile: Optimizations for faster builds
- Updated README.md: Documentation for new build tool

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,11 +1,1 @@
-# Set the default behavior, in case people don't have core.autocrlf set.
-* text=auto
-
-# Explicitly declare text files you want to always be normalized and converted
-# to native line endings on checkout.
-*.c text
-*.h text
-
-# Declare files that will always have CRLF line endings on checkout.
-*.sln text eol=crlf
-*.vcproj text eol=crlf
+docker/entrypoint.sh text eol=lf
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,6 +5,8 @@ ARG DEBIAN_FRONTEND=noninteractive
 ARG USERNAME=devuser
 ARG UID=1000
 ARG TZ="Etc/UTC"
+ENV LANG=C.UTF-8
+ENV LC_ALL=C.UTF-8
 
 RUN echo "$TZ" > /etc/timezone
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,5 @@
-FROM --platform=linux/amd64 ubuntu:20.04
+ARG BUILDPLATFORM=linux/amd64
+FROM --platform=$BUILDPLATFORM ubuntu:22.04
 
 ARG DEBIAN_FRONTEND=noninteractive
 ARG USERNAME=devuser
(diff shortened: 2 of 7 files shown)


DeDaMrAzR:
For context make sure you RTF readme.md 🙂 but worth putting it here again **WARNING!!!** - 40+ Gb of HDD space needed for docker volumes, images and cache!!!

DeDaMrAzR:
ATM tested on BK, BL and ESP platforms only, please report here any errors on other platforms and thanks!

MaxineMuster:
Thanks for this work!
No error, but confirmation LN882H works!


Maybe you could revise the md "documentation"?
The information about build_tool.py is not 100% straight (for me).

First it claims:

_NOTE!! - build_tool.py goes in the root folder, while the other two go in the docker folder._

Later:

Copy Files (Crucial Step):

Copy/replace Dockerfile, **build_tool.py** and entrypoint.sh to: OpenBK7231T_App\\**docker**\\
(The docker folder is already present in the repository).

Final Structure must look like this:
```

OpenBK7231T_App\

├── docker\
│   ├── Dockerfile         <-- Docker config is HERE
│   ├── entrypoint.sh      <-- Entry script is HERE
│   └── build_tool.py      <-- Script is HERE
├── platforms\
├── src\
└── ... (other git files)
```
Since python code checks for the case, where it's in the docker directory, both selections are fine:
```

if os.path.basename(_script_dir) == "docker":
    # If we are in docker/, repo root is one level up
    REPO_ROOT = os.path.dirname(_script_dir)
else:
    # Assume script is in root
    REPO_ROOT = _script_dir

```
And thanks for pointing out bl_shared was still depending on NTP. Hope this will be resolved soon with #1984

DeDaMrAzR:
@MaxineMuster thank you for the test, I appreciate it. I am on a marathon dev session on this so yeah readme is not 100% accurate and/or complete - it will be soon. For now monitor this thread for updates - https://www.elektroda.com/rtvforum/viewtopic.php?p=21834411#21834411

DeDaMrAzR:
image

my last clean check (from zero, new folder and all) was successful, added hardening for SDK downloads as much as I think I should but github is funny with large downloads so be patient....

it was tested against latest upstream main, play around, test it, break it and report in the forum

DeDaMrAzR:
PR complete unless someone finds anything wrong, I did multiple stress-tests and found it working as it should. One thing to mention if you encounter build errors - use clean checkbox and rebuild, if that fails tick no cache checkbox, it will rebuild everything - if the build still fails report with specifics.

DeDaMrAzR:
Last commit added local OTA option, some guardrails are implemented like platform check and basic IP structure check. Source Dir path should be blank now on every start - no persistence at this point. Driver list is a checkbox "table" beter visibility and it is not cleared on Start Build button push. UTF-8 is also checked/forced now so localized Windows problems should be gone ( I have no way of testing that - report back please). Since Dockerfile is changed full rebuild of containers and volumes should be done.

DeDaMrAzR:
not complete and not ready, still a lot of bugs to iron out.

DeDaMrAzR:
working to get this completed now

PR #1985, opened 2026-02-09
fix plausible regression from d491a71a938135f9, made avty disable global

avg-I:
It's easier to reason about the change if we consider when the availability topic is not added.
Originally that was when `isSensor && flagavty`, that is, when an entity was considered to be a sensor _and_ flag 35 was set.
After commit d491a71a938135f9, it became `isSensor || flagavty`. That is, (A) if flag 35 is set then all entities are stripped of the availability topic and (B) even when the flag is not set, then any entity considered to be a sensor would also be stripped of the topic.
I am confident that (A) was the intent and (B) is a bug.
So, we should simply check only `flagavty`.


avg-I:
The issue was also described and discussed here https://www.elektroda.com/rtvforum/topic4162760.html

avg-I:
Thank you!

PR #1990, opened 2026-02-13
HTTP TCP server: fix realloc failure handling to prevent request buffer leak

divadiow:


MaxineMuster:
Hopefully your approach is more successful than my first try ;-)
https://github.com/openshwprojects/OpenBK7231T_App/pull/1672

divadiow:
Oh! I remember that now. How come yours didn't get merged? I can't compare at the moment, but which one should be preferred?

MaxineMuster:
I also forgot about until your PR remembered me...
Will check tomorrow, the main fix also was to prevent losing the pointer during a failed reallocation, but iirc there also was a unrestricted increase of buffer which might lead to running out of memory.

MaxineMuster:
Regarding unfreed memory, our approaches are close.
You introduce "newbuf", I (re-)used "buf" to keep the original pointer:

image


So only real difference to my PR:
In actual code, if buffer is not sufficient, we grow it by 1024 until we run out of memory (reaching the problem addressed, where realloc fails).

I also tried to introduce an upper limit, "MAX_INCOMING_BUFFER_SIZE" if this is reached, buffer is no longer increased.
image

Can't tell, if this is reasonable, it was just an idea.
 
The rest of the old PR is code (no longer needed) to allow using new server on W800...

openshwprojects:
merged this one, if any more fixes, please open new prs

PR #2003, opened 2026-02-27
BT Proxy & minimal esphome api

NonPIayerCharacter:
Original ESP code and ESPHome api by @DeDaMrAzR 

Currently only passive/active scan on RTL8720D, no active connections.
ESP bt is disabled.

### DIFF
--- a/src/driver/drv_esphome_api.c
+++ b/src/driver/drv_esphome_api.c
@@ -10,7 +10,7 @@
 
 #define INVALID_SOCK -1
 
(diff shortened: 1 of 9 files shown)


DeDaMrAzR:
Since I never paid attention on ALT builds what happens to the binary size with BT compiled in?

NonPIayerCharacter:
N_ALT normal rbl - 496kb, with bt - 605kb.
7238 normal - 496kb, bt - 639kb.

Issue #1805, opened 2025-09-21
LN882H device suddenly stopped connecting to local network


bausensi:
**Describe the bug**
This RGBCCT lightbulb I flashed OpenBeken to, worked wonderfully for weeks once adjusted for power save and current limits. I controlled it through Home Assistant connected to my local WiFi network with a static IP. The device worked flawlessly for several weeks, but one day it suddenly stopped responding to commands. Safe mode works well and AP mode accepts clients without any issues. But it cannot connect to the local network once configured, whether through DHCP or with a static IP, I tried both. Erasing the flash and changing the MAC address didn't fix the issue. I've tried several other APs including my mobile hotspot and it couldn't join a single one.

I did a teardown surgery on it and reflashed it but it didn't work. Once reset it became apparent only the AP mode worked. Upon attaching UART wires I found out it tries to connect to my local network and even though it seems to succeed, an error code is returned and tries again forever. I found no information anywhere about this error code and I'm not familiar with the source code to figure out a solution, so I request help here.

**Firmware:**
- Version: 1.18.177
- Device: Generic Tuya lightbulb based on W509Z2 kit
- Chip/model: LN882HKI
- Device config: RGBCCT based on BP5758

**To Reproduce**
1. Establish UART connection on RX1/TX1 pair at 921600 baud
2. Reset device in safe mode
3. Fill out SSID and password, set a static IP configuration
4. Enable WiFi network mode and restart device
5. See UART logs

**Screenshots**

![Image](https://github.com/user-attachments/assets/f48bf021-4af0-4146-88ce-7121a769c18f)

**Additional context**
This is my only IoT device and it used to work well for 2 weeks.

MaxineMuster:
I never heard about this and I fear, it will be hard to fix, at least only with the error code 0xc4.
This is because all WiFi related code is in a library provided as a binary blob by the vendor and I don't think even this code is documented...
Do you have the possibility to test another access point, maybe a "simple" one with only WPA2 in case your has WPA3?

MaxineMuster:
Made a quick search in SDK for this code and only found one "near" match in bl (Bluetooth) code:

[components/ble/mac/ble/hl/api/rwble_hl_error.h](https://github.com/openshwprojects/OpenLN882H/blob/f9a0edae1d8087f4789e1ff78826a858301425dc/components/ble/mac/ble/hl/api/rwble_hl_error.h#L493)

```
/// Reserved slot violation
--
[...]

PR #1688, opened 2025-06-24
Try finding best BSSID for a SSID

MaxineMuster:
A try to address https://github.com/openshwprojects/OpenBK7231T_App/issues/1684

NOT tested. If it works as expected, it should log the best AP for a SSID (on UART).
If it works as expected, I coud try to set BSSID for the "wifi_sta_connect()"-call.

So if there is anybody able to test this PR? @divadiow ?

PR #2019, opened 2026-03-07
LN882H try using best BSSID

MaxineMuster:
New try for #1688 hopefully fixing issue #1805


MaxineMuster:
This firmware will do two scans to find the AP with the best RSSI for the actual SSID.

Later call to scan will use channel and BSSID found during these scans:

LN882H_BSSID

PR #2022, opened 2026-03-09
Shutter driver enhancements and enable

D-Fed11:
This pull request adds several improvements to the shutter driver:

Features Added

1. Direction Change Delay (Critical Safety Feature!)

- Adds a 200ms delay when switching shutter direction to prevent relay/motor damage
- Automatically turns off both relays temporarily when direction changes
- Relays are re-engaged after the delay expires

2. Persistent Shutter State Storage (New Command: ShutterChannels)

- Allows assigning dedicated channels to store shutter position, open time, and close time
- These values are automatically saved to persistent configuration and restored on device reboot
- Usage: ShutterChannels <shutterIndex> <positionChannel> <openTimeChannel> <closeTimeChannel>
- Example: ShutterChannels 1 20 21 22 - stores shutter 1's state in channels 20-22
- Use 0 for any channel to disable persistence for that value

3. Improved MQTT Performance

- Reduced excessive MQTT chatter by only publishing shutter position changes when the value actually changes
- Previously published every second even when idle, now publishes only on position updates
- Enhanced Command Parsing

4. Fixed ShutterMove command to support both legacy and standard argument formats. This was preventing "ShutterMove 0 stop" from being used as a command for scripting. 

- Legacy: ShutterMove1 stop (index as part of command name)
- Standard: ShutterMove 1 stop (index as first argument)
- Both formats now work correctly

Implementation Details

- Added direction change detection and delay mechanism (switchDelayRemaining used during relay switching)
- Added positionChannel, openTimeChannel, and closeTimeChannel fields to shutter_t struct
- Added lastPublishedPos for MQTT throttling
- Modified Shutter_Save() to persist values to both flash variables and assigned channels
- Modified Shutter_Read() to load from assigned channels first, falling back to flash variables
- Added Shutter_AssignChannels() function to configure channel persistence
- Updated DRV_Shutter_Tick() to handle relay switching delays properly
- Updated DRV_Shutters_RunEverySecond() to update UI channels and throttle MQTT
- Moved RegisterShutterForChannel() for proper function ordering

Testing
Driver was tested on a Loratap SC420W double shutter switch, and confirmed to be fully functional.

### DIFF
--- a/src/driver/drv_shutters.c
+++ b/src/driver/drv_shutters.c
@@ -39,16 +39,16 @@ shutter_t *g_shutters = 0;
 
 
 void Shutter_Save(shutter_t *s) {
-	HAL_FlashVars_SaveChannel(s->channel * 3 + 0, s->openTimeSeconds * 10);
-	HAL_FlashVars_SaveChannel(s->channel * 3 + 1, s->closeTimeSeconds * 10);
+	HAL_FlashVars_SaveChannel(s->channel * 3 + 0, s->openTimeSeconds * 100);
+	HAL_FlashVars_SaveChannel(s->channel * 3 + 1, s->closeTimeSeconds * 100);
 	HAL_FlashVars_SaveChannel(s->channel * 3 + 2, s->frac * 100);
 }
 void Shutter_Read(shutter_t *s) {
-	s->openTimeSeconds = HAL_FlashVars_GetChannelValue(s->channel * 3 + 0) * 0.1f;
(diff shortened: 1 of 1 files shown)


openshwprojects:
Not bad, however channels storage seems a bit redundant. I think this PR has simpler implementation:
https://github.com/openshwprojects/OpenBK7231T_App/pull/2013
Could you reuse the simple storage of shutters state from my PR and merge it to your other changes, which seems very useful?

PR #2024, opened 2026-03-14
Add commands to tokenizer for "options" in command line

MaxineMuster:
I often struggle on the parameters order e.g. when starting a driver like AHT2x: 
is it `<sda> <scl> <channel temp> <channel hum>` ?
or `<scl> <sda> <channel hum> <channel temp>` ?
or ...

So I like a call with "named" arguments like

`SDA=17 SCL=19 chan_h=3 chan_t=4`
which is totally equal to
`chan_h=3  SCL=19  chan_t=4 SDA=17`

So I tried to add some "useful" extension to tokenizer code to allow this sort of arguments.

With this change you can check for options like this:

`SDA=<pin-ID or pin name> / type=SHT4x / chan_t=<channel>`
&emsp;&emsp;  or even  `str="a string"` 	&emsp;&emsp;&emsp;&emsp;// use new TOKENIZER_ALLOW_QUOTES_IN_NAMEDARG_VALUE for this

`-SCL <pin-ID or pin name> / -type SHT3x / chan_h <channel>`
&emsp;&emsp; or  `-str "another string"` 		&emsp;&emsp;&emsp;// will "only" need TOKENIZER_ALLOW_QUOTES

new command are (all will need a default in case the option is not found):
&emsp; const char ***Tokenizer_GetArgEqualDefault**(const char *search, const char *def);
&emsp; int **Tokenizer_GetArgEqualInteger**(const char *search, const int def);
&emsp; int **Tokenizer_GetPinEqua**l(const char *search, const int def);

You can also check for a "single" option (like "-v" or "-single_mode")
&emsp; bool **Tokenizer_IsStringPresent**(const char *search);

Added selftest for the new commands, too.


MaxineMuster:
Streamlined and minimized addition: 

Now only args like "-SDA 4 -SCL P5" are possible (removed "XY=9" style)

Only one new function, all others as "optional" functions (enable with  "#define TOKENIZER_PARAM_EXTENSION 1"), else they can be used as #defines using the one new function.

PR #2030, opened 2026-03-16
Update PWM initialization for BK7231 and BK7238

abeltranconicet:


openshwprojects:
What's the point here? I was sure PWM was working before?

divadiow:
@abeltranconicet what does this do?

PR #2032, opened 2026-03-16
Configure Git settings for network stability

abeltranconicet:
Added Git configuration for network stability in workflow.


divadiow:
just curious, how do these improve stability?

Issue #725, opened 2023-03-13
Smooth lighting breats CW emulation


sanjay900:
**Describe the bug**
When smooth lighting is turned on (flag 18), CW emulation does not work right (flag 24). It seems that with smooth lighting turned on, the emulated cool white is unable to go to full brightness.

**Firmware:**
- Version 1.15.583
- Device BrilliantSmart E27 bulb
- Chip/model: BK7231T

**To Reproduce**
1. Use a bulb that requires CW emulation
2. Turn on CW emulation
3. Turn on smooth lighting
4. Set the bulb to Cool white. The bulb will be white, but at minimum brightness

HRDHRD5:
@openshwprojects 
Since #2043 is merged, this issue should be fixed and can probably be closed.

PR #2043, opened 2026-03-23
Fixing bug that causes cool emulated LED with RGB to be way to dimm, when smooth transition is enabled.

HRDHRD5:
Issue:
When smooth transition (flag 18) and emulate cool white with RGB (flag 24) are enabled, cool lighting is extremely dim.

Resolution:
Not setting the C Channel of RGBCW to 0 on every update.

Explanation:
In cmd_newLEDDriver.c:409 the lerp value array is given to LED_I2CDriver_WriteRGBCW.
The lerp value represents the current state of all led colors (RGBCW) in the color fade.
However the LED_I2CDriver_WriteRGBCW currently sets array[3] (C-Channel) to 0, which in turn causes the current lerp value of the cool LED Channel to be set to 0 in every Tick. This causes current lerp to be 5 after lerp adjustment in LED_RunQuickColorLerp.

### DIFF
--- a/src/cmnds/cmd_newLEDDriver.c
+++ b/src/cmnds/cmd_newLEDDriver.c
@@ -262,8 +262,7 @@ void LED_I2CDriver_WriteRGBCW(float* finalRGBCW) {
 			// the format is RGBCW
 			// Emulate C with RGB
 			LED_CalculateEmulatedCool(finalRGBCW[3], finalRGBCW);
-			// C is unused
-			finalRGBCW[3] = 0;
+			// keep C unchanged, because it is the lerp value for emulated cool LED
 			// keep W unchanged
 		}
 	}


HRDHRD5:
This Pull Request should fix #725

HRDHRD5:
I tested the fixed new firmware with a bl602 (issue is gone when testing...)

HRDHRD5:
Hey @openshwprojects, I dont want to be impatient, but can someone review and potentially merge this?
I dont know how I can state that this PR is ready for main.
And I want this to be on main, so I dont have to build the firmware on my fork in the future.😅

openshwprojects:
Hey can you check if it still works good if you disable smooth transitions?

Or maybe we should copy argument of LED_I2CDriver_WriteRGBCW to another array and send to I2C LED driver with fourth component set to 0?

HRDHRD5:
Thank you for your quick response!
When testing the change Emulate Cool Light with RGB works the same without smooth transition.
The rest also behaves normal.
I dont know exactly where else LED_I2CDriver_WriteRGBCW is used in detail, but for apply_smart_light (the non smooth transition function) there should be no difference with not setting C to 0.
Thist is because the Cool Warm Emulation is applied even before LED_I2CDriver_WriteRGBCW is called (cmd_newLEDDriver.c:690).

But I can also create a copy of led_rawLerpCurrent and use it as argument for LED_I2CDriver_WriteRGBCW if you want.

PR #2047, opened 2026-03-25
Update obk_config.h

hardwire1010:
Enable TCP UART Correctly


mfarrell-dev:
Custom build to enable UART over TCP

Issue #2052, opened 2026-03-27
Flag 12 causes hard crash and wrong channel values saved


WereCatf:
On a BL602-based RGBC-controller enabling flag 12 causes each channel to be saved with a value of 2147483647 and when attempting to load the device's homepage in the browser, that causes a hard crash. One has to open the configuration webapp, go to the Status-page and fiddle with the channels to bring them down to reasonable values.

**Firmware:**
- Version 1.18.279
- Device? A BL602-based RGBC-controller
- Chip/model: BL602
- Device config: `{
  "vendor": "Tuya",
  "bDetailed": "0",
  "name": "Full Device Name Here",
  "model": "enter short model name here",
  "chip": "BL602",
  "board": "TODO",
  "flags": "276825088",
  "keywords": [
    "TODO",
    "TODO",
    "TODO"
  ],
  "pins": {
    "3": "PWM;1",
    "4": "PWM;0",
    "12": "IRRecv;5",
    "20": "PWM;3",
    "21": "PWM;2"
  },
  "command": "",
  "image": "https://obrazki.elektroda.pl/YOUR_IMAGE.jpg",
  "wiki": "https://www.elektroda.com/rtvforum/topic_YOUR_TOPIC.html"
}`

**To Reproduce**
1. Go to configure flags
2. Turn flag 12 on to remember RGBCW settings -- specifically, nothing else appears matter. Just turning flag 12 on is the problem.
3. Restart
4. Browse to the device's index/homepage
5. Crash

openshwprojects:
PR merged so solved, I assume?

WereCatf:
> PR merged so solved, I assume?

Aye.

PR #2053, opened 2026-03-28
bl602 fv

NonPIayerCharacter:
#2052


WereCatf:
> #2052

I was just doing this myself when I noticed you had created a PR. I can confirm that this is working on my BL602 RGBC-controller.

PR #2055, opened 2026-03-28
Add Ariston Velis 80 ESP32C6 driver integration

DeDaMrAzR:
First iteration of the working Ariston driver for OBK. 

WIP - do not merge yet!!


DeDaMrAzR:
If anybody is interested in testing at this stage, currently enabled and working only on ESP32C6 modules, autoexec.bat should be:

//startDriver Ariston _MAC SN_ those must be obtained from your NFC boiler tag, example
startDriver Ariston 00:11:22:33:44:55 123456XX1234 
//NTP is needed for boiler internal clock sync 
startDriver NTP
// your time zone offset 
ntp_timeZoneOfs  +2:00

Schematic how to connect the module will be uploaded at some point when I find time, but soon. 

If you need to reset Energy Est (total) you can run 

> ariston_energyReset

image

Full water heater integration for HA implemented 

image

image

image

image

DeDaMrAzR:
Driver init sequence is now:
```
// MAC and serial number must be obtained from the boiler's NFC tag.
// RX and TX are required GPIO numbers, specified from the ESP module's perspective.
// startDriver Ariston <MAC> <SN> <RX_GPIO> <TX_GPIO>

startDriver Ariston 00:11:22:33:44:55 123456XX1234 6 7

// NTP is recommended for boiler clock synchronization.
startDriver NTP
//your time zone offset
ntp_timeZoneOfs +2:00
```
**RX and TX pins are mandatory with this change. The driver will not start if they are omitted, invalid, or identical!!**

Working on the original module that requires a chip transplant as it is locked down for programming. I'll provide results for that along with ESP32-C6 wiring if the original module is not used.

fabiocguerreiro:
Just a thought regarding the original ESP32-WATG-32D module and the decision to replace it due to the burned eFuses / flash protection.

I went back through the Elektroda thread. The original module boot log shows:

* `Key partition found`
* `Reserved build encrypted OK`
* `flash_crypt_cnt: 7`

and later @DeDaMrAz mentioned that the eFuses were all burned, so replacing the module was chosen instead.

However, I noticed that someone asked for the output of `esptool get-security-info`, but I couldn't find a later post containing that result. So unless I missed it, the exact Secure Boot state and silicon revision of the original chip were never published.

This may be worth revisiting because the module is an ESP32-WATG-32D. Espressif documentation lists the WATG-32D family among the older ESP32 v1.0/v1.1 silicon rather than ECO V3.

Those pre-V3 ESP32 revisions are affected by the published fault-injection/eFuse attack CVE-2019-17391. Espressif's own advisory states that voltage fault injection can bypass eFuse read protection and potentially recover the Flash Encryption and Secure Boot keys. The mitigation was added in ESP32-D0WD-V3.

So I'm wondering: has anybody actually checked the chip revision / full `get-security-info` output on the original Ariston module, or considered CVE-2019-17391 against it?

I'm not suggesting this is necessarily easier than replacing the module — fault injection obviously requires physical access and additional hardware — but if the WATG really contains vulnerable v1.x silicon, it might theoretically allow keeping the complete original module/PCB and running custom firmware on the original ESP32 rather than replacing the module or transplanting the chip.

It could be an interesting alternative path, especially now that the Ariston UART protocol itself is already understood.

DeDaMrAzR:
I just finished another project and will get back to this asap. I need to wrap it up and clear my queue. 

I know about that CVE and will look into it in more depth.

fabiocguerreiro:
> I just finished another project and will get back to this asap. I need to wrap it up and clear my queue.
> 
> I know about that CVE and will look into it in more depth.

Thank you very much for the reply. I’ll definitely be looking forward to any information you can share once you’ve had the chance to investigate that CVE in more depth.

If it turns out to be possible, I’d also be very interested in developing an ESPHome-based version so I can integrate the water heater directly with my Home Assistant setup and keep everything fully cloud-free.

Thanks again for all the effort you’ve already put into researching the protocol and developing this integration. It’s really appreciated.

PR #2056, opened 2026-03-29
Update obk_config.h

hardwire1010:
custom


openshwprojects:
You can't have two IR libraries at once. You must enable either one or second

mfarrell-dev:
> You can't have two IR libraries at once. You must enable either one or second

Ahhh makes sense, thank you. Learning as i go.

PR #2059, opened 2026-03-29
adds safety watchdog for led thermal throttling

### LINKED ISSUE #2054: Feature request: Thermal throttle for lightbulbs (LN882H, BP5758D)
**Describe the problem**
First of all thank you for this great firmware! It is really impressive what you've done!

Currently I am using it on https://openbekeniot.github.io/webapp/devices/Tuya_W505Z2.html E14 9W bulbs. They work great, but they do get hot 90-130C depending on current configuration and unit. Some units when idle show 22C (at 21C ambient temperature) while others show 50C, and another shows 45C at the same time.

**Firmware:**
- Version 1.18.276
- Device: Tuya E14 9W RGBCCT bulb
- Chip/model: LN882H, BP5758D
- Device config:
```
startDriver BP5758D
PowerSave 2
BP5758D_Current 7 7 7 7 7 // reaches 90C max temp
BP5758D_Map 3 4 2 0 1
```
```
startDriver BP5758D
PowerSave 2
BP5758D_Current 10 10 10 10 10 // reaches 100-110C max temp
BP5758D_Map 3 4 2 0 1
```

**Proposed solution**
Considering that these bulbs get hot and temperature reading is already available it might be useful to implement a safety feature (already present in flashlights for example) called "thermal throttling". This could be enabled/disabled by a flag. Basically user could set a certain temperature threshold i.e 100C. In such case if the temp is > 100C, light bulb brightness would be decreased by let's say 5-10% every 10-60 seconds (best if user could configure these params).

This would greatly improve safety of these bulbs, as they would adapt to environment conditions such as ambient temperature, wear parameters drift, accidental physical covering of light bulbs by a blanket etc. 

I am not really familiar with the project codebase, so I'd be mostly offering help in testing and feedback. If you decide to follow up on this idea I'd really appreciate it.

kamilsss655:
Closes: https://github.com/openshwprojects/OpenBK7231T_App/issues/2054

Tested this with 3 different bulbs, it works really well. My stronger bulbs are throttled less, than the ones that get hot more easily. The constants I've added are sane defaults that I think make sense for the model I tested with, however ideally these could be configured by users themselves. I didn't dive into config implementation so I didn't want it to be part of this PR.

I've enabled it only for LN882H, because that's what I could test with. However it likely could work with other platforms as well.

Notes:
- Additionally I fixed multiple re-definitions of `g_wifi_temperature` - as including `new_common.h` is sufficient and proper approach IMO
- Added low pass IIR filter to LN882H temp readings to smooth it out
  - you might consider porting it to all the other sensors as it really smooths out the signal and helps the throttling algorithm 

Please let me know if you have any questions/suggestions.

### DIFF
--- a/platforms/obk_main.cmake
+++ b/platforms/obk_main.cmake
@@ -51,6 +51,7 @@ set(OBKM_SRC
 	${OBK_SRCS}new_ping.c
 	${OBK_SRCS}new_pins.c
 	${OBK_SRCS}rgb2hsv.c
+	${OBK_SRCS}safety/safety.c
 	${OBK_SRCS}tiny_crc8.c
 	${OBK_SRCS}httpclient/http_client.c
 	${OBK_SRCS}httpclient/utils_net.c
--- a/platforms/obk_main.mk
+++ b/platforms/obk_main.mk
@@ -66,6 +66,7 @@ OBKM_SRC  += $(OBK_SRCS)new_common.c
 OBKM_SRC  += $(OBK_SRCS)new_ping.c
 OBKM_SRC  += $(OBK_SRCS)new_pins.c
 OBKM_SRC  += $(OBK_SRCS)rgb2hsv.c
+OBKM_SRC  += $(OBK_SRCS)safety/safety.c
 OBKM_SRC  += $(OBK_SRCS)tiny_crc8.c
 OBKM_SRC  += $(OBK_SRCS)httpclient/http_client.c
 OBKM_SRC  += $(OBK_SRCS)httpclient/utils_net.c
--- a/src/hal/espidf/hal_main_espidf.c
+++ b/src/hal/espidf/hal_main_espidf.c
@@ -17,7 +17,6 @@ void Main_OnEverySecond();
 void HAL_BTProxy_PreInit(void);
 void HAL_BTProxy_OnEverySecond(void);
 #endif
-float g_wifi_temperature = 0;
 
 #if !CONFIG_IDF_TARGET_ESP32 && !PLATFORM_ESP8266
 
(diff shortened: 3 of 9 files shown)


MaxineMuster:
since the values are "hardcoded" for the moment and you have the bulbs ;-):

maybe you could try my try for a simple "config" command?

grafik

there's probably much room for improvement, but as a start ...

[config.diff.txt](https://github.com/user-attachments/files/26352400/config.diff.txt)

kamilsss655:
Hey @MaxineMuster these look good. I don't see any changes to the throttling logic, so if the command stuff (which I am unfamiliar with) is ok it will likely work.  :+1: 

I'd imagine what you've done to be a natural progression of the initial thermal throttling implementation, so I view your changes as positive addition. I am not the maintainer of this repo, but I'd suggest you follow-up with a PR once this one get's merged. I won't be around to test these soon, but the code looks good to me.

WDYT @openshwprojects ?

openshwprojects:
It's very good, but it does not justify separate directory. We need a file for that. Either next to newLEDDriver or just make it a driver, put entry in drv_main.c, call it LEDThrottle or anything, etc. So basically just rename/organization, so it's more intuitive. 

While seeing "safety" name myself, i would never guess it's a LED dimmer limit.

kamilsss655:
Thanks for the review and the feedback — glad you find the feature useful.

I understand the concern about naming and placement. At the moment I’m already using this in my setup and don’t really have the bandwidth to rework the structure.

If you’d prefer it organized differently (e.g. merged into an existing driver or renamed), feel free to adjust it as you see fit. I’m happy for the contribution to be adapted to match the project’s conventions.

Let me know if anything functional needs clarification from my side.

PR #2076, opened 2026-04-05
Implement logging to LFS 

MaxineMuster:
Implement logging to LFS feature with two commands:

`logStartup2lfs [seconds- default 10]` &emsp; Enable writing startup log to lfs for this number of seconds - on every start(!).
&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;Disable with `logStartup2lfs 0`
`log2lfs [seconds - default 10]`	&emsp;&emsp;&emsp;&emsp;	write messages to LFS for this number of seconds. 
&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;Might be stopped before by calling `log2lfs 0`

For now only tested on W800 and Linux simulator

Enabled for testing on all platform using LFS


MaxineMuster:
Bildschirmfoto

Bildschirmfoto

NonPIayerCharacter:
Must Linux be 32-bit? We just got rid of apt-get's from workflows.

MaxineMuster:
It's definitely not a "must".

I mistakenly assumed my change had caused the error messages about "incorrect offsets" in mainConfig_t, but later learned, the issue was not my change but present for some time and resulting from the 64-bit compiler, which uses larger types, more bytes and larger alignment/padding.

You can work around this by using fixed 4‑byte types (which would require several #if/#else branches), or (my approach) by building as 32‑bit, which requires additional libraries.

For the GitHub build it doesn't matter at all — the config for test ruuns is always "empty" at startup and filled with defaults — you could simply omit the tests to avoid the "confusion".

For a locally compiled simulator it also doesn't really matter — I still haven't succeeded in making a "restart", so the config can never actually be used there either.

I'll split this off — just log2lfs in this PR, with no changes to the Linux simulator.

this "issue" can be discussed maybe in another PR.

Side note — it might be more sensible to revise the "mainConfig_t" structure to make offsets consistent, e.g. by standardizing padding after the pin fields (which vary by MCU) or using a uniform "maximum" pins structure. I tried to add comments showing the actual situation with "shifted" offsets and a general possibility to run the test also on other platforms, not only Linux/Windows.

MaxineMuster:
Made new/cleaner version in #2105 - closed this PR

PR #2078, opened 2026-04-06
Add PulseClock driver

nerdb0y:
Old site-wide clock installations used electrical pulses to synchronously advance (or reverse) clocks at the appropriate rates. These are sometimes known as secondary or slave clocks, eg: Seiko MC-009

This driver allows generation of appropriate pulses to actuate these clocks.


openshwprojects:
very nice! Would you also be able to write a short guide on how to use this driver?
https://www.elektroda.com/rtvforum/forum51.html

We can merge it, but it would need to be disabled in obk_config by default

MaxineMuster:
Yes, quite a cool idea!

The only remark from my side: 
Maybe it turns out I'm mistaken, but I would assume the number of users to be not too high.
Therefore using 6 IOroles is filling up the list with entries not used by the majority of users making the dropdown list long.

Since [advertisement] as long as we don't hide unused IOroles like in #1826 [/advertisement] I would suggest to go the hard way setting the pins by command. 
I know, 6 additional arguments are quite a challenge, but possible (maybe even by introducing a second command to set the pins (or movee the actual "optional" args from the startdriver command  to a seperate one setting the pins with startdriver.)

Personally I like some "speaking" args, especially if there are many, like `-Fwd P1 -FwdNeg p3 -FwdPos p4 -Rev p10 -RevPos p11 -RevNeg p17`
	
With this approach you could even fit everything in one `startdriver`, because though it's long, you always know which parameter you are setting (and the order of arguments doesn't matter)
 
` -Fwd P1 -FwdNeg p3 -FwdPos p4 -Rev p10 -RevPos p11 -RevNeg p17 -Res 60 -PulseOffs 0 -PulseMS 500 -MaxSec 43200`

is equal to 
`-PulseMS 500 -Fwd P1 -Rev p10 -FwdNeg p3 -PulseOffs 0 -RevNeg p17  -FwdPos p4  -RevPos p11 -Res 60  -MaxSec 43200`

[advertisement] I still hope, my PR  #2024 will be accepted [/advertisement] but this is also possible in your own code.

PR #2085, opened 2026-04-12
Disable Ping watchdog UI

giedriuslt:
Functionality is disabled, but UI is not. This also disables UI, reducing confusion and saving space


divadiow:
how come you closed ?

PR #2089, opened 2026-04-15
Mosquitto 7.0.0 fix?

NonPIayerCharacter:
Use short name as mqtt client id


openshwprojects:
Shall we merge it?

divadiow:
any truth to this analysis?

`**CLOSE: INCORRECT SEMANTICS** | Code reviewed | Forces short device name as MQTT client ID and ignores the configured client ID, creating collision risk without evidence that Mosquitto 7 requires it.`

PR #2093, opened 2026-04-19
add XR806 INT DCDC variant

divadiow:
INT DCDC required for 3 of the 5 XR806 devices we've seen surface on Elektroda

image


divadiow:
draft. currently set for 64mhz flash speed for testing on 1 device in https://www.elektroda.com/rtvforum/viewtopic.php?p=21887826#21887826

majority appear to be 96Mhz
image

PR #2101, opened 2026-05-03
initial display experiment

openshwprojects:


DeDaMrAzR:
at least two different displays are on order for me and are coming this week for testing - will report here

NonPIayerCharacter:
Are there any source changes in these libraries?
If not, why not add them as submodules?

DeDaMrAzR:
some tough time after.... Elecrow 7" LCD running OBK!

image

PR #2104, opened 2026-05-07
Update http_fns.c

mfarrell-dev:
HASS: Add support for 4 PWM channel discovery (LIGHT_RGBCW)

This enables automatic Home Assistant discovery for RGBW devices like the Arlec Grid Connect Smart Puck Lights.

### DIFF
--- a/src/httpserver/http_fns.c
+++ b/src/httpserver/http_fns.c
@@ -2125,7 +2125,7 @@ void doHomeAssistantDiscovery(const char* topic, http_request_t* request) {
 	}
 	else if (pwmCount > 0) {
 		if (pwmCount == 4) {
-			addLogAdv(LOG_ERROR, LOG_FEATURE_HTTP, "4 PWM device not yet handled");
+			dev_info = hass_init_light_device_info(LIGHT_RGBCW);
 		}
 		else if (pwmCount == 3) {
 			// Enable + RGB control


mfarrell-dev:
# Device Support Report: Arlec Grid Connect Smart RGBW Puck Lights (4-Pack)

**Product:** Arlec Grid Connect Smart RGBW Puck Lights  
**Sourcing:** Bunnings Australia  
**Chipset:** BK7231N  

## Issue Summary
The device uses a 4-PWM configuration (RGB + Warm White). Previously, the OpenBeken firmware lacked automatic Home Assistant discovery support for exactly 4 PWM channels, resulting in an "unhandled" error log and no automatic entity creation. Additionally, manual configurations using `schema: json` failed because the firmware's default state updates are plain text.

## Firmware Patch
A patch was applied to `src/httpserver/http_fns.c` to treat 4-PWM devices as `LIGHT_RGBCW` during the Home Assistant discovery process. This allows the device to be automatically recognized as a color-capable light with white channel support.

### Code Change:
```c
// File: src/httpserver/http_fns.c
// In function: doHomeAssistantDiscovery

        else if (pwmCount > 0) {
                if (pwmCount == 4) {
-                       addLogAdv(LOG_ERROR, LOG_FEATURE_HTTP, "4 PWM device not yet handled");
+                       dev_info = hass_init_light_device_info(LIGHT_RGBCW);
                }
                else if (pwmCount == 3) {

openshwprojects:
Wait a sec. Is LIGHT_RGBCW already supported?

mfarrell-dev:
> Wait a sec. Is LIGHT_RGBCW already supported?

From all my tests, I couldn't get it to show up via MQTT as a RGBCW light. This was the only way I could get it to function. Happy to run further verifications before merging, of course.

mfarrell-dev:
> > Wait a sec. Is LIGHT_RGBCW already supported?
> 
> From all my tests, I couldn't get it to show up via MQTT as a RGBCW light. This was the only way I could get it to function. Happy to run further verifications before merging, of course.

Just to be absolutely thorough and rule out any config issues on my end, I ran a clean test using the current unpatched release (v1.18.288).

My configuration is set properly with my flags at 138513683487 (which includes Flag 30 for Tele publishes and Flag 37). During this test, the device connects to the broker perfectly and publishes all of its diagnostic sensors.

However, when it attempts the light discovery process, the unpatched firmware hits the if (pwmCount == 4) limitation. It intentionally halts the payload generation and throws this exact error in the internal device console:

Info:HTTP:HASS counts: 0 rels, 4 pwms, 0 inps, 0 excluded
Error:HTTP:4 PWM device not yet handled

Because it halts there, no light payload is ever sent to the broker. With my patch applied, it successfully maps to the LIGHT_RGBCW struct, bypasses this error log, and successfully fires the unified HA payload.

NonPIayerCharacter:
What about flag 24? It's usually used in those cases. You would have to reassign you cold white channel to warm (channel 4) for it to work though.

mfarrell-dev:
> What about flag 24? It's usually used in those cases. You would have to reassign you cold white channel to warm (channel 4) for it to work though.

Ah, you are completely right! I totally missed Flag 24 for this one.

I just tested it by remapping the Cool White pin to PWM;4 and toggling Flag 24, and the discovery payload generated perfectly without the patch.

Closing this PR since the native logic handles it. Thanks for the heads up!

mfarrell-dev:
So I ran some more testing with Flag 24. While it does successfully fix the discovery payload, it noticeably changed how the physical lights behave. This actually helped me realize that this hardware is RGBW (Warm White only), not RGBCW!

Enabling Flag 24 and remapping to PWM;4 is definitely the correct 'standard' configuration to get this board to pass discovery. The weird 'Cool White' behavior I noticed afterward is simply the firmware switching from driving the physical Warm LED to using RGB-emulation to fake the missing Cool White channel.

A photo of the LED PCB in these pucks confirming the layout can be seen in this post: [https://www.elektroda.com/rtvforum/topic4173127.html](https://github.com/openshwprojects/OpenBK7231T_App/pull/2104)

Given this pure RGBW layout, is there another existing workaround to avoid the RGB-emulation? Alternatively, could we treat this as a feature request for a new Flag? (e.g., A flag that tells the firmware 'This is a 4-channel RGBW device: disable CW emulation and just map the entire Home Assistant white temperature slider to the single Warm White pin').

Happy to keep this PR closed, but I'd love to hear your thoughts on how to best handle pure RGBW setups!

PR #2112, opened 2026-05-22
Refactor SHIFTREGISTER management for multiple instances

D-Fed11:
Support Multiple Independent Shift Register Instances

## Motivation
This change was driven by a real-world device that uses 2 separate 74HC595 shift register chips connected to different GPIO pins on the BK7231N. The previous implementation only supported a single shift register instance, causing the second chip to be ignored when `startDriver ShiftRegister` was called a second time.

## Changes
Modified the shift register driver to support up to 4 independent shift register instances, each with separate GPIO pins and channel mappings.

### Files Modified

1. **src/driver/drv_shiftRegister.c**
   - Replaced single global variables with a struct array (`g_shiftRegs[MAX_SHIFT_REGISTERS]`)
   - Updated `Shift_Init()` to automatically find the next available slot instead of overwriting global state
   - Updated `Shift_OnChannelChanged()` to iterate through all active instances and route channel changes to the correct chip

2. **src/driver/drv_main.c**
   - Modified `DRV_StartDriver()` to allow ShiftRegister to be re-initialized multiple times
   - Added special handling so subsequent `startDriver ShiftRegister` calls create new instances instead of being rejected

## Usage Example
// First 74HC595 chip (pins 2, 3, 4, channels 10-17)
startDriver ShiftRegister 2 3 4 10 1 1 0

// Second 74HC595 chip (pins 19, 20, 21, channels 18-25)
startDriver ShiftRegister 19 20 21 18 1 1 0


openshwprojects:
It is very useful idea but we should do it without "hacking" main drivers system:

```
			// ShiftRegister supports multiple independent instances
			if (g_drivers[i].bLoaded && !stricmp("ShiftRegister", name)) {
				// Allow re-initialization for multiple instances
				if (g_drivers[i].initFunc) {
					g_drivers[i].initFunc();
				}
				addLogAdv(LOG_INFO, LOG_FEATURE_MAIN, "Initialized new ShiftRegister instance.\n");
				bStarted = 1;
				break;
			}
```

PR #2115, opened 2026-05-31
set BK7238 LFS to 0x1D8000

divadiow:
otherwise OTA always wipes LFS


divadiow:
https://www.elektroda.com/rtvforum/viewtopic.php?p=21913160#21913160

PR #2116, opened 2026-06-01
GD32VW553 support

NonPIayerCharacter:


openshwprojects:
That's a great addition!

PR #2134, opened 2026-06-16
Current and Power limit scripting for power metering

DeDaMrAzR:
Some of the command examples :

```
- PowerLimitStatus
- PowerLimit <relay_channel> <maxCurrentA> <maxPowerW> <delaySeconds>
```

This is WIP and I will appreciate any feedback, currently enabled and tested on BK7231N and T platforms.

### DIFF
--- a/src/driver/drv_bl_shared.c
+++ b/src/driver/drv_bl_shared.c
@@ -10,6 +10,7 @@
 #include "../logging/logging.h"
 #include "../mqtt/new_mqtt.h"
 #include "../hal/hal_ota.h"
+#include "../httpserver/hass.h"
 #include "drv_local.h"
 //#include "drv_ntp.h"
 #include "drv_deviceclock.h"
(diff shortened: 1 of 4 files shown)


DeDaMrAzR:
`PowerLimit <channel> <maxCurrentA> <maxPowerW> [delaySeconds]`

Example - `PowerLimit 1 10 2000 2`

This means:

channel = 1
max current = 10 A
max real power = 2000 W
delay = 2 seconds

Check current status cmd - `PowerLimitStatus`

Example expected log:

PowerLimit: ch=1 current=10.000A power=2000.0W delay=2s tripped=0 override=0 trips=0

PowerLimitReset cmd - This clears the trip latch and rearms protection. It does not turn the relay back on automatically. If the load is still over the configured limit and the relay is turned on again, it should trip again.

Enable/dissable override cmd's:

```
PowerLimitOverride 1

PowerLimitOverride 0
```

Override allows the protected channel to stay on while keeping trip evidence. Disabling override while still tripped should force the channel off again on the next metering update.

You can also add this to your autoexec for manual reset, override and restore

```
addEventHandler OnHold 11 PowerLimitReset
addEventHandler OnDblClick 11 backlog PowerLimitOverride 1; setChannel 1 1
addEventHandler On3Click 11 PowerLimit 1 10 2000 2
```

Meaning:

(Btn pin role is on defined on pin 11 in this case)
OnHold = acknowledge/reset trip
OnDblClick = enable service override and turn channel 1 on
On3Click = restore normal limit


and with this PR I exposed 3 new "sensors" 
image

DeDaMrAzR:
Right now for N and T platforms only, expand if you see fit. 

`DelaySeconds` is optional and defaults to 0 if not set

PR #2137, opened 2026-06-17
Add event rate controls and thresholds for hwl8112

wsot:
The aim of this PR is to add controls to avoid the HLW8112 pushing MQTT output every second, potentially overloading receives of the events.

It's in draft because I'm just starting to experiment with it, and it's a PR because I wanted to attempt a build. (I'm currently trying to get builds going locally or using my own Github Action minutes.

Maintainers: feel free to close this if it is getting in the way; I'll close it myself if I don't make some sensible headway.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -54,97 +54,97 @@ jobs:
           echo "sha_short=SHA-$(git rev-parse --short=12 HEAD)" >> $GITHUB_OUTPUT
           echo "version=${VERSION}" >> $GITHUB_OUTPUT
 
-  build2:
-    name: Build Simulator
-    needs: refs
-    runs-on: windows-2022
+  # build2:
+  #   name: Build Simulator
+  #   needs: refs
+  #   runs-on: windows-2022
 
-    steps:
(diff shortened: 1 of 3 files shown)


DeDaMrAzR:
have you seen these script commands from here - https://github.com/openshwprojects/OpenBK7231T_App/blob/main/docs/commands.md

image

wsot:
Thanks @DeDaMrAzR - I'll take a closer look at those script commands. (My challenge is not actually the MQTT messages, but the total number of readings I'm ending up storing in Home Assistant).

I also have local builds going now, so I'll close this PR until I've looked through those docs in more detail, and _if_ a change is potentially needed I'll put together one that builds before opening a PR.

Thanks for the help!

wsot:
Doesn't look like changing those broadcast intervals makes a difference: they seem to control the 'send everything' broadcast interval (where all valid channels for sending are sent), but the hlw8112_RunEverySecond triggers an updated of every channel, and each of those updates seems to trigger its own MQTT sent - so even if the mqtt_broadcastinterval is set to 500, there are still updates sent for every HLW8112 channel every second.

PR #2143, opened 2026-06-23
Fix webui slider behavior

DeDaMrAzR:
sliders work after releasing not mid drag and value is shown above them

image


DeDaMrAzR:
image

Color picker fix introduced.

divadiow:
much better ux. currently page does glitching and mouse drag gets dropped sometimes as sliders are moved. seems good to merge @openshwprojects ? 

image

PR #2146, opened 2026-06-28
rename xr806 output file

divadiow:


divadiow:
https://www.elektroda.com/rtvforum/viewtopic.php?p=21929217#21929217

PR #2149, opened 2026-06-30
Update MQTT and WiFi configuration parameters

giannajuncal:


divadiow (commented):
this contains your credentials?

I suggest close and purge delete if so

PR #2162, opened 2026-07-13
Add 4-PWM RGBW LED mode

lumixen:
## Summary

Adds Flag 52 (OBK_FLAG_LED_4PWM_RGBW_MODE) for 4-PWM RGB + independent White LED strips, exposed as a single HA light entity with RGB/White color modes, including led_enableWhite command, led_colorMode MQTT topic, HA discovery via template schema, Web UI mode selector with dynamic RGB picker hide/show, and selftest coverage.

I have tested and am using these changes for the Magic Home RGBW controller:

RGB Mode:

White Mode:

HA light entity:

Essentially this PR implements a request for RGB+separate white channel from https://www.elektroda.pl/rtvforum/topic4003384.html

### MQTT structure

| Direction | Topic | Value |
|-----------|-------|-------|
| Set RGB | `cmnd/<device>/led_basecolor_rgb` | `RRGGBB` hex |
| Set White | `cmnd/<device>/led_enableWhite` | any value (ignored) |
| Set enable | `cmnd/<device>/led_enableAll` | `0` or `1` |
| Set brightness | `cmnd/<device>/led_dimmer` | `0`–`100` |
| Get RGB state | `<device>/led_basecolor_rgb/get` | `RRGGBB` hex |
| Get color mode | `<device>/led_colorMode/get` | `"rgb"` or `"white"` |
| Get enable | `<device>/led_enableAll/get` | `0` or `1` |
| Get dimmer | `<device>/led_dimmer/get` | `0`–`100` |

HA discover payload:
```json
{
  "rgb_cmd_tpl": "{{'#%02x%02x%02x0000'|format(red,green,blue)}}",
  "rgb_val_tpl": "{{ value[0:2]|int(base=16) }},{{ value[2:4]|int(base=16) }},{{ value[4:6]|int(base=16) }}",
  "rgb_stat_t": "~/led_basecolor_rgb/get",
  "rgb_cmd_t": "cmnd/<device>/led_basecolor_rgb",
  "whit_cmd_t": "cmnd/<device>/led_enableWhite",
  "whit_scl": 100,
  "clrm_stat_t": "~/led_colorMode/get",
  "stat_t": "~/led_enableAll/get",
  "cmd_t": "cmnd/<device>/led_enableAll",
  "bri_stat_t": "~/led_dimmer/get",
  "bri_cmd_t": "cmnd/<device>/led_dimmer",
  "bri_scl": 100
}
```

@openshwprojects, thank you for all your great work in making this project possible. Please let me know your feedback and whether exposing this feature as a new flag is an acceptable approach.


openshwprojects:
Seems nice, I replied there:
https://www.elektroda.com/rtvforum/viewtopic.php?p=21944154#21944154

lumixen:
Hi @openshwprojects , thank you for the feedback!

> -An old saved colour-temperature value is reused as the saved white-channel value once Flag 52 is enabled, without an explicit migration.
> -A previously saved CW/temperature mode is not converted automatically into the new White mode.

I intentionally didn't add migration logic from other flag combinations as it would complicate the code and maintenance. I anticipate that the device is configured once, and whenever the hardware or LED strip configuration changes, it should be a fresh configuration.

> -The implementation expects white on logical channel 4, so an existing contiguous RGBW assignment using channels 0, 1, 2 and 3 would not work without reconfiguration.

Let me know if that's something we should handle differently. I adjusted this logic specifically for my hardware.

> -Some alternative commands and routes, such as HSB controls, led_temperature, CT, led_basecolor_rgbcw and Device Groups, can still select the older RGB/CW/All modes without necessarily keeping the new

I can add validation logic to those commands to prevent state modifications if flag 52 is enabled.

> -Home Assistant sends a value to the white command topic, but `led_enableWhite` currently ignores that value and relies entirely on the normal dimmer setting. 

That's intentional. Home Assistant sends a value to the white command topic only once when it transitions from RGB mode to White mode. After that, it never publishes a white value again unless it transitions to RGB and back to White mode again. Brightness is always controlled by the `bri_cmd_t` topic regardless of mode. I've added an extended comment to explain this behavior.

ryan-ronnander:
Only spotted this after opening #2236, which overlaps this PR somewhat (much much smaller).

But if I'm reading this right for your case:

- A dummy 5th PWM (the maintainer's suggestion [here](https://www.elektroda.com/rtvforum/topic4012186.html#20801045)) gets you HA discovery
- #2236 makes `CTRange 500 500` collapse the temperature slider to a single white point instead of dividing by zero, so your white channel drives at full.

Think of it like an RGBCW bulb with one locked color temp that cannot do white + RGB at the same time, only one mode or the other. Simple, but works.

This PR is likely a better true fix though for PWM bulbs. They get a proper white mode and discovery for a bare 4-PWM device. My RGBW bulb uses an SM2135 and wouldn't benefit from this PR unfortunately, since `pwmCount` is 0 for a driver-chip bulb and the `pwmCount == 4` gate never fires, maybe it could end up covering more than just PWM?

I don't see why both PRs can't coexist though, the code paths tackle similar problems in entirely different ways.

PR #2188, opened 2026-08-18
drv_battery: keep the configured voltage divider, and don't publish ADC errors as a voltage

DrDoof:
Two things I ran into getting battery reporting working on a camera board. They're
unrelated, but both live in `Batt_Measure()`, so splitting them into separate PRs would
just create a conflict.

The first one: `Battery_Setup` takes a divider argument, and `Batt_Measure()` throws it
away again unless a `BAT_Relay` pin happens to be assigned.

```c
if (PIN_FindPinIndexForRole(IOR_BAT_Relay, -1) == -1 && PIN_FindPinIndexForRole(IOR_BAT_Relay_n, -1) == -1) {
    g_vdivider = 1;
}
```

My board has the divider soldered on permanently with no relay to switch it. I first
wrote that I didn't think this was unusual, and I've since been put right on that: most
battery-powered designs use a dedicated FET to switch the measurement on and save the
current, so mine is the odd one out. Either way, `Battery_Setup 3400 4100 3 2400 4096`
reported a third of the real voltage, and nothing in the log hinted at why. Took me
longer than I'd like to admit to spot, since the rest of the arithmetic checks out fine.

I kept the fallback for anyone who never passes a divider at all. It only stops
overriding a value that was set on purpose.

The other one: `HAL_ADC_Read` signals failure with a negative return value. On Beken
that's -1 for a pin with no channel behind it, -2 if the ADC won't open, -3 on a
conversion timeout. Nothing checks for it, so it gets scaled and published to MQTT and
the index page. One failed conversion shows up as -3.5 mV sitting in among normal
4000 mV readings, and drops the percentage to 0 for that cycle.

Now it keeps the previous value and logs the failure instead.

Tested on BK7252N.

### DIFF
--- a/src/driver/drv_battery.c
+++ b/src/driver/drv_battery.c
@@ -29,6 +29,9 @@ static void Batt_Measure() {
 	if (g_vdividerFromUser == false
 		&& PIN_FindPinIndexForRole(IOR_BAT_Relay, -1) == -1
 		&& PIN_FindPinIndexForRole(IOR_BAT_Relay_n, -1) == -1) {
+		if (g_vdivider != 1) {
+			ADDLOG_INFO(LOG_FEATURE_DRV, "DRV_BATTERY: no relay pin and no divider given, ignoring divider %f and measuring directly", g_vdivider);
+		}
 		g_vdivider = 1;
 	}
 	// if divider equal to 1 then no need for relay activation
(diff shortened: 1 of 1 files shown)


DeDaMrAzR:
> My board has the divider soldered on permanently with no relay to switch it, which I
> don't think is unusual.

On most battery operated devices that were dealt with there is a dedicated FET(relay) for turning battery measurement on so without that condition being triggered there is no measurement. Design choice I guess for lowering power consumption and since many of them (if not all) are not running all the time it makes sense. Your device is unusual 🙂 in that regard, keep that in mind.

DrDoof:
Fair correction, I'll take that. I based "not unusual" on the one board in front of me,
which isn't much of a sample, and the power argument for switching the divider makes
obvious sense on something that spends most of its life asleep.

It doesn't change what the patch does though, and I don't think it puts the common
design at any risk. The old behaviour is kept everywhere except one case:

- relay pin assigned: the condition was already false, so nothing changes at all
- no relay pin and no divider argument: still falls back to 1, same as before
- no relay pin but `Battery_Setup` was given a divider: now it's used instead of being
  dropped

So the only path that behaves differently is the one where someone passed a value that
was previously thrown away without a word in the log. That was what caught me out, not
the fallback itself, which is a sensible default.

If you'd rather not carry the extra flag for a rare board, an alternative would be to
leave the override in place and just log when a configured divider gets reset. I'd
still lean toward honouring an argument the user explicitly passed, but I'm happy either
way and can rework it.

DeDaMrAzR:
I was not criticizing your implementation - more over I can even agree with it. 

Couple of things to pay attention to, TAB indent only, no breaking changes (unless 100% needed) and describe the PR as much as possible ( as you already did 🙂 ).

Feel free to tag @openshwprojects, me, @NonPIayerCharacter, @divadiow or any of the contributors so we can check and review proposed changes and implement them. One "limiting" factor may be a test on the single device, OBK supports A LOT of platforms and chips so testing it and verifying behavior is a process just be patient. Or feel free to join us in our chip hoarding game 😁

DrDoof:
Understood, and thanks for spelling the conventions out, that's more useful than a
review comment after the fact.

I went back over all of the branches I've got open rather than just this one. Two
of them had spaces on lines I'd added, in the ADC read where the surrounding code
is already mixed, and I've pushed fixups for both. This one and the listen backlog
were already tabs throughout.

Point taken on the single device too. I've tried to keep each of these to one file
and one problem so they're easier to reason about without the hardware in front of
you, and I'm in no hurry.

divadiow:
`**FIX CLEANUP PATH, THEN MERGE** | Code reviewed | Both intended fixes are valid, but the ADC-error return occurs before the existing relay/FET switch-off. One failed conversion can leave a switched measurement divider powered. Use a common cleanup path and make divider-state reset semantics explicit`

DrDoof:
You're right, and it's an unfortunate one given the context: the relay exists to save current, and my error path was the one case that left it drawing some.

The read happened between switching the divider on and switching it back off, so a failed conversion returned straight past the switch-off and the divider stayed powered until the next measurement came round. I've moved the switch-off to sit immediately after the read, so both paths go through it and the divider is off as soon as the sample is taken, whether or not the sample was any good.

On the reset semantics, `g_vdividerFromUser` is sticky by design and I'd like to keep it that way, since omitting an argument means "leave it alone" for vref and adcbits too and it would be odd if the divider alone reverted. What was genuinely implicit is the other direction: the fallback silently discarding a configured divider is exactly what cost me the afternoon that started this PR, and it never said so. It does now. The log can only fire once, because `g_vdivider` is 1 from that point on.

PR #2190, opened 2026-08-19
adc: a timed out read leaves the SARADC open and poisons the next one

DrDoof:
Three problems in the Beken ADC read. They are all in the same file and they
interact, so one PR.

**Correction to what this description said before.** I originally wrote that the
permanent failure comes from the device never being closed. That is backwards. It
comes from the device being closed too many times, and I have since measured it.
The rest of the original description stands.

### A read that times out skips the cleanup

The interrupt handler does the cleanup for a successful read: once the buffer fills
it calls `temp_single_get_disable()` and signals the semaphore. A read that times
out skips all of that. It returns -3 with the device still open, and if the
conversion lands a moment after the 1 second deadline the handler signals a
semaphore nobody is waiting on any more. The next read then returns immediately on
the leftover count and takes `pData` while the fresh conversion is still in flight,
so it gets the previous one's sample, on whatever channel that was.

### The handler can close the device more than once

`saradc_isr` calls `p_Int_Handler` on every interrupt where the sample count has
reached the buffer size. That test sits outside the FIFO drain loop, and the
descriptor keeps a full buffer after the conversion, so any later SARADC interrupt
runs the handler again. Both driver variants are written this way, `saradc.c` for
BK7231/7231N/7231U/7236/7271 and `saradc_bk7238.c` for 7238/7252N.

Every run reaches `ddev_close`, and `ddev_close` decrements `use_cnt` before it
tests it. `use_cnt` is `UINT32`, so one close too many wraps it to `0xFFFFFFFF`, the
zero test never passes again, and the device is left in `DD_STATE_OPENED` for good.
`ddev_open` then takes its already-open branch, which hands back a valid handle
without calling `saradc_open`. No conversion is ever started, so every read waits
out the full second and returns -3, until the board is power cycled.

Worth noting because it changes how the first fix has to be written: the close this
PR adds on the timeout path can itself be the extra one. The handler clears
`tmp_single_hdl` only after `ddev_close` returns, so a timeout landing in that
window still sees a valid handle and closes a second time.

So `temp_single_get_disable()` now takes the handle out of `tmp_single_hdl` with
interrupts masked and returns early if it has already been taken. That makes it safe
from both sides and lets the timeout path call it unconditionally.

### Nothing holds a second reader off

`tmp_single_desc`, `tmp_single_hdl` and `tmp_single_semaphore` are file statics. Two
readers on different tasks overwrite each other's channel in the descriptor, both
block on the single semaphore, and one walks away with the other's sample. There are
already four callers in tree: `Main_OnEverySecond` loops over every `IOR_ADC` pin,
plus `drv_battery`, `drv_adcSmoother` and `drv_adcButton`. It only takes one read
from outside the main loop.

I flagged when opening this that the mutex was created lazily and left a window on
the very first call. That is fixed here too: both handles are now created under one
interrupt mask, which is what the rest of the file already does and runs at most
once. It seemed a better fit than restructuring `HAL_ADC_Init`, which is empty and
only runs for pins carrying an ADC role, but say the word if you'd rather have it
there.

### Measured

On BK7252N, repeating a read across channels 0, 1, 2, 3, 8 and 9:

- before: two good conversions, then every later read failed until a power cycle
- after: 18 of 18 good

The failure shows as -2 rather than -3 on the tree I measured it on, because that
tree also checks the status `ddev_open` leaves untouched on its already-open branch.
On this branch the same wedge times out instead. Same cause either way.

Only device I have, so a second pair of eyes on the interrupt reasoning would be
welcome.

### DIFF
--- a/src/hal/bk7231/hal_adc_bk7231.c
+++ b/src/hal/bk7231/hal_adc_bk7231.c
@@ -159,6 +159,7 @@ int HAL_ADC_Read(int pinNumber)
     UINT32 ret;
     int result;
 	int channel;
+	int value;
 
 	channel = gpioToAdc(pinNumber);
 
@@ -170,17 +171,25 @@ int HAL_ADC_Read(int pinNumber)
         result = rtos_init_semaphore(&tmp_single_semaphore, 1);
         ASSERT(kNoErr == result);
     }
-    
+
+	while(rtos_get_semaphore(&tmp_single_semaphore, 0) == kNoErr) {
+	}
+
 	if(temp_single_get_enable(channel)==SARADC_FAILURE) {
 
 		return -2;
 	}
-    
+
     ret = 1000; // 1s
     result = rtos_get_semaphore(&tmp_single_semaphore, ret);
     if(result == kNoErr) {
-        return tmp_single_desc.pData[0];
-    } 
-    return -3;
+        value = tmp_single_desc.pData[0];
+    } else {
+		if(tmp_single_hdl != DD_HANDLE_UNVALID) {
+			temp_single_get_disable();
+		}
+		value = -3;
+	}
+    return value;
 }
 
--- a/src/hal/bk7231/hal_adc_bk7231.c
+++ b/src/hal/bk7231/hal_adc_bk7231.c
@@ -92,20 +92,25 @@ static beken_mutex_t tmp_single_mutex = NULL;
 
 static void temp_single_get_disable(void)
 {
-    UINT32 status = DRV_SUCCESS;
-    
-    status = ddev_close(tmp_single_hdl);
-    if(DRV_FAILURE == status )
-    {
-        //TMP_DETECT_PRT("saradc disable failed\r\n");
-        //return;
-    }
-    saradc_ensure_close();
+    UINT32 status;
+    DD_HANDLE hdl;
+    GLOBAL_INT_DECLARATION();
+
+    GLOBAL_INT_DISABLE();
+    hdl = tmp_single_hdl;
     tmp_single_hdl = DD_HANDLE_UNVALID;
-    
+    GLOBAL_INT_RESTORE();
+
+    if(DD_HANDLE_UNVALID == hdl)
+    {
+        return;
+    }
+
+    ddev_close(hdl);
+    saradc_ensure_close();
+
     status = BLK_BIT_TEMPRATURE_SENSOR;
     sddev_control(SCTRL_DEV_NAME, CMD_SCTRL_BLK_DISABLE, &status);
-    
 }
 static void temp_single_detect_handler(void)
 {
@@ -193,9 +198,7 @@ int HAL_ADC_Read(int pinNumber)
     if(result == kNoErr) {
 		value = tmp_single_desc.pData[0];
 	} else {
-		if(tmp_single_hdl != DD_HANDLE_UNVALID) {
-			temp_single_get_disable();
-		}
+		temp_single_get_disable();
 		value = -3;
 	}
 	rtos_unlock_mutex(&tmp_single_mutex);
--- a/src/hal/bk7231/hal_adc_bk7231.c
+++ b/src/hal/bk7231/hal_adc_bk7231.c
@@ -173,13 +173,18 @@ int HAL_ADC_Read(int pinNumber)
 		return -1;
 	}
 
-    if(tmp_single_semaphore == NULL) {
-        result = rtos_init_semaphore(&tmp_single_semaphore, 1);
-        ASSERT(kNoErr == result);
-    }
-	if(tmp_single_mutex == NULL) {
-		result = rtos_init_mutex(&tmp_single_mutex);
-		ASSERT(kNoErr == result);
+	{
+		GLOBAL_INT_DECLARATION();
+		GLOBAL_INT_DISABLE();
+		if(tmp_single_semaphore == NULL) {
+			result = rtos_init_semaphore(&tmp_single_semaphore, 1);
+			ASSERT(kNoErr == result);
+		}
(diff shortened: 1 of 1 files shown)


DrDoof:
Pushed a fixup for the indentation, all the lines this adds are tabs now. The
function is a bit mixed already (`UINT32 ret;` on spaces, `int channel;` two lines
down on a tab) so I'd been matching whatever line I was touching, but I've left the
existing lines alone rather than reformatting around the change.

@openshwprojects @DeDaMrAzR @NonPIayerCharacter @divadiow if any of you get a
moment for this one. It's the ADC read on Beken not cleaning up after a timeout,
and no mutual exclusion between the four callers already in tree. No rush, and I
know a single device isn't much to go on. Happy to explain anything or split it
differently if that helps.

openshwprojects:
This seems nice but I am not able to test it at the moment. @divadiow  do you have bk with potentiometer to try? Before and after, does issue really exist and is fixed by this PR?

divadiow:
I have a small 3-pin P103 potentiometer I've never used before, but automated analysis I'm hoping will give a much more thorough review

code-only analysis
"Core race fixes: good.
Timeout handling: good.
Handle ownership fix: good.
Stale semaphore handling: good.
Lazy mutex/semaphore initialization under globally-disabled IRQs: worth confirming."

after automated testing with this setup, where Codex leverages ESP32-WROOM-32 IOs 25+26 for ADC

```
ESP32-WROOM-32                         BK7231N
================                       =======

GPIO25 / DAC1 ---- 10k ----+---------> P26 / ADC1
                           |
                          10k
                           |
GND -----------------------+

GPIO26 / DAC2 ---- 10k ----+---------> P24 / ADC2
                           |
                          10k
                           |
GND -----------------------+

ESP32 GND ---------------------------- BK GND
USB-TTL GND -------------------------- BK GND

ESP32 GPIO27 ---- 1k ----------------> BK CEN
```

"#2190's HAL mutex appears to do what it intends for HAL callers, but on BK7231N the shared SARADC has real production users outside that mutex. A collision with either temperature path hits the SDK's unusual “valid handle + failure status” already-open behavior; HAL ignores that status and can turn the collision into a persistent timeout state."

with modified OBK build... "with the two non-HAL owners removed, the exact PR HAL produced 50,006 successful HAL requests out of 50,006, with no open-status failures and no SARADC busy failures"

I have the attached instructions for Codex to work on next
[instructions_codex.txt](https://github.com/user-attachments/files/31330723/instructions_codex.txt)

divadiow:
AI:

```
I tested this fairly extensively on a real BK7231N. The normal workload on stock 1.18.302 (which is the exact PR base b9285de) didn't reproduce the problem, but targeted testing did uncover another SARADC ownership case that affects BK7231N.

#2190's mutex does appear useful for HAL-vs-HAL serialization. The complication is that BK7231N also has production users which open the same SARADC directly, outside that mutex — notably the temperature paths.

The failure I reproduced reaches the SDK's ddev_open() already-open behaviour from a legitimate competing owner rather than from the use-count-underflow path described in the PR: ddev_open() returns the existing valid handle with status == DRV_FAILURE, increments the use count, and leaves the current owner's descriptor installed. A handle-only check therefore mistakes the borrowed handle for successful ownership and waits for a conversion on a descriptor that was never installed.

I isolated both active channel-7 temperature paths independently; either could reproduce it. Suppressing both gave 50k rapid reads clean, which is what led to the ownership diagnosis.

I then tested a coordinated app+SDK candidate which:

retains the #2190 HAL mutex;
treats valid-handle + DRV_FAILURE as BUSY and rolls back exactly that borrowed reference with one ddev_close();
retries BUSY with a bounded shared ~1 s deadline;
applies the same status-aware ownership rule to the SDK temperature paths/Tuya ADC adapter;
removes the redundant saradc_ensure_close() after normal ddev_close().

Results on BK7231N:

rapid public HAL test: 50,000/50,000, zero -1/-2/-3, zero wrong levels, worst call 42 ms;
original deterministic barrier suite: 1,349,000 HAL calls, all sequential/2-worker/4-worker/dynamic stages complete, zero negatives, wrong levels, stalls or resets;
stock ADCSmoother/Battery/temperature workload: 1,050 samples complete, normal Battery and temperature activity, no watchdog/panic/reset.

The clean candidate build contains no HIL/diagnostic instrumentation.

So my reading is that #2190 is addressing a real HAL-side problem, but on BK7231N there is an additional cross-API SARADC ownership case which the HAL-local mutex cannot cover by itself. I have clean app and OpenBK7231N SDK diffs from the tested candidate if useful. The SDK change would need to land first, followed by the app submodule update/app-side change.

I left the RF-calibration channel-8 and calibration-CLI direct-open paths as separate hardening follow-ups rather than expanding this runtime fix into those lifecycles.
```

I didn't test BK7231N_ALT, which uses same SDK at BK7252N though

DrDoof:
That `ddev_open()` behaviour explains more than my patch does. Returning a valid handle with `DRV_FAILURE` while still bumping the use count means a handle-only check can't tell "I opened it" from "someone else holds it", and the HAL mutex is blind to the temperature paths anyway since they open the device directly. So the mutex is necessary but not sufficient, and the rollback of exactly that borrowed reference is the part I got wrong.

Yes please to the diffs, and landing the SDK side first is the right order.

One thing on sequencing though. BK7231N_ALT and BK7252N both build against `beken_freertos_sdk`, and the app's submodule pointer for it is still on an April commit, three behind master. So an SDK fix won't reach either target until that pointer moves, regardless of when it merges. I've raised it separately in #2213 since it's already biting something else.

I can cover the BK7252N side once there's something to build against. I have a board where the battery path drives the SARADC, so the contention case is easy to reach there.

PR #2191, opened 2026-08-19
http: listen() is called with a backlog of 0

DrDoof:
The HTTP server calls `listen()` with a backlog of 0.

lwIP doesn't take that literally, `tcp_backlog_set` maps 0 to 1, so the accept queue
holds exactly one connection. When a second SYN turns up while one is still waiting to
be accepted, `tcp_listen_input` bails out early:

```c
#if TCP_LISTEN_BACKLOG
    if (pcb->accepts_pending >= pcb->backlog) {
      LWIP_DEBUGF(TCP_DEBUG, ("tcp_listen_input: listen backlog exceeded ...
      return;
    }
#endif
```

That's a bare return, no SYN-ACK and no RST, so the client sits there until it
retransmits. `TCP_LISTEN_BACKLOG` is on in the Beken ports, so this is live and not
compiled out.

It shows up as a page that takes a beat longer than it should, because browsers open
several connections in parallel for anything with assets on it. The rest of the server
is already set up for that, `max_socks` worker slots and an accept loop that hands each
one off, so the queue in front of it was the only part sized for a single client.

`max_socks` is the value I went with. Worth being precise about what that does and
doesn't buy, since the queue and the workers aren't two separate pools here: the Beken
ports define `MEMP_NUM_TCP_PCB` as `MAX_SOCKETS_TCP` (8, 10 or 12 depending on the lwIP
version), and `max_socks` is that minus one, so both come out of the same allocation
that MQTT and everything else also draws on.

That means the number can't reserve anything. A queued connection is already a PCB, so
the pool stays the ceiling whatever goes in here, and the worker array was already sized
at `max_socks` before this change. What the patch removes is the cap of one, not a
capacity limit.

Tested on BK7252N.

### DIFF
--- a/src/httpserver/new_tcp_server.c
+++ b/src/httpserver/new_tcp_server.c
@@ -247,7 +247,7 @@ static void tcp_server_thread(beken_thread_arg_t arg)
 	}
 	ADDLOG_EXTRADEBUG(LOG_FEATURE_HTTP, "Socket bound on 0.0.0.0:%i", HTTP_SERVER_PORT);
 
-	err = listen(listen_sock, 0);
+	err = listen(listen_sock, (int)max_socks);
 	if(err != 0)
 	{
 		ADDLOG_ERROR(LOG_FEATURE_HTTP, "Error occurred during listen");


divadiow:
ChatGPT analysis, if curious.

```
I see no technical reason to reject this.

I'd rate it roughly:

Correctness: 9/10
Likelihood it addresses the reported symptom: 9/10
Regression risk: low
Patch scope: excellent
PR explanation: essentially correct, with only the resource-occupancy nuance above

In fact, given the server's deliberate 20 ms + 10 ms pauses around accepting/launching clients, listen(..., 0)/effective backlog 1 looks like a genuine pre-existing design mistake. It is surprising it hasn't caused more obvious web-UI flakiness.

So my recommendation would be merge #2191 as written. I would not ask the author to change the code.

The only optional review comment I'd make is to soften “keeps the two in step” because pending-backlog capacity and active worker capacity aren't literally the same resource, and perhaps explicitly say that the larger backlog can occupy additional existing lwIP TCP PCBs under bursts. Neither changes the validity of the fix.

One thing that may be worth investigating separately is whether the 20 ms delay immediately after every accept() is necessary at all. That's making the undersized backlog problem significantly easier to hit, and it smells like legacy defensive timing. I can trace that line through git history and the platform SDKs next to find out why it was added and whether it can safely go.
```

DrDoof:
Thanks for looking at it.

The wording point is fair, and on Beken it's actually tighter than not-the-same-resource.
The ports define `MEMP_NUM_TCP_PCB` as `MAX_SOCKETS_TCP` (8, 10 or 12 depending on which
lwIP version the platform pulls in), and `max_socks` is that minus one, so the queued
connections and the worker connections come out of the same pool, which MQTT and
everything else also draws on.

That cuts both ways and I think it lands on the side of the patch. The backlog number
can't reserve anything, since a queued connection is already a PCB, so the pool stays the
ceiling whatever value goes in. And the worker array was already sized at `max_socks`, so
HTTP could take most of the pool before this change too. The queue depth only shows up
transiently, while connections wait to be picked up. What it does change is the artificial
cap of one, which was the actual problem.

Happy to reword the description to say that rather than implying two separate capacities.

The 20 ms after `accept()` is a good catch. I left it alone because I couldn't tell from
the code whether anything in the SDK depends on the pause, and it felt like a separate
question from the backlog. I'd be interested in what the history turns up.

divadiow:
oh sure, I mean, it's just me relying on AI, so do whatever you think is best. it'd be up to @openshwprojects to review and merge

PR #2192, opened 2026-08-19
BKCharge: driver for the BK7252N on-chip Li-ion charger

DrDoof:
BK7252N carries a Li-ion charger on the die, with CC and CV phases, a trickle current for
deeply discharged cells, an end of charge comparator and a USB presence detector. Nothing
in the tree touches it, so on a battery board there is no way to tell whether charging is
happening short of watching the voltage move.

The state lives in bits 15 to 8 of the charger status register and maps onto the eight
names in `CHARGE_INDEX`. Reading it is a plain register read, no device open and no pin
configuration, so the driver stays out of the way of anything else using the charger.

The index page gains a line with the phase, whether the charger is switched on, the current
limit and USB presence. `charging`, `charge_full` and `usb_power` go out over MQTT on state
change rather than every second, since a charge cycle only passes through a handful of
states. `ChargeChannel` points the charging flag at a channel for use in rules, and
`ChargeState` logs the full breakdown including the raw bits.

`ChargeEnable` is there because the SDK's `charge_init` configures and calibrates
everything and then leaves the charger off. Bit 31 of `ANALOG_CTRL7` stays clear until
something asks for it, so without that command the driver could only ever report an idle
charger. Nothing turns charging on by itself.

Named after the `BKPartitions` pattern, since like that one this is a facility that only
exists on one vendor's silicon. Gated on `ENABLE_DRIVER_BKCHARGE`, defined for BK7252N
only, so other platforms compile none of it.

The SDK side of this landed in NonPIayerCharacter/beken_freertos_sdk#4, so nothing is
outstanding here any more. Leaving the reason on record: without that flag `charge.c`
compiles to an empty object, `charge_init` is never linked in, and this driver would
report reset state rather than anything real. It built either way, which is why the
dependency was worth spelling out at the time.

Worth knowing before anyone reads too much into the status bits: not every BK7252N board
uses this charger. Mine turns out to have a separate linear charger part handling the
battery, so the on-chip block sits unconnected and its status register reads the same
whatever the cell is doing. I found that out by swapping a flat cell for a healthy one and
watching the register not move a single bit, while an ADC reading of the battery rail
tracked the change. That is why the index line says off rather than inventing a phase
while the charger is disabled.

On that board `usb_ready` also stays at 0 with the camera powered over USB and the charger
initialised, which fits the same explanation, the 5 V never reaches the on-chip charger
input. I have no board that uses the internal charger, so the CC and CV paths are reported
as the hardware presents them but I have not watched them run.

Tested on BK7252N.


divadiow:
nice. very interesting.

are you on Elektroda btw? I don't see the obvious possible username in the autocomplete list 

image

openshwprojects (approved):

openshwprojects:
Seems nice, please consider posting a tutorial for it on our forum. I think I'll merge it now.

PR #2199, opened 2026-08-27
Berry: Wire methods never touched the bus

EinsEins11:
While bringing up an I2C accelerometer from Berry I could not get a single byte out of the bus - `read()` returned `nil` every time and `scan()` finished in 57 ms, which is far too fast for 127 addresses with a 5 ms delay each. It turns out none of the `Wire` methods have ever done anything.

Every method starts with `be_iscomptr(vm, 1)`, but on a method call Berry puts the *instance* at index 1, never a comptr. The check is always false, so `scan`, `start`, `stop`, `read`, `read_bytes` and `write` all fall through to `be_return_nil()` without touching the pins.

`init()` has the matching problem from the other side: it reads its arguments from index 1..3, which is `self` plus the first two arguments, and then returns the comptr via `be_return()`. Berry ignores what `init` returns, so the instance never gets a `.p` member - the class declares `.p, var` but nothing ever fills it.

This resolves the `softI2C_t` through the instance`s `.p`, stores it there in `init()`, and shifts `init()`s argument indices by one. Raw comptrs are still accepted, so any direct C-side call keeps working.

One subtlety worth flagging for anyone touching this later: `be_newcomobj()` stores a `BE_COMOBJ`, and `be_iscomptr()` returns false for that. My first attempt guarded the `be_getmember()` result with `be_iscomptr()` and threw the pointer away again. `be_tocomptr()` unwraps both types, so that is what the helper uses.

Tested on BK7238 with a MiraMEMS DA213B on a soft-I2C bus: before the change `Wire(9, 24, 0x26).read()` returned `nil`; after it, register 0x01 reads back 0x13 (the chip ID), the accelerometer can be configured, and its motion interrupt drives a GPIO as expected.

### DIFF
--- a/src/berry/modules/be_i2c.c
+++ b/src/berry/modules/be_i2c.c
@@ -7,18 +7,40 @@
 #include "../../driver/drv_local.h"
 #include "../../logging/logging.h"
 
+// Berry calls these as methods, so stack index 1 is the instance and the
+// softI2C_t lives in its ".p" member. The old code read index 1 as a raw
+// comptr, which is never true for a method call - every call silently did
+// nothing. Raw comptrs are still accepted so direct calls keep working.
+static softI2C_t* i2c_self(bvm* vm)
+{
+	softI2C_t* i2c = NULL;
+	if(be_isinstance(vm, 1))
+	{
+		// note: be_newcomobj() stores a BE_COMOBJ, not a BE_COMPTR, so this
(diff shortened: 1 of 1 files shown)


divadiow:
AI
`**REQUEST SMALL FIX** | Code reviewed | The Berry calling-convention fix is correct and hardware-proven. Add malloc checks, bound read_bytes sizes and validate pins/address before merge.`

PR #2208, opened 2026-08-31
driver: add an RTC driver for BK7252N

DrDoof:
The BK7252N has an always on RTC counter. The SDK already uses it as the time base for `rtos_get_time_us()`, which on this chip reads the RTC register directly rather than anything derived from the scheduler tick, but nothing exposes it to the application.

This driver anchors that counter to NTP and derives a wall clock from it in hardware. Being precise about what that does and does not give you: it is a separate RTC backed time source, not a replacement for the device clock. Nothing outside the driver consumes it yet, and charts and deviceclock still take their time from NTP, so what this driver keeps is its own epoch rather than the rest of the firmware's notion of time. Wiring it into the global clock is a sensible next step, but it is a different change and I would rather not fold it in here.

What it is useful for today is measurement. The index page and the new `RTCTime` command report how far the software second counter has drifted from the hardware one, which turns an assumption into a number.

Anchoring happens only on a genuinely new NTP response. `NTP_IsTimeSynced()` is a latch and `NTP_GetCurrentTime()` returns the software clock, so the NTP driver now counts the responses it actually receives and the RTC driver acts only when that count moves. The epoch comes from the no offset UTC time, and the clock is anchored again only when a fresh response disagrees with it by five seconds or more. The driver counts how often that happened.

**Scope**

* `src/driver/drv_rtc.c`, the new driver
* `src/driver/drv_ntp.c` and `.h`, `NTP_GetSyncCount()`, a counter of genuine NTP responses
* prototypes in `drv_local.h`, an entry in the driver table, the source added to `obk_main.mk`
* enabled by `ENABLE_DRIVER_RTC`, set for `PLATFORM_BK7252N` only, so no other platform is affected

**Testing**

An earlier revision was run on a BK7252N device. After `startDriver BKRTC` the driver anchored on the first NTP response, and the clock it derived then held a constant offset against an external reference across repeated samples taken tens of seconds apart. The offset did not grow, which is what distinguishes a hardware advanced clock from a drifting one. Hardware uptime advanced exactly with wall time, and the reported drift stayed at zero to one second.

The rebased head carrying the anchoring change is built for BK7252N but has not been run on hardware yet.

Deep sleep behaviour is not verified either. The counter sits in the 32kHz always on domain and `rtc_get_counter_val()` assembles a genuine 64 bit count from the low and high registers, so it ought to survive sleep, but I have not measured it and would rather say so than assume.


openshwprojects:
Nice, we needed this. This is also accepted for merge but please resolve conflicts.

divadiow:
AI:

`The feature is BK7252N-only and its hardware-counter epoch calculation is reasonable. Use signed wide arithmetic for drift, verify deep-sleep behaviour, and describe it accurately as a separate RTC-backed time source unless it is integrated into the global clock.`

DrDoof:
Three separate things in there, and they don't all land the same way.

The drift arithmetic I'd push back on. Both sides are `unsigned int`, and `(int)(a - b)` on two unsigned counters is the usual way to get a signed difference. It's well defined for anything under 2^31, so roughly 68 years of skew before it misbehaves. Widening it wouldn't buy anything either, since these are epoch seconds and 32 bits runs out in 2106 regardless of what I do here.

Deep sleep is a fair thing to ask about, and it sent me back to the SDK, where the answer is better than I expected:

```c
uint64_t rtos_get_time_us(void)
{
    #if !(CFG_SOC_NAME == SOC_BK7252N)
    return (uint64_t)cal_get_time_us();
    #else
    return (uint64_t)rtc_reg_get_time_us();
    #endif
}
```

On BK7252N that's a direct read of the RTC register rather than anything tick derived, and `rtc_get_counter_val()` builds a genuine 64-bit count from the low and high registers with a re-read loop to avoid tearing, so there's no wrap to worry about either. 31.25us per count puts it in the 32kHz always-on domain. All of which says it should survive sleep, but I'd rather measure that on a board than tell you it does, so please treat it as unverified for now.

The third point is the one that's right, and it's the one I got wrong in the description. Nothing outside the driver calls `BKRTC_GetEpoch()`. Charts and deviceclock still take their time from NTP, so what stays correct while NTP is away is the driver's own epoch, not the device's clock. Saying "the wall clock stays correct" oversells it. What this actually is, is an RTC-backed reference you can compare the software clock against, and I'll reword the description to say so. Wiring it into the global clock is a real thing to do, but it's a different change and I'd rather not smuggle it in here.

One last thing that isn't mine but is worth someone's attention: `#define RTC_CLK_PER_CNT_EQL_31_25_US 3125 / 100` in the SDK has no parentheses. It happens to be correct where it's used, because `cnt * 3125 / 100` evaluates left to right, but written the other way round it quietly becomes `31 * cnt`. That's 0.8%, or about eleven minutes a day.

divadiow:
AI:
---------------------
Thanks, the SDK references do strengthen the case that `rtos_get_time_us()` on BK7252N is genuinely RTC-backed rather than scheduler-tick based, and I agree the revised PR description is much more accurate about this being a separate time source rather than replacing the global device clock.

I still think the drift subtraction should be made explicitly signed, though. In:

```c
g_rtcLastDelta = (int)(ntp - BKRTC_GetEpoch());
```

the subtraction happens as unsigned arithmetic before the cast. On the current ARM GCC target it will almost certainly behave as intended, but the conversion back to signed is implementation-defined when the unsigned result is out of range. An `int64_t` subtraction would make the intended semantics explicit and portable.

More importantly, looking at the surrounding NTP/device-clock code exposed a functional issue I think needs fixing before merge.

`NTP_IsTimeSynced()` is effectively a latched "NTP has synced at least once" state. It remains true after NTP becomes unreachable. Also, `NTP_GetCurrentTime()` does not return a fresh NTP sample, it returns the software device clock.

That means after one successful NTP sync, if NTP later disappears, this driver will continue comparing the RTC-derived epoch against the software clock every second. If that software clock drifts by 5 seconds, the current code then rebases the RTC clock onto the drifting software clock:

```c
if (abs(delta) >= 5)
    BKRTC_SetBase(ntp);
```

So the RTC time source can eventually lose exactly the independence from the software clock that this driver is intended to provide.

I think re-anchoring should only happen when a genuinely new NTP response has been received, not merely while `NTP_IsTimeSynced()` remains true. A last-success timestamp or NTP sync-generation counter would make that distinction cleanly.

A couple of smaller points remain too:

* `BKRTC_GetEpoch()` is anchored from `NTP_GetCurrentTime()`, which includes the device clock's timezone/DST offset. If this is meant to be an epoch value, I think it should use the no-offset/UTC time instead.
* The generated driver description still says the clock "stays correct while NTP is unreachable", which is stronger than the revised PR description and, with the stale-sync behaviour above, currently isn't guaranteed.
* The metadata still has `"title":"TODO"`.
* The testing text still says `startDriver RTC`, but the driver is now `BKRTC`.
* It would be worth guarding against `BKRTC_GetUptimeUs()` moving backwards/resetting, otherwise the unsigned elapsed subtraction can underflow badly.
* Deep-sleep continuity still looks plausible from the SDK, but I think keeping it explicitly unverified is the right wording until tested.

Also, the branch is now behind current `main` and needs rebasing again before the latest CI result can be treated as validation of the final merge state.

So I think the general approach is sound, and the BKRTC rename clearly fixed the cross-platform symbol collision, but I wouldn't merge the current head until the stale-NTP re-anchoring issue is fixed.

DrDoof:
You're right about the reanchoring, and it's the one that actually breaks what the driver is for. `NTP_IsTimeSynced()` latches once and never goes back, and `NTP_GetCurrentTime()` hands back the software device clock rather than anything freshly off the wire, so the moment the server disappears the driver was quietly comparing the RTC against a drifting local counter and then pulling the RTC onto it, which is exactly the independence it was supposed to give you.

So there's a counter now. `NTP_GetSyncCount()` lives in `drv_ntp.c` and its header, incremented at the point where the device time is actually set from a received packet, and in the Windows simulated time path too, not at the point where the synced flag gets read. `drv_rtc.c` remembers the count it last acted on, sets the initial base only when that count has moved, and rebases only when the count has moved and the drift is five seconds or more. A stale sync can't touch the clock at all any more. The drift against the software clock is still computed every second, since reporting that number is most of the point of the driver, it just has no authority to rebase on its own.

The epoch is anchored from `NTP_GetCurrentTimeWithoutOffset()` now, so it carries no timezone or DST offset, and the drift is computed by widening both operands to `long long` before the cast back to `int`, so it's signed and wide with no unsigned wraparound to reason about. For the uptime case, if `BKRTC_GetUptimeUs()` ever comes back lower than the stored base, the base is dropped with a warning and established again on the next fresh NTP response, so the unsigned elapsed subtraction has nothing left to underflow into.

On the text, the drvdetail description no longer claims the clock stays correct while NTP is unreachable, it now says the clock is anchored to NTP and anchored again only when a fresh response disagrees by five seconds or more. The title field is filled in as `BK7252N RTC clock` instead of TODO, and the PR body says `startDriver BKRTC`. The branch is rebased on current main, where it ran into the audio and SD card drivers that landed today, and both driver table entries are kept.

Being straight about testing: the earlier revision ran on real BK7252N hardware and held a constant offset against an external reference, but this head with the anchoring change has only been built for BK7252N, not run yet. Deep sleep behaviour is still unverified.

openshwprojects (approved):

PR #2210, opened 2026-08-31
driver: add a microphone driver for BK7252N

DrDoof:
Adds a `BKAudio` driver that captures from the BK7252N on-chip audio ADC and reports the input level.

**Depends on** NonPIayerCharacter/beken_freertos_sdk#6. The SDK currently keeps the audio block switched off for this chip, so `aud_adc` is never registered and the driver will report `microphone not open` until that lands. Everything is behind `ENABLE_DRIVER_BKAUDIO`, set for `PLATFORM_BK7252N` only, so no other platform is affected either way.

**What it does**

- captures at 8 kHz mono in interrupt mode; DMA mode resets this part, so the mode is an argument of `BKAudioOpen` rather than a compile-time choice
- reports RMS, per-second peak, and a **peak held until it is read**, so a short sound is detectable without having to poll at the right moment
- reports **the sample rate it actually observes** next to the nominal one, plus failed-read and overflow counters

**On that last point**

The ring buffer holds 128 ms at 8 kHz. Draining it from `onEverySecond`, as the first version did, means observing roughly a tenth of the signal and reporting a snapshot of it — which produced a confident false negative during bring-up: clapping next to the microphone changed nothing measurable. Draining from `runQuickTick` instead brings the observed rate to 7873–8080 samples per second against a nominal 8000, and the same clap then moves the held peak from 144 to 5956 out of full scale with the RMS going from 6 to 189, returning to the noise floor immediately afterwards.

The observed-rate figure is printed for exactly that reason: without it, "no signal" and "not looking" are indistinguishable.

**Commands**

`BKAudioOpen [modeBits] [rate]`, `BKAudioClose`, `BKAudioInfo` (prints statistics and clears the held peak), `BKAudioDump` (sixteen most recent raw samples), `BKAudioGain [0..124]`.

Gain currently appears to have no effect, most likely because AGC is configured in `audio_hardware_init()` and overrides the manual PGA value. Left as is rather than guessed at.


openshwprojects:
Nice, we needed this. This is also accepted for merge but please resolve conflicts.

PR #2211, opened 2026-08-31
driver: add an SD card driver for BK7252N

DrDoof:
The SDK ships a complete SD host driver for this part, but nothing calls it, so the linker drops everything except the registration stub. This adds a driver that mounts the card, reports what is in the slot, reads raw blocks, and can lay down a fresh FAT32 filesystem.

Commands: `SDCardMount`, `SDCardUnmount`, `SDCardInfo`, `SDCardDetectPin`, `SDCardRead`, `SDCardDump`, `SDCardFs`, `SDCardFsBase`, `SDCardFormat`, `SDCardZero`. Card identity, capacity and access counters appear on the status page. Everything is behind `ENABLE_DRIVER_BKSDCARD`, which is defined for BK7252N only.

The bus is on P14 to P19 on this part; the alternate group at P34 to P39 is unavailable because the camera interface occupies P27 to P39. Card presence is read from a configurable pin, defaulting to P23, since the SDK's own detection is compiled out.

### Two things worth flagging

`bk_sd_card_init()` configures the SDIO peripheral but never initialises the host driver itself, which the vendor examples do from their CLI layer. Without that call it returns `BK_ERR_SDIO_HOST_NOT_INIT` whatever is in the slot, so the driver makes the call itself. The stock vendor firmware for this part does the same, so this looks like a gap in the SDK rather than a misuse of it.

The FatFs the SDK ships is cut down to a music player: it exports `f_mount`, `f_read`, `f_lseek` and `f_close`, but no `f_open`, no directory API, and no write path at all. So `SDCardFs` can tell you a card holds a FAT filesystem, and no more than that. Its three entry points, `disk_initialize`, `disk_status` and `disk_read`, are supplied here because the shipped `disk_io.c` supplies none of them: its body sits behind `#if 0` and targets the older card driver this part does not use. Note also that `f_mount` alone does not touch the medium in this vintage of FatFs, it registers the volume and returns success on an empty slot, so mounting goes through `chk_mounted_con` instead.

This layer is thin, but the block driver reaches into it in nine places, and separating them would need a third switch and nine more guards carried forever.

### Formatting

`SDCardFormat` writes a partition table with one FAT32 entry, the boot record and its backup, the information sector and its backup, and both allocation tables. It takes the card's exact sector count and refuses to run unless it matches what the card reports, so a stray argument cannot trigger it and the argument only works on the card it was read from.

`bk_sd_card_erase` is attempted for the table area but not relied upon: a card may refuse it, and the standard permits erased blocks to read back as ones rather than zeroes. When that happens the command reports the format as incomplete and names the range that still has to be cleared, which `SDCardZero` does in bounded runs so a long wipe cannot starve the watchdog.

Two limits worth knowing: the format uses a fixed cluster size of 64 sectors and a partition starting at LBA 2048, so a card too small for FAT32 is rejected by the cluster count check; and `SDCardZero` will clear any range it is given, its only guard being the card's size.

### Tested on hardware

On a 64 GB card: identified as SDHC/SDXC with 124735488 sectors, matching the capacity computed from its CSD. Sector 0 ended `55AA`, sector 1 held `EFI PART`, and the backup GPT header sat in the last sector, which is where the specification puts it and which confirms the sector count independently.

After formatting, every field was read back from the card and checked against the specification rather than against the code that wrote it: partition type, start and length, the whole BPB, the information sector signatures and free count, both backups, and the terminator in both allocation tables. The filesystem mounts and survives a reboot, and the camera pipeline kept running throughout without errors.

### DIFF
--- a/platforms/obk_main.mk
+++ b/platforms/obk_main.mk
@@ -75,6 +75,8 @@ OBKM_SRC  += $(OBK_SRCS)littlefs/lfs.c
 OBKM_SRC  += $(OBK_SRCS)littlefs/our_lfs.c
 
 OBKM_SRC  += $(OBK_SRCS)driver/drv_main.c
+OBKM_SRC  += $(OBK_SRCS)driver/drv_sdcard.c
+OBKM_SRC  += $(OBK_SRCS)driver/drv_sdcard_fs.c
 
 OBKM_SRC  += $(OBK_SRCS)driver/drv_adcButton.c
 OBKM_SRC  += $(OBK_SRCS)driver/drv_adcSmoother.c
--- a/src/driver/drv_local.h
+++ b/src/driver/drv_local.h
@@ -137,6 +137,20 @@ void DRV_HTTPButtons_AddToHtmlPage(http_request_t *request);
 void DRV_InitHTTPButtons();
 #if ENABLE_HA_DISCOVERY
 void DRV_HTTPButtons_OnHassDiscovery(const char *topic);
+
+void BKSDCard_Init(void);
+void BKSDCard_OnEverySecond(void);
+void BKSDCard_StopDriver(void);
+void BKSDCard_AppendInformationToHTTPIndexPage(http_request_t *request, int bPreState);
+int BKSDCardFs_Mount(void);
+int BKSDCardFs_Unmount(void);
+int BKSDCardFs_IsMounted(void);
+int BKSDCardFs_LastResult(void);
+void BKSDCardFs_SetBase(unsigned int lba);
+unsigned int BKSDCardFs_GetBase(void);
+unsigned int BKSDCardFs_Reads(void);
+unsigned int BKSDCardFs_ReadFails(void);
+
 #endif
 
 void CHT83XX_Init();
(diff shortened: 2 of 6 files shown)


openshwprojects:
Very nice, those drivers can be merged now. Can you make some kind of presentation of new BK7252N features for our forum?

DrDoof:
Yes, I'd like to do that. You asked the same when the charger driver went in and I haven't delivered yet, so let me be straight with you about the timing rather than going quiet: I've got a pile of personal things I need to clear first, and it won't be this week. It will happen, I'm just not going to promise you a date I can't keep.

When I get to it I'd rather write one piece covering the whole BK7252N picture than scatter it over the individual PRs. Charger, SD card, microphone, RTC and the audio ADC, with the wiring and the actual config lines someone needs to get a board up. Anyone arriving from the forum wants the whole thing in one place, and it saves you stitching five notes together.

Thanks for merging these. And thanks for moving the SDK submodule as quickly as you did, that one mattered more than it looked, the charger driver was reporting reset state to everyone on 1.18.303 without it.

PR #2221, opened 2026-09-06
Fix ESP-IDF MQTT lwIP thread ownership

DeDaMrAzR:
This fixes an ESP-IDF-specific MQTT/lwIP thread ownership problem found while investigating intermittent watchdog resets.
Incoming MQTT callbacks run in the TCP/IP task, while OBK processes the received data from its own task. The shared receive buffer previously used the broader MQTT mutex, while raw lwIP operations could also be called without ESP-IDF TCP/IP core ownership. Under sustained MQTT traffic, this could produce an invalid interleaving and eventually trigger a very random watchdog reset that I spent 10 days chasing.

The change is intentionally small:
- Enable `CONFIG_LWIP_TCPIP_CORE_LOCKING` for ESP-IDF builds.
- Use a dedicated mutex for the MQTT receive ring buffer on ESP-IDF.
- Keep the existing mutex behavior unchanged on other platforms.
- Route the remaining direct reconnect-time disconnect through OBK’s core-locking wrapper.

Just to expand on the matter, I devised and ran multiple crash tests and out of 4 on demand induced crashes all 4 were pointing to this issue. After this edit I was not able to trigger a crash just a normal MQTT reconnect.


NonPIayerCharacter (approved):

DeDaMrAzR:
Created a CI run on my fork and doing another 120+ hours of live testing on the device with so far no errors, no reboots, strange crashes, core dumps or mqtt disconnects. 

to stress again I am doing all the tests that I can think of on actual HW and will present final results after extensive testing, but I would love to get any feedback if anybody did a similar check on their device as I can make all of mine work on my bench 😁

PR #2227, opened 2026-09-14
Pir fixes

Sn0w3y:


openshwprojects:
We could merge it but we need compile/test errors fixes first

PR #2232, opened 2026-09-16
HA discovery: publish the real light for TuyaMCU LEDs (fixes #2218)

### LINKED ISSUE #2218: TuyaMCU LED driver (tuyaMcu_setupLED): HA discovery publishes a light bound to raw channels 1/2 instead of the LED driver
**Device:** Feit FETAPE/RGBW/CONT LED strip controller
**Chip:** BK7231T (bare QFN32, `BK7231TQN32`)
**Firmware:** OpenBK7231T 1.18.304
**Setup:** TuyaMCU (no PWM pins configured — the BK7231T is a WiFi bridge, a
separate MCU drives the LEDs)

### Startup command
```
backlog tuyaMcu_setBaudRate 9600; startDriver TuyaMCU; tuyaMcu_defWiFiState 4; tuyaMcu_setupLED 24 1; tuyaMcu_sendQueryState
```
dpIDs: 20 on/off, 21 work_mode, 22 white brightness, 24 colour (HSV string).
No dpID 23 — the hardware is RGB + a single warm white.

### What works
Everything except HA discovery. Over MQTT `cmnd/<topic>/...` and from the web UI:
`POWER ON/OFF`, `led_dimmer 0-100`, `CT 500` (warm white), and
`led_basecolor_rgb RRGGBB` all drive the strip correctly. The device reports as a
full colour light: `{"Dimmer":100,"Color":"255,255,255,0,0","CT":154,"POWER":"ON"}`

### The bug
Pressing "Home Assistant Configuration -> publish discovery" makes Home Assistant
register a light and then delete it milliseconds later. From the HA log:

```
22:16:37.320 INFO [homeassistant.helpers.entity_registry] Registered new light.mqtt entity: light.stove_side
22:16:37.322 INFO [homeassistant.components.mqtt.entity] Removing component: light.stove_side
22:16:54.122 INFO [homeassistant.helpers.entity_registry] Registered new light.mqtt entity: light.sink_side
22:16:54.133 INFO [homeassistant.components.mqtt.entity] Removing component: light.sink_side
```

It publishes the config for the OLD channel-1 light, then an empty payload to the
same topic to remove it. Removing it is correct — `tuyaMcu_setupLED` means there
is no channel-1 light any more — but **no replacement RGBCW light config is ever
published**, so the device ends up with no light entity in Home Assistant at all.

Before `tuyaMcu_setupLED` was added (when the config used
`linkTuyaMCUOutputToChannel 20 bool 1` / `22 val 2` and `setChannelType`), HA
discovery produced a working brightness-only light. Adding the LED driver gains
colour but loses the HA entity.

### Steps to reproduce
1. TuyaMCU device, LED driver set up with `tuyaMcu_setupLED <colorDp> 1`
2. Configure MQTT, connect to Home Assistant
3. Press publish discovery on the Home Assistant Configuration page
4. HA registers a light and removes it in the same instant; no light entity remains

### Tried, did not help
- `SetFlag 4 1` ([LED] Force show RGBCW controller) — flags became 1040, no change
- Republishing discovery repeatedly — identical result each time
- Reloading the HA MQTT integration — only re-reads the same stale retained config

### Expected
With the TuyaMCU LED driver active, discovery should publish an RGBCW light
config (colour + brightness) in place of the removed per-channel light.

### Workaround
Define the light by hand in HA's YAML against `cmnd/<topic>/POWER`,
`cmnd/<topic>/led_dimmer` and `cmnd/<topic>/led_basecolor_rgb`, which all work.

codypendant:
Fixes #2218.

A TuyaMCU LED configured with `tuyaMcu_setupLED` is driven over the serial link
to the MCU. It has no PWM pins and no channel bindings, and `doHomeAssistantDiscovery`
got it wrong in two independent ways.

### 1. The LED light was never published

```c
if (ledDriverChipRunning) {
    pwmCount = CFG_CountLEDRemapChannels();
}
if (pwmCount == 5 || ...) { ... }
else if (pwmCount > 0) { ... }
```

`CFG_CountLEDRemapChannels()` counts non-`0xff` entries in `g_cfg.ledRemap.ar[]`.
A TuyaMCU LED has no PWM/remap channels at all, so this returns **0** and neither
branch runs.

Worth noting `ledDriverChipRunning` was never the problem —
`LED_IsLedDriverChipRunning()` already returns true for TuyaMCU LEDs. Only the
remap-channel count is meaningless for this device class.

### 2. A light bound to raw channels was published instead

The toggle/dimmer pairing block then paired the first `ChType_Toggle` with the
first `ChType_Dimmer` and published:

```json
{"stat_t":"~/1/get","cmd_t":"~/1/set",
 "bri_stat_t":"~/2/get","bri_cmd_t":"~/2/set","bri_scl":100}
```

On a TuyaMCU LED those channels drive nothing, so HA creates a light that cannot
control the device. That block is *intended* for TuyaMCU devices with custom
channel types — its own comment says so — but a TuyaMCU LED is not channel-driven.

### The fix

`TuyaMCU_IsLEDRunning()` already exists and is already exported in `drv_public.h`;
the discovery path just never consulted it. When it returns true:

* skip the toggle/dimmer pairing, since the light is not channel-driven
* treat the device as **RGBCW**

RGBCW is the correct type because `TuyaMCU_OnRGBCWChange` implements exactly that:
RGB through the colour dpID, and white through the brightness and temperature
dpIDs.

### Test

`Test_HassDiscovery_TuyaMCU_LED` in `selftest_hass_discovery_ext.c`, registered in
`Test_HassDiscovery_Ext()`. It sets up a TuyaMCU LED *and* toggle/dimmer channel
types, then asserts both halves: the `led_*` topics are published, and nothing is
bound to `~/1/get` / `~/2/get`.

The test resets the LED config at the end with `tuyaMcu_setupLED -1 0` — those are
file-static globals that `SIM_ClearOBK` does not clear and `TuyaMCU_Shutdown()`
does not reset, so without it every later test in the suite would run with an LED
attached.

### Testing done

Reported and reproduced on two BK7231T devices (`tuyaMcu_setupLED 24 1`, dp20
on/off, dp22 brightness, 9600 baud), captured from the broker.

**I have not been able to build or run the selftests myself** — they are
`#ifdef WINDOWS` and built with MSBuild, and I don't have a Windows toolchain
available. The change is reviewed by inspection; CI needs to validate it. Happy
to iterate if the test needs adjusting.


codypendant:
CI is green now (66 pass). For anyone reading the earlier red run: that failure
was my test being wrong, not the fix.

The first version asserted that nothing publishes `stat_t: "~/1/get"`. But
channel 1 in the test is a `ChType_Toggle`, and a toggle channel legitimately
gets its own switch entity on that topic — suppressing the toggle/dimmer *light*
pairing must not suppress ordinary channel entities. I'd written an assertion
that would fail on correct behaviour.

Everything that actually tests the fix passed on that first run:

* `bri_stat_t` = `~/led_dimmer/get` — pass
* `stat_t` = `~/led_enableAll/get` — pass
* `rgb_stat_t` = `~/led_basecolor_rgb/get` — pass
* `clr_temp_stat_t` = `~/led_temperature/get` — pass
* no `bri_stat_t` = `~/2/get` — pass (the bogus light is gone)

The signature of #2218 is a *light* carrying brightness on a raw channel topic,
so the assertions now key on `bri_stat_t` / `bri_cmd_t`. I also flipped the bad
assertion into a positive one — the toggle channel must still get its own entity
— so the test now guards against over-suppressing as well.

Passing under both the Address and Undefined Behavior sanitizer runs.

codypendant:
**Correction — the premise of this PR is wrong, and I'd like to revise it.**

I claimed the LED light was never published because
`CFG_CountLEDRemapChannels()` returns 0 on a TuyaMCU LED, so `pwmCount` stayed 0
and neither branch of the `ENABLE_LED_BASIC` block ran.

That is not what happens. Testing #2234 on the affected device shows the LED
block **was always running** — `pwmCount` is 5 here. The reason no usable light
appeared is the use-after-free in #2234: the block's `if (dev_info == NULL)`
guard was defeated by a dangling pointer, so instead of allocating its own
entity it re-published the freed channel-light's stale topic with `{}`.

With #2234 alone, and **without** this PR, the correct entity is published:

```json
{"name":"Light","uniq_id":"OpenBK7231T_0668EFA8_light",
 "stat_t":"~/led_enableAll/get","bri_stat_t":"~/led_dimmer/get",
 "rgb_stat_t":"~/led_basecolor_rgb/get","clr_temp_stat_t":"~/led_temperature/get"}
```

Home Assistant creates a working light from it. So **#2234 is the actual fix for
 #2218**, not this PR.

### What this PR is still good for

The inert channel-bound light is still published alongside the real one:

```json
{"name":"1","uniq_id":"OpenBK7231T_0668EFA8_light_1",
 "stat_t":"~/1/get","cmd_t":"~/1/set","bri_stat_t":"~/2/get","bri_cmd_t":"~/2/set"}
```

On a TuyaMCU LED those channels drive nothing, so this is a second, dead light
entity next to the working one. Suppressing the toggle/dimmer pairing when a
TuyaMCU LED is running still seems right to me — but as a duplicate-entity
cleanup, not as the fix for a missing light.

The `pwmCount = 5` forcing in this PR is now unnecessary and I should drop it,
since the LED block already runs. Happy to rewrite this as a
skip-the-pairing-block-only change with a corrected description and test, or to
close it if you'd rather handle the duplicate differently. Let me know which you
prefer and I'll do the work — I don't want to leave a PR sitting here with a
wrong explanation attached to it.

codypendant:
Closing this. My explanation was wrong, the LED block does run here (pwmCount is 5), so this was never the reason the light was missing. #2234 is the real fix and I've tested it on the device.

The leftover duplicate channel light is minor. I'll open a separate issue for it if it's worth bothering with.

PR #2234, opened 2026-09-16
HA discovery: fix use-after-free / double free of dev_info (#2230)

### LINKED ISSUE #2218: TuyaMCU LED driver (tuyaMcu_setupLED): HA discovery publishes a light bound to raw channels 1/2 instead of the LED driver

### LINKED ISSUE #2230: MQTT: 1.18.311 emits malformed packets and reconnect-loops; rolling back to 1.18.310 fixes it (regression from #2222)

codypendant:
Fixes #2230.

`doHomeAssistantDiscovery()` frees `dev_info` in the toggle/dimmer pairing block
without clearing it. `hass_free_device_info()` takes the pointer by value, so the
`dev_info == NULL` guard in the LED block below never fires — that block then
dereferences and re-frees freed memory.

The two other `hass_free_device_info()` call sites in the same function already
set `dev_info = NULL;` afterwards. This one was missed.

### Tested on hardware

BK7231T, flags 1040, TuyaMCU LED (`tuyaMcu_setupLED 24 1`), OTA of this PR's
`BK7231T_tuyaMCU` artifact.

Before, on 1.18.311 — same topic twice, second payload is the freed cJSON:

```
Publishing val (374 bytes) to .../OpenBK7231T_0668EFA8_light_1/config retain=1
Publishing val {} to .../OpenBK7231T_0668EFA8_light_1/config retain=1
```

After — distinct topics, real payloads:

```
Publishing val (374 bytes) to .../OpenBK7231T_0668EFA8_light_1/config retain=1
Publishing val (848 bytes) to .../OpenBK7231T_0668EFA8_light/config retain=1
```

9 queued, 9 published, no disconnects, and no `malformed packet` on the broker
since the flash.

### Also fixes #2218

The 848-byte entity is the correct LED light (`~/led_enableAll/get`,
`~/led_dimmer/get`, `~/led_basecolor_rgb/get`). The LED block was always running
on this device — it was publishing the freed pointer's stale topic with an empty
body instead of allocating its own entity.

#2232 and #2233 are both closed: #2232's premise was wrong, and the queue issues
in #2233 are real-looking but unrelated to this.


codypendant:
Tested — **the fix works.** Thanks for the pointer to the PR artifacts, that made
this easy.

OTA'd `OpenBK7231T_2234_merge_58cd7fb871dc_tuyaMCU.rbl` onto the affected device
(BK7231T, flags 1040, TuyaMCU LED via `tuyaMcu_setupLED 24 1`).

### Before, on 1.18.311

```
Queued topic=homeassistant/light/OpenBK7231T_0668EFA8_light_1/config, 1 items in queue
Queued topic=homeassistant/light/OpenBK7231T_0668EFA8_light_1/config, 2 items in queue
Publishing val (374 bytes) to homeassistant/light/OpenBK7231T_0668EFA8_light_1/config retain=1
Publishing val {} to homeassistant/light/OpenBK7231T_0668EFA8_light_1/config retain=1
```

Second entry has the **same** topic (the freed `->channel`) and `{}` (the freed
cJSON).

### After, with this PR

```
Queued topic=homeassistant/light/OpenBK7231T_0668EFA8_light_1/config, 1 items in queue
Queued topic=homeassistant/light/OpenBK7231T_0668EFA8_light/config, 2 items in queue
Publishing val (397 bytes) to homeassistant/light/OpenBK7231T_0668EFA8_light_1/config retain=1
Publishing val (848 bytes) to homeassistant/light/OpenBK7231T_0668EFA8_light/config retain=1
```

Different topic, real payload. 9 queued, 9 published, no disconnects, and **no
`malformed packet` on the broker since the flash**.

### It also fixes #2218

I did not expect this. The 848-byte entity is the correct LED light:

```json
{"name":"Light","uniq_id":"OpenBK7231T_0668EFA8_light",
 "stat_t":"~/led_enableAll/get","cmd_t":"cmnd/obk0668EFA8/led_enableAll",
 "bri_stat_t":"~/led_dimmer/get","bri_cmd_t":"cmnd/obk0668EFA8/led_dimmer",
 "rgb_stat_t":"~/led_basecolor_rgb/get","clr_temp_stat_t":"~/led_temperature/get"}
```

Home Assistant created a working light from it. So the LED block was *always*
running on this device — it was just publishing the freed pointer's stale topic
with an empty body instead of allocating its own `dev_info`. That is the real
cause of #2218 as well, and it means my analysis in #2232 was wrong about *why*
the light was missing. I'll correct that PR.

The inert channel-bound light (`_light_1`, `stat_t: ~/1/get`) is still published
alongside it, which is what #2232 addresses — but that is cosmetic next to this.

openshwprojects (approved):

Issue #866, opened 2023-06-27
Feature Request: White Blend Mode for RGBW lights


neominik:
I have a MagicHome RGBW LED strip where the white LED is neither cold nor warm, but pretty neutral. The default firmware uses the red and blue LEDs in addition to the white channel to mix the proper temperature. Here, I can use flag 24, but that assumes the white to be warm white, which is not the case. I also can't use the normal RGBCW mode, because warm white is just off (unmapped channel) in that case.

Tasmota has a solution for this, namely the [white blend mode](https://tasmota.github.io/docs/Lights/#white-blend-mode), where the light is exposed as RGB, but those values get the white led mixed in to improve brightness and color output. It works pretty well, but unfortunately not on the BL602 :(
I'm not sure how easily such a mode can be replicated in this project, since I haven't really figured out the codebase yet, but this would be great to have. Most people seem to have RGBCW lights, but I've stumbled across comments from other people with similar problems in other issues as well. Unfortunately, none of those have a concrete solution.

openshwprojects:
Hello, so you are basically looking for modified flag 24? I haven't been aware about a "non cold non warm" RGBW lights, but maybe it could work with some modifications to flag 24 handling?

As far as I remember, we still have to improve the flag 24 itself, there have been some reports saying that colour fractions in C emulation should be improved.

neominik:
I'm not sure if we should modify flag 24 add a separate flag or modify 24 with a separate flag. I guess 24 might make sense for some people as is, but I don't know how (much) it is being used.
The behaviour of the proposed white blend mode (as implemented in Tasmota) is actually pretty simple:
Convert an incoming RGB value to RGBW in this way: `W = min(R, G, B); R = R - W; G = G - W; B = B - W`

panchomira:
Hello! Something similar happened to me. This is my documentation about the issue (https://github.com/openshwprojects/OpenBK7231T_App/issues/763)

Issue #763, opened 2023-04-03
RGBCW Bulbs with no COOL LEDs issue. Inaccurate Temperature setting 


panchomira:
_First at all. Thanks to the owner of this project! Its amazing to be able to really be the owners of the products that we buy, ant the possibility to run open source firmware._  

![AGC_20230403_065551593](https://user-images.githubusercontent.com/98633070/229437407-9c00a057-55e6-4993-bf1b-c01a868a4c0f.jpg)

**LAMP:** Etheos 9w E27
**Chip: Chipset:** BK7231N
**Firmware:** Version: 1.15.637

**The problem that, besides me, I saw another one have it and probably a few more will have.** 

As far as I can tell, my bulb does not have a cool white LED. Therefore, the original Tuya firmware achieves Cool white with RGB LEDs.

![AGC_20230403_065641214](https://user-images.githubusercontent.com/98633070/229437379-dfbf763f-241c-4c6a-a52c-d261ed98d1ce.jpg)

Currently, I am using Flag 24 ([LED] Emulate Cool White with RGB in device with four PWMS - Red is 0, Green 1, Blue 2, and Warm is 4), which tries to do the same thing, but it's way off because it doesn't work the same way as the original firmware.

**In the original firmware, when you set the temperature to around 4500K, all the warm white LEDs and RGB LEDs turn on (although I don't think they're all at the same intensity). Then, when you select the coolest white, only the warm white LED, the blue and green RGB LED turn on, not red.**
_See documentation at the end_

I tested the white balance on the original firmware using a camera, and it's pretty accurate in terms of the Kelvin temperature. Even when emulating temperature with RGB, it produces very little color in the final light projection, staying very true to a pure white.

**Now, on OpenBK, it goes with all the RGB LEDs as soon as soon as the lamp its not in the warmest setting, and stays the same (but changing the intensity of all RGB LEDs) even in the Coolest setting (not the way the original firmware does it)** 
_See documentation below_

Here are screenshot of the configuration and how it reflects on the LEDs of the bulb. Both in stock firmware and OpenBK
Original Firmware: 

Warmest
![Screenshot_20230403-035835_Tuya Smart](https://user-images.githubusercontent.com/98633070/229438816-e38048f0-ef72-46c2-97d0-44e44606cd70.png)
[...]

Issue #714, opened 2023-03-07
BP5758D does not respect Flag24


no2chem:
**Describe the bug**

Hi All, I have a new Treatlife SL10 bulb which appears to be based on CB2L (BK7231N). I am able to get it to talk to the BP5758D on board (P7=BP5758D_CLK, P8=BP5758D_DAT), mapping (BP5758D_Map 4 3 2 1 0) but it turns out that it only has 4 channels, analyzing the board, channel 0 appears to be floating and the white LEDs are WW. 

Flag24 should cause CW to be emulated by RGB, but it doesn't seem to work with BP5758D. Instead the light sort of just dims out due to the empty channel.

**Firmware:**
- Version [e.g. 1.14.129] 1.15.559
- Device? Treatlife SL10 CB2L
- Chip/model: BK7231T? BK7231N
- Device config? P7=BP5758D_CLK, P8=BP5758D_DAT

**To Reproduce**
Configure BP5758D
Turn on flag 24
Set color temperature to 6000K

**Screenshots**

openshwprojects:
It's not a bug, it's a lack of feature. I will try to get it working for you today, can you help with testing?

openshwprojects:
I am not sure about this:
https://github.com/openshwprojects/OpenBK7231T_App/commit/add67dd280fea13d5c18bf1f6f712241f6fea62b
Can you check, what do you think?

no2chem:
Hm strange, I moved W to channel 4 (`BP5758D_Map 4 3 2 0 1`) but there's no W and it's now blue on cool white. At 2000K the bulb is now dark. 

I think at 6000K W needs to be at 100% on this bulb still

openshwprojects:
What may be the problem? I am not sure yet.

panchomira:
Hello! I have a similar problem, I think.

As far as I can tell, your bulb does not have a cool white LED. Therefore, the original Tuya firmware achieves it with RGB LEDs.

Currently, I am using Flag 24 ([LED] Emulate Cool White with RGB in device with four PWMS - Red is 0, Green 1, Blue 2, and Warm is 4), which tries to do the same thing, but it's way off because it doesn't work the same way as the original firmware.

In the original firmware, when you set the temperature to around 4500K, all the warm white LEDs and RGB LEDs turn on (although I don't think they're all at the same intensity). Then, when you select the coolest white, only the warm white LED and the blue RGB LED turn on, not the green or red.

[...]

PR #2236, opened 2026-09-18
CTRange: allow a single value range for bulbs with one white LED

ryan-ronnander:
I spent some time trying to get a [SYLVANIA Wifi RGBW bulb](https://shop.ledvanceus.com/products/sylvania-wifi-led-smart-a19-rgbw-color-amber-finish-filament-light-bulb-7w-2000k-no-hub-required-clear-1pk/) working with OpenBeken. It's an RGB plus a single fixed 2000 K filament on W, warm white 2K, no cold LED.

sylvania_filament_a19_75807<br />

- Cloudcutter [profile](https://github.com/tuya-cloudcutter/tuya-cloudcutter.github.io/blob/master/devices/sylvania-smartplus-40a19filcclwifi-rgbw-filament-bulb-v1.0.1.json)
- OpenBeken JSON [profile](https://github.com/user-attachments/files/32397276/sylvania-40a19filcclwifi-openbeken-singlevalue-ctrange.json)

I tried to configure this bulb as a color temperature range with a single point (`CTRange 500 500`). When trying this the code technically divided by zero in `LED_GetTemperature0to1Range()` and wrote NaN into both white channels.

---

To fix it, a single value "_range_" is now treated as one point, returning an even warm/cool split, which the existing normalization code path then drives to full on both white channels. That covers all three builds without the firmware knowing which channel is populated:

1. Only a cool LED channel = native bulb white
1. Only a warm LED channel = native bulb white
1. Both cool and warm LED channel = A blended single white in between max cool and max warm

A one unit range (`CTRange 499 500`) is the obvious way to dodge the divide by zero, and it does light the filament at the warm end, but it leaves a dead spot on the slider:

| Config | R G B C W | Result |
| --- | --- | --- |
| `CTRange 499 500` at 499, the cool end | `0 0 0 100 0` | dark, only the empty cool channel is driven |
| `CTRange 499 500` at 500, the warm end | `0 0 0 0 100` | filament at full |
| `CTRange 500 500`, with this patch | `0 0 0 100 100` | filament at full, and no position turns the bulb off |

Flag 24 (emulate cool with RGB) looked promising at first, but it synthesizes the cool half as plain R=G=B. On this bulb it meant the slider jumped to a purpleish white instead of sweeping through the color temperature range. The bulb dimmed as soon as you moved close to 2K, but also drove "RGB white" over the filament at full strength, so the bulb was no longer showing its 2000 K white at all.

The web slider and HA discovery already read `led_temperature_min`/`led_temperature_max`, so the slider shows a fixed 2000 K and HA gets `min_mirs` = `max_mirs` = 500 and accepts it. All of the default warm color presets in Home Assistant show up as 2K and toggle over accordingly:

image

---

For testing dual white channel LEDs I temporarily borrowed the RGB channels to stand in for them. Feeding the warm value into the red slot and the cool value into the blue slot (`SM2135_Map 4 1 3 0 2`) lit the bulb magenta, and moving warm over to green (`SM2135_Map 0 4 3 1 2`) lit it cyan, with the channels reading `0 0 0 100 100` throughout. Both white channels really are driven, and which LEDs that lights is a property of the mapping rather than of this code.

Added `Test_LEDDriver_CW_SingleValueRange`, which fails without the fix.

In ESPHome I have this bulb configured as an `rgbw` light with `color_interlock: true`. This PR gets OpenBeken working similarly. OpenBeken always presents a color temperature axis for a driver chip bulb, so a single value CTRange seems the only way to accomplish the same thing in the same bulb class.

It seems a little hacky, but simplistic enough to "just work" for these use cases.


ryan-ronnander:
I went looking for existing related reports and found three, all cases where only one white channel is physically populated and the temperature axis assumes two.

- #763 reports the white LEDs turning off entirely at the cool end, which is the same dead position as the `CTRange 499 500` row in the table above.
- #714 reports the bulb going dark at 2000 K.
- #866 reports a neutral white RGBW strip where normal RGBCW mode is unusable because the other white channel is unmapped.

This does not close any of them. #763 and #714 are asking for the cool emulation to track the stock firmware more closely, and nothing here touches flag 24. #866 is asking for Tasmota's white blend mode, which is a separate feature.

What this PR does is add the other operation mode for that type of bulb: rather than synthesizing a cool end that is not there, a single value CTRange pins the slider to the one "true white" point the bulb actually has, and the existing normalization drives it at full, whichever slot it is wired to.

Also related to #2162, which adds flag 52 for a warm only bulb, but PWM only. Solves the same problem through an entirely different code path.

ryan-ronnander:
Needed a couple extra commits to make sure the web server GUI can change between RGB and color temp mode when there is a single "white point" slider value. 

Everything was working great through Home Assistant, but a browser cache glitch masked a real issue, there was no way to toggle between RGB and W mode.

With these extra commits a single mouse click will trigger the change back (to 2K on this test bulb):

image

---

068b4f7f stands on its own here. Any bulb running flag 12 with a CTRange narrower than the default 154-500 could restore a temperature outside its own range on boot, single value or not. Maybe harder to hit in real world use, but it closes a path that was "possible" before.

